PR #51, opened 2019-05-20
Persistentvolumes

nsmith-:
Main lesson: need to grab _all_ the old labels from the openstack template when creating the cluster.
Need to make a separate persistent volume for each instance I think, for now just dev gets one.
Open question: when a cluster is re-created, what is the effect of `kubectl create -f rucio-graphite-pvc.yaml` ?  Does it pick up the old storage, or does storage have the same lifetime as the cluster?


vkuznet:
Nick, it is a good question, I don't really know if storage and clusters are
independent entities in OpenStack.

I usually avoid using `create` and instead use `apply` command. This guarantee
that resource will be created and my yaml will be applied.

Do you want me to merge this PR?

nsmith-:
Not yet, I want to figure out if I can point all instances to the same volume/share.

vkuznet:
ok, when you'll be ready to merge please explicitly mention this in PR.

nsmith-:
Ok this can be merged now

PR #55, opened 2019-06-11
Move configmap into describing YAML

ericvaandering:
Move interval to 4 hours, update scripts, and add sites

Keep int and testbed up to date

Add all sites for NanoAOD

Add resources for sync job


ericvaandering:
@vkuznet please merge this too

PR #74, opened 2019-10-02
Add connection monitoring for dev and int

ericvaandering:


vkuznet:
Eric, I don't mind to merge this PR but I want to share with you my recent inquiry/issue I found with running frequent cronjobs on k8s. I made autoscalers and as you have in this PR I run frequent cronjobs. This made real effect on k8s status. Turns out that frequent crons significantly affect k8s cluster and it starts consuming resources on master node which become busy to process cron's requests. The k8s cluster become less responsive. Therefore I suggest that you check if your PR does not affect your k8s cluster status. You may use kubectl top node command to check CPU/RAM of your nodes. If it is above 80% then you have a problem. I don't know exact remedy, may be we need to increase number of masters, but I spent one morning at CERN with Spyros (before k8s training session) to figure out this issue. He suggested that k8s team will work on solution for autoscalers with custom metrics but I suspect that your PR may inherit the same issue.

ericvaandering:
Go ahead and merge. But thanks for the heads up. I can always reduce the frequency of this if needed.

PR #80, opened 2019-10-23
use CRABSERVER 3.3.1911.rc3

belforte:


belforte:
this is what has been deployed in cmsweb-testbed by Muhammad
I have already installed this in cmsweb-test via k8s. This PR is to keep GIT aligned .

PR #91, opened 2019-12-04
K8s development

muhammadimranfarooqi:
@vkuznet ,

Could you please review and merge this PR. I have added namespace in cinder storage and updated cinder storage in couchdb and dqmgui.


vkuznet:
Changes are fine and I'll merge them, but it is not clear (yet) if we going to use cinder for them or CephFS. Let's discuss this in a meeting.

PR #103, opened 2019-12-15
K8s

muhammadimranfarooqi:


kubernetes/cmsweb/docs/cmsweb_testbed_cluster_doc.md
```diff
@@ -0,0 +1,49 @@
...
+- Go into CMSKubernetes/kubernetes/cmsweb directory:
+
+- Update yaml file with the new version. For example open relevant service file (xxx.yaml e.g crabserver.yaml) and change version:
+   - `vim services/xxx.yaml`
+
+- Deploy newly changed yaml file. First delete old service from K8S and deploy new one. For example: 
+
+   - `kubectl delete -f services/xxx.yaml`
+   
+ - Then, following command can be used to deploy service in testbed cluster.
```
vkuznet: Please add explicitly that `#PROD#` should be replaced with **6** spaces, otherwise the markdown strip off all six spaces into one when people visually see the document. I just noticed this when I read your markdown in gist.

May be it is because you used single back-quote, instead use 3 back-quotes for command below, i..e. 3 back-quotes, then on new line put the command and then on new line put 3 back-quotes again. This would make a block in markdown to be used to declare the command.

muhammadimranfarooqi: @vkuznet ,

I have u

> I see that you managed to make proper commit for dbs-global.yaml, right? If not I wanted to suggest that you can do these steps:
> 
> 1. rm services/dbs-global.yaml
> 2. git checkout -- services/dbs-global.yaml
>    The first step just remove the file and second step restore it as is from git repository. This was you can restore file w/o your tweaks of editor.

@vkuznet, yes, I managed to make proper commit and did exactly what you said here and changed accordingly. Now it should be fine.

muhammadimranfarooqi: > Please add explicitly that `#PROD#` should be replaced with **6** spaces, otherwise the markdown strip off all six spaces into one when people visually see the document. I just noticed this when I read your markdown in gist.
> 
> May be it is because you used single back-quote, instead use 3 back-quotes for command below, i..e. 3 back-quotes, then on new line put the command and then on new line put 3 back-quotes again. This would make a block in markdown to be used to declare the command.

@vkuznet , I changed file according to your instructions and gist is properly displaying six spaces. I have also added comments about #PROD# string. 

Thanks

Best
M Imran

PR #118, opened 2020-01-24
Updating documentation for cmsweb clusters

muhammadimranfarooqi:


vkuznet:
Muhammad, even though I merged this PR I suggest that you change location of k8s config files from your AFS public area to something common on EOS. We need a project area such that it will not be tight to a single user (think long term). Just think/explore where it can be done.

PR #137, opened 2020-02-24
Adding NATSSecrets in deployment script

muhammadimranfarooqi:


kubernetes/cmsweb/scripts/deploy.sh
```diff
@@ -336,7 +336,9 @@ deploy_secrets()
             # special case for DBS instances
             if [ "$srv" == "dbs" ]; then
                 files="--from-file=$conf/dbs/DBSSecrets.py"
-                for inst in $dbs_instances; do
+                files="$files --from-file=$conf/dbs/NATSSecrets.py"
```
vkuznet: Please add condition to check if such file is present (and similar to DBSSecrets.py) since for some deployments we may not have this file and then I don't know if kubectl will fail or not. Something like this:
```
if [ -f $conf/dbs/NATSSecrets.py ]; then
   files="$files --from-file=$conf/dbs/NATSSecrets.py"
fi
```

muhammadimranfarooqi: Please check now

PR #143, opened 2020-03-20
Putting back old configuration for frontend

muhammadimranfarooqi:


muhammadimranfarooqi:
The new configuration in alignment with backend services, I don't know why frontend does not start and log files are not geneated. Old configuration with explicit hostcert declaration works only for frontend. We even don't need securitycontext in frontend. May be its in default namespace, this might be the reason.

Best regards

muhammadimranfarooqi:
We'll use explicit path declaration for certificates in frontend because we require certificates directory in FE which is not created without explicit declaration.

PR #151, opened 2020-04-17
Robot p12

ericvaandering:


kubernetes/rucio/install_rucio_helm3.sh
```diff
@@ -0,0 +1,55 @@
...
+# Filebeat and logstash
+#helm install --name logstash --values cms-rucio-logstash.yml,${INSTANCE}-logstash-filter.yaml stable/logstash
+#helm install --name filebeat --values cms-rucio-filebeat.yml  stable/filebeat
+
+# Create a job NOW to start setting the proxies. 
+kubectl delete job --ignore-not-found=true fts
+kubectl create job --from=cronjob/${DAEMON_NAME}-renew-fts-proxy fts
+
+# Label is key to prevent it from also syncing datasets
+kubectl apply -f dataset-configmap.yaml
```
vkuznet, line 33: Eric, can you educate me why for certain manifest files you use hehelmml/helm3 and for others like here you still use kubectl? I want to understand this fact since we're discussing to move our cmsweb deploy procedure to helm and I want to avoid supporting 2 different tools.

ericvaandering:
The kubectl things are just very simple config files which I didn’t get around to parameterizing. Oh, and that one is temporary and we change it often.

But you can see the cms-kubernetes repository is already setup for the one chart.

PR #158, opened 2020-04-23
Adjusting deploy scripts for cmsweb

muhammadimranfarooqi:


muhammadimranfarooqi:
@vkuznet 
There are following adjustments:
1) OI was getting error like in namespaces like default\r . So I included this manually and removed from grep.
2) put kube-eagle in monitoring namespace.
3) --dry-run is deprecated. New parameter is --dry-run=client

Best

PR #171, opened 2020-05-06
adds script to test validity of endpoints for cern-security

Panos512:


Panos512:
I just noticed that all the other scripts are bash scripts. I have already written that in Python but if we need a bash script I can gladly re-write it.

vkuznet:
Panos, I prefer either bash scripts or Go executables since they are environment independent. With python scripts someone needs to have proper (version) of python, e.g. python should be compiled with open SSL, etc. Do not underestimate the maintenance effort with the tools you add.

Therefore, if it is not too complicated, please convert python to bash+curl.

Panos512:
OK yes it makes sense @vkuznet . Thanks for the explanation. I converted it based on the already existing test_k8s.sh

PR #180, opened 2020-05-16
add indication of where log files are

belforte:


belforte:
I could not find any place where this is documented...

PR #181, opened 2020-05-17
uses existing containers of httpgo in service-mesh example

Panos512:
The service-mesh example was not properly working because referenced docker containers have been removed from docker hub.

### DIFF
--- a/kubernetes/cmsweb/service-mesh/httpgo.yaml
+++ b/kubernetes/cmsweb/service-mesh/httpgo.yaml
@@ -21,12 +21,12 @@ metadata:
     account: httpgo
 ---
 ###
-### Deployment using 1.0.1 version
+### Deployment using v1 version
 ###
 apiVersion: apps/v1 # istio version
 kind: Deployment
 metadata:
-  name: httpgo-1.0.1
+  name: httpgo-v1
   namespace: dev
   labels:
     app: httpgo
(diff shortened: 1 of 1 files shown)


muhammadimranfarooqi:
Panos,

Did it work for you? I already tried like this in the last week but was not working.

Panos512:
Hi Muhammad,
Yes, I was able to run it. It created the 2 pods (one for each version) and I made 3 curl requests which were split 2-1 so I guessed it was ok :) This is what I should observe right?

vkuznet:
In this case why do you use "HG2005d" which will be gone as well after new release will be created? If you want to use existing container just leave it without tag which will force to use latest one. So far I don't see any reason to keep specific tag in R&D phase.

Panos512:
Makes perfect sense Valentin, thanks! 
I have used the latest release in both `Deployments` just tagging them with a different version for Istio to use. I tested and it still works, doing a 50-50 balancing of requests.

vkuznet:
how did you test that it works, if you have identical images?

Panos512:
I checked the logs for the two pods that were created and the requests are split amongst them. Images might be the same but I would guess Istio gateway looks at the version label and accordingly distributes requests.

vkuznet:
ok, I just wanted to know how you did the tests, so I'm merging the PR.

PR #198, opened 2020-05-29
Update statsd-prometheus-mapping.yaml

dmielaikaite:
added new probes with tags


dmielaikaite:
@ericvaandering waiting for approval

PR #203, opened 2020-06-01
Do some cleanup as we no longer need anything with helm2

ericvaandering:


vkuznet:
Eric, I don't think you need my approvals for rucio PRs, feel free to merge at your convenience.

PR #204, opened 2020-06-02
Adding reqmgr2ms services

muhammadimranfarooqi:

PR #206, opened 2020-06-03
Adjusting manifest files for reqmgr2ms

muhammadimranfarooqi:


muhammadimranfarooqi:
@amaltaro 

Alan, this is new PR which is made according to the changes you suggested in PR [204](https://github.com/dmwm/CMSKubernetes/pull/204)

kubernetes/cmsweb/services/reqmgr2ms-output.yaml
```diff
@@ -1,25 +1,25 @@
...
-    app: reqmgr2ms-o
+    app: ms-output
   ports:
-    - port: 8248
-      targetPort: 8248
-      name: reqmgr2ms-o
+    - port: 8245
+      targetPort: 8245
+      name: ms-output
     - port: 18248
```
amaltaro, line 13: don't you want to update the port used by prometheus to collect service metrics as well?

kubernetes/cmsweb/services/reqmgr2ms-transferor.yaml
```diff
@@ -1,25 +1,25 @@
...
-    app: reqmgr2ms-t
+    app: ms-transferor
   ports:
-    - port: 8248
-      targetPort: 8248
-      name: reqmgr2ms-t
+    - port: 8247
+      targetPort: 8247
+      name: ms-transferor
     - port: 18248
```
amaltaro, line 13: same comment here. Shall we update the monitoring port as well?

muhammadimranfarooqi: @vkuznet 

Valentin, could you please comment on port number for monitoring.

vkuznet: We applied common convention to assign monitoring ports as 10000+port of the service. Therefore, if trasferor port is 8247 the monitoring port would be 18247, etc.

But please keep in mind that in order for Prometheus to start using new service metrics we also need to make changes to Prometheus configuration, see https://github.com/dmwm/CMSKubernetes/blob/master/kubernetes/cmsweb/monitoring/prometheus.yaml

So for all services we have appropriate section there, e.g. for DAS we have
```
      - job_name: 'das-service'
        static_configs:
            - targets: ['das.das.svc.cluster.local:18217']
```
As you can guess you need to create a section for each individual service using <srv>-service as a name, and put in targets list the service entry in a form
```
<servicename>.<servicenamespace>.svc.cluster.local:<port>
```
Therefore, I suggest that you adjust PR to add appropriate section for discussed services.

muhammadimranfarooqi:
@vkuznet 

I have adjusted this PR according to your suggestions. Please take a look. Thanks

amaltaro:
reqmgr2ms.yaml is no longer needed...

PR #212, opened 2020-06-06
Disable crons proxy,logarchive,reboot for k8s services

muhammadimranfarooqi:


docker/acdcserver/install.sh
```diff
@@ -72,5 +72,6 @@ sed -i -e "s#ServerMonitor/2.0#ServerMonitor-acdcserver#g" /data/srv/current/con
 # add proxy generation via robot certificate
 # disable workqueue/reqmon/couch on acdcserver pod
 crontab -l | egrep -v "workqueue|reqmon|couchdb|reboot" > /tmp/mycron
```
vkuznet, line 74: here you left out reboot, but define it in a line below, please clean-up accordingly.

docker/dqmgui/install.sh
```diff
@@ -68,6 +68,6 @@ done
 sed -i -e "s#ServerMonitor/2.0#ServerMonitor-dqmgui#g" /data/srv/current/config/admin/ServerMonitor
 
 # adjust crontab
-crontab -l | egrep -v "reboot|Proxy" > /tmp/mycron
```
vkuznet, line 71: Here please leave Proxy as we need to check if it is specifically used by DQM.

muhammadimranfarooqi:
@vkuznet 

Changes applied. Thanks

PR #214, opened 2020-06-09
Only start mongodb for ms-output; fix logic of config files

amaltaro:
This PR fixes a few commits recently pushed in, such as:
https://github.com/dmwm/CMSKubernetes/commit/0251c3a8cef30e0226c8818c51406e990a14873f
https://github.com/dmwm/CMSKubernetes/commit/5f195f62de26e5979401ac06dc63feba226f151c
https://github.com/dmwm/CMSKubernetes/commit/86f83f3fbca3a4571e54bec2d8f19dd3a18d232a

In short, it fixes:
* only start mongodb service when reqmgr2ms is running the `ms-output` component
* fix logic of deployment/services_config configuration files for reqmgr2ms (and dbsmigration), we were deleting all the config files every time it evaluated to true

@muhammadimranfarooqi @vkuznet please review. I will not manage to test these changes today/tomorrow.

### DIFF
--- a/docker/reqmgr2ms/run.sh
+++ b/docker/reqmgr2ms/run.sh
@@ -40,21 +40,25 @@ if [ -f /etc/secrets/hmac ]; then
     sudo chown $USER.$USER /data/srv/current/auth/$srv/header-auth-key
 fi
 
-# use service configuration files from /etc/secrets if they are present
+# In case there is at least one configuration file under /etc/secrets, remove
+# the default config files from the image and copy over those from the secrets area
 cdir=/data/srv/current/config/$srv
-# check if /etc/secrets area contain at least one dbs configuration files
 cfiles="config-monitor.py config-output.py config-transferor.py"
 for fname in $cfiles; do
     if [ -f /etc/secrets/$fname ]; then
-        cd $cdir
+        pushd $cdir
         rm $cfiles
-        cd -
+        popd
+        break
     fi
+done
+for fname in $cfiles; do
     if [ -f /etc/secrets/$fname ]; then
         sudo cp /etc/secrets/$fname $cdir/$fname
         sudo chown $USER.$USER $cdir/$fname
     fi
 done
+
 files=`ls $cdir`
 for fname in $files; do
     if [ -f /etc/secrets/$fname ]; then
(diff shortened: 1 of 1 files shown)


docker/dbsmigration/run.sh
```diff
@@ -42,19 +42,18 @@ fi
 
 # use service configuration files from /etc/secrets if they are present
 cdir=/data/srv/current/config/$srv
-# check if /etc/secrets area contain at least one dbs configuration files
 cfiles="DBSDef.py DBSMigrate.py DBSGlobalReader.py DBSGlobalWriter.py DBSPhys03Reader.py DBSPhys03Writer.py Migration.py"
+pushd $cdir
```
vkuznet, line 46: Alan, this PR should not change logic of another deployment, and the changes your provide are not correct. They will disallow the use case when we do not have these files in config area, i.e. when we want to run service with default (image provided) config files. Please revert the changes

docker/reqmgr2ms/run.sh
```diff
@@ -40,21 +40,20 @@ if [ -f /etc/secrets/hmac ]; then
...
 fi
 
-# use service configuration files from /etc/secrets if they are present
+### For this specific reqmgr2ms service, we need to replace the default
+# configuration files, by those provided under the proper services_config dir
 cdir=/data/srv/current/config/$srv
-# check if /etc/secrets area contain at least one dbs configuration files
 cfiles="config-monitor.py config-output.py config-transferor.py"
+pushd $cdir
+rm $cfiles
```
vkuznet, line 48: The logic is incorrect here. If you rm all files it means that your POD will not start the service if you k8s do not provide config files. You only need to remove them ONLY if at least one file is present in the k8s configuration area, and leave all files to allow default behavior.

docker/dbsmigration/run.sh
amaltaro: That's a good point. I didn't consider the use case where we want to run the service with the default image configuration. 
I can revert these changes, but note that your implementation will only work for a *single* configuration file provided in services_config. If someone wants to provide 2 or more files, and remove whatever comes in the image, it will not work.

docker/reqmgr2ms/run.sh
amaltaro: That's a valid use case, let me update it.

amaltaro:
Valentin, can you please have another look?
The DBS changes were removed (even though I pointed a flaw in the original code) and I changed reqmgr2ms to keep original configs in case nothing gets provided under /etc/secrets area

PR #215, opened 2020-06-09
Increase Memory requirements for reqmon and reqmon-tasks PODs

amaltaro:
Increase memory limit to 8GB.

I was going to increase the minimum required memory from `256Mi` to `1Gi`, but I'm not sure about the impact it will have on the test clusters...

@muhammadimranfarooqi @vkuznet is kubernetes smart enough to only schedule such PODs in minions providing up to `            memory: "8Gi"` memory? Otherwise, POD recreation will be more common than we expected.

### DIFF
--- a/kubernetes/cmsweb/services/reqmon-tasks.yaml
+++ b/kubernetes/cmsweb/services/reqmon-tasks.yaml
@@ -41,7 +41,7 @@ spec:
             memory: "256Mi"
             cpu: "300m"
           limits:
-            memory: "3Gi"
+            memory: "8Gi"
             cpu: "1000m"
         command:
         - /bin/bash
--- a/kubernetes/cmsweb/services/reqmon.yaml
+++ b/kubernetes/cmsweb/services/reqmon.yaml
@@ -61,7 +61,7 @@ spec:
             memory: "256Mi"
             cpu: "300m"
           limits:
-            memory: "3Gi"
+            memory: "8Gi"
             cpu: "1000m"
         ports:
         - containerPort: 8249


vkuznet:
Alan, please justify the increase in resources. We allocated them based on usage of services in VM based cluster.

k8s will schedule pod using resources and it will check if the resources fit into minion capacity.

amaltaro:
Sure, I checked those numbers but I forgot placing them in this PR.

cmsweb-testbed (vocms0731) is currently using 1.5GB
cmsweb (vocms0766) is currently using 4.5GB , however, during the recent very high load in the WM system, it went up to 8GB or so.

Can you please clarify whether POD scheduling always consider the upper limits too?

vkuznet:
Alan, to better understand the resource please read this doc:
https://cloud.google.com/blog/products/gcp/kubernetes-best-practices-resource-requests-and-limits
In particular, it says:

_Kubernetes then checks to see if the Node has enough resources to fulfill the resources requests on the Pod’s containers. If it doesn’t, it moves on to the next node._

_If none of the Nodes in the system have resources left to fill the requests, then Pods go into a “pending” state._

And, in k8s we use m2.2xlarge nodes which have 30GB of RAM. So you should plan accordingly, i.e. check how much resources used by other services. From my understanding, we have capacity to run processes with 8GB of RAM, but we should be careful with resource assignments to allow us to run **all** services smoothly.

amaltaro:
Thanks for the pointer, Valentin.

From the documentation (I did not watch that video), it looks like PODs are schedule based on the *requests*, not on the *limits*. IF my understanding is correct, then once we migrate to production, we would be scheduling 256MB PODs, and they would ramp up to a few GBs of memory footprint in the first minutes, possibly saturating the minion resources where it was scheduled to. In that case, I think we should bump
```
requests.memory: "256Mi"
```
to at least:
```
requests.memory: "1Gi"
```

What do you think?

vkuznet:
the request resources will be used by k8s to determine if pod can start-up, therefore you should appropriately setup the values based on your app requirement. Since I don't know exact numbers required for every single deployment I rely on developers who will provide them. I have no objection to increasing requests.memory value.

amaltaro:
Well, I have a dilemma here... For production services, it would make A LOT of sense to set it to `1Gi`, but then we might have problems scheduling it in dev nodes, where `256Mi` would be more than enough...

Perhaps once things are in production, it's a better moment to update those requirements(?)

vkuznet:
Alan, for that we certainly need move to templating supported by helm. But so far it is not the case (I want to do once thing at a time and introducing templates creates even another level of complexity in deployment procedure). Instead, we use a little trick like this:
https://github.com/dmwm/CMSKubernetes/blob/master/kubernetes/cmsweb/services/reqmgr2.yaml#L44
and then we replace prod prefix with proper value, see
https://github.com/dmwm/CMSKubernetes/blob/master/kubernetes/cmsweb/scripts/deploy.sh#L37-L43
https://github.com/dmwm/CMSKubernetes/blob/master/kubernetes/cmsweb/scripts/deploy.sh#L559

Therefore, I suggest that we do the same, i.e. you define request limits and add `### PROD <value>` for those values which will need to be changed in production environment. And, of course, we need to adjust our deploy script to add another sed substitution similar to line L559 above (of course, it should be properly done in all places of deploy script).

Feel free to create PR and/or coordinate with Muhammad how to do that.

Doing this way we will have proper limits for dev/preprod/prod k8s clusters.

PR #222, opened 2020-06-12
Fix dbsmigration k8s

muhammadimranfarooqi:


muhammadimranfarooqi:
@vkuznet , @yuyiguo 

Together with this change https://gitlab.cern.ch/cmsweb-k8s/services_config/-/blob/preprod/dbsmigration/Migration.py#L16 and having view_instances_k8s_migration.json file in /etc/secrets area. 

dbsmigration runs new k8s instance. However, it does not start automatically. When I login to the pod and check status is not running. 

However after running ./run.sh after login to the pod dbsmigration starts. Everything looks fine except this automatically starts of the service. 

Please guide.

vkuznet:
@muhammadimranfarooqi , did you inspect the pod log to see why DBS does not start automatically? The log reveals the following
```
k -n dbs logs dbsmigration-57cfd98d78-79pzx
....
/data/srv/current/config/dbsmigration/manage: line 94: /data/srv/state/dbsmigration/view_instances_k8s_migration.json: Permission denied
....
```
So, try to figure out what's wrong with permissions.

muhammadimranfarooqi:
@vkuznet 
Even after fixing permission issue in the latest commit, it does not start automatically. 

```
[mimran@lxplus765]~/CMSKubernetes-testbed/CMSKubernetes/kubernetes/cmsweb% kubectl -n dbs exec -it dbsmigration-57cfd98d78-fcjxt bash
kubectl exec [POD] [COMMAND] is DEPRECATED and will be removed in a future version. Use kubectl kubectl exec [POD] -- [COMMAND] instead.
[root@dbsmigration-57cfd98d78-fcjxt data]# /data/srv/current/config/dbsmigration/manage status
dbsmigration is NOT RUNNING.
[root@dbsmigration-57cfd98d78-fcjxt data]# ./run.sh
rm: cannot remove ‘DBSDef.py’: No such file or directory
rm: cannot remove ‘DBSMigrate.py’: No such file or directory
rm: cannot remove ‘DBSGlobalReader.py’: No such file or directory
rm: cannot remove ‘DBSGlobalWriter.py’: No such file or directory
rm: cannot remove ‘DBSPhys03Reader.py’: No such file or directory
rm: cannot remove ‘DBSPhys03Writer.py’: No such file or directory
/data
stopping dbsmigration
starting dbsmigration
crond: can't lock /var/run/crond.pid, otherpid may be 152: Resource temporarily unavailable
[root@dbsmigration-57cfd98d78-fcjxt data]# /data/srv/current/config/dbsmigration/manage status
dbsmigration is RUNNING, PID 452
```

muhammadimranfarooqi:
```
CherryPy Checker:
The Application mounted at '' has an empty config.
```

This is the only entry in the log file when it is deployed. When I run run.sh script, following logs come. 
```
[12/Jun/2020:16:14:31] ENGINE Error in 'start' listener <bound method Server.start of <cherrypy._cpserver.Server object at 0x7f883b700790>>
Traceback (most recent call last):
  File "/data/srv/HG2007a/sw/slc7_amd64_gcc630/external/cherrypy/5.4.0-comp2/lib/python2.7/site-packages/cherrypy/process/wspbus.py", line 203, in publish
    output.append(listener(*args, **kwargs))
  File "/data/srv/HG2007a/sw/slc7_amd64_gcc630/external/cherrypy/5.4.0-comp2/lib/python2.7/site-packages/cherrypy/_cpserver.py", line 168, in start
    ServerAdapter.start(self)
  File "/data/srv/HG2007a/sw/slc7_amd64_gcc630/external/cherrypy/5.4.0-comp2/lib/python2.7/site-packages/cherrypy/process/servers.py", line 170, in start
    wait_for_free_port(*self.bind_addr)
  File "/data/srv/HG2007a/sw/slc7_amd64_gcc630/external/cherrypy/5.4.0-comp2/lib/python2.7/site-packages/cherrypy/process/servers.py", line 438, in wait_for_free_port
    raise IOError("Port %r not free on %r" % (port, host))
IOError: Port 8251 not free on '127.0.0.1'

[12/Jun/2020:16:14:31] ENGINE Shutting down due to error in start listener:
Traceback (most recent call last):
  File "/data/srv/HG2007a/sw/slc7_amd64_gcc630/external/cherrypy/5.4.0-comp2/lib/python2.7/site-packages/cherrypy/process/wspbus.py", line 241, in start
    self.publish('start')
  File "/data/srv/HG2007a/sw/slc7_amd64_gcc630/external/cherrypy/5.4.0-comp2/lib/python2.7/site-packages/cherrypy/process/wspbus.py", line 221, in publish
    raise exc
ChannelFailures: IOError("Port 8251 not free on '127.0.0.1'",)
```

yuyiguo:
@muhammadimranfarooqi  @vkuznet 

I will look into details after my meeting.
I just want to comment two things that  I just spotted by a quick look.

1. https://gitlab.cern.ch/cmsweb-k8s/services_config/-/blob/preprod/dbsmigration/Migration.py#L9
shoud not be "k8s" . This should be hardcode for "k8s". If you do so, i t will broken when go to others.

2. Why do we have  view_instances_k8s_migration.json file in /etc/secrets area ? 

The secrets should be only for the secrets.

vkuznet:
Yuyi, I don't understand what you said in your 1 statement. You wrote:
```
... should not be "k8s" . This should be hardcode for "k8s".
```
which seems to me two opposite things. The file you refer resides in gitlab area for k8s deployment. Therefore it is k8s configuration file and I think it should have k8s part.

Now, the /etc/secrets area in k8s contains all configurations and secrets. We only use one area. The service configuration files are supplied via gitlab repository and we use another gitlab for secrets. This area also contains robot certificates etc. This makes deployment management easy since we only look-up one area which we mount as /etc/secrets in all k8s deployments.

Now, regarding the actual issue here. You have curl calls in your service manage script (which I think bad as is since curl may fail), but even that you download the JSON configuration and validate it with concrete name, see
https://github.com/dmwm/deployment/blob/master/dbsmigration/manage#L91-L102
(see view_instances.json at line 93), while in k8s deployment we need to use view_instances_k8s_migration.json file. 

I suggest that you revisit your deploy script and do proper adjustments. In particular, I suggest the following:
- remove all curl calls as they may fail and it will cause failure on start-up of the service
- only rely on files from provided configuration, if you will use $CFGDIR/view_instances.json then we can keep proper json in different configuration areas and your manage script will work across VM and k8s deployment
- if you will use curl, please add checks that the curl call succeed
- and, if you use curl please make sure to adapt to proper file names since you name them differently, in VM you seems to use view_instances.json in k8s you use view_instances_k8s_migration.json , i.e. your deploy script should properly handle different names for different deployments

vkuznet:
@muhammadimranfarooqi , could you please redeploy again dbsmigration but this time do not run manually run.sh since it only hides the problem. It seems to me that when you deploy the service it runs the service since port is already in use when you run run.sh a second time.

Just deploy the service and leave that we can have a look at what is broken.

muhammadimranfarooqi:
@vkuznet 

I have redeployed services after creating new cluster. Please get new config from gitlab repository. 

I have not run `run.sh` script now.

vkuznet:
So, to me everything is running fine:
```
[root@dbsmigration-57cfd98d78-9jrsw data]# ps auxwwww
USER         PID %CPU %MEM    VSZ   RSS TTY      STAT START   TIME COMMAND
root           1  0.0  0.0  11704  2792 ?        Ss   20:21   0:00 /bin/bash /opt/setup-certs-and-run/setup-certs-and-run.sh
root           8  0.0  0.0  81996  3824 ?        S    20:21   0:00 su -c cd /data && /data/run.sh --preserve-environment _dbsmigration
_dbsmig+       9  0.0  0.0  11700  2596 ?        Ss   20:21   0:00 bash -c cd /data && /data/run.sh
_dbsmig+      10  0.0  0.0  11704  2752 ?        S    20:21   0:00 /bin/bash /data/run.sh
_dbsmig+     135  1.0  0.1 1713948 53720 ?       Sl   20:21   0:28 python -u /data/srv/HG2007a/sw/slc7_amd64_gcc630/cms/dbs3-migration/3.13.2//lib/python2.7/site-packages/dbs/components/migration/StartUp.py --config=/data/srv/current/config/dbsmigration/Migration.py
_dbsmig+     136  0.0  0.0  18172  1892 ?        S    20:21   0:00 rotatelogs /data/srv/logs/dbsmigration/dbsmigration-%Y%m%d-dbsmigration-57cfd98d78-9jrsw.log 86400
_dbsmig+     139  0.0  0.0  11704  2712 ?        S    20:21   0:00 /bin/bash /data/cmsweb/bin/process_monitor.sh .*dbsmigration dbsmigration :18251 15
_dbsmig+     141  0.5  0.2 2144456 83516 ?       Sl   20:21   0:15 /usr/share/filebeat/bin/filebeat -path.home /usr/share/filebeat -path.config /etc/filebeat -path.data /var/lib/filebeat -path.logs /var/log/filebeat -c /etc/secrets/filebeat.yaml --path.data /tmp/filebeat/data --path.logs /tmp/filebeat -e
root         143  0.0  0.0 131448  7136 ?        S    20:21   0:00 sudo /usr/sbin/crond -n
root         158  0.0  0.0  22704  2700 ?        S    20:21   0:00 /usr/sbin/crond -n
root        2501  0.0  0.0  19652  1472 ?        Ss   21:01   0:00 /usr/sbin/anacron -s
root        2857  0.0  0.0  13568  3588 pts/0    Ss   21:06   0:00 /bin/bash
_dbsmig+    2948  0.0  0.0 225916  9540 ?        Sl   21:07   0:00 process_exporter -pid 2925 -prefix dbsmigration -address :18251
_dbsmig+    2949  0.0  0.0   4376   652 ?        S    21:07   0:00 sleep 15
root        2956  0.0  0.0  53368  3876 pts/0    R+   21:07   0:00 ps auxwwww

[root@dbsmigration-57cfd98d78-9jrsw data]# curl -v http://localhost:8251/
* About to connect() to localhost port 8251 (#0)
*   Trying 127.0.0.1...
* Connected to localhost (127.0.0.1) port 8251 (#0)
> GET / HTTP/1.1
> User-Agent: curl/7.29.0
> Host: localhost:8251
> Accept: */*
>
< HTTP/1.1 200 OK
< Date: Fri, 12 Jun 2020 19:08:43 GMT
< Content-Length: 22
< Content-Type: text/html;charset=utf-8
< Server: CherryPy/unknown
<
* Connection #0 to host localhost left intact

[root@dbsmigration-57cfd98d78-9jrsw data]# cat /data/srv/logs/dbsmigration/dbsmigration-20200612-dbsmigration-57cfd98d78-9jrsw.log
CherryPy Checker:
The Application mounted at '' has an empty config.

127.0.0.1 - - [12/Jun/2020:21:08:07] "GET / HTTP/1.1" 200 22 "" "curl/7.29.0"
127.0.0.1 - - [12/Jun/2020:21:08:43] "GET / HTTP/1.1" 200 22 "" "curl/7.29.0"
```

I don't know why you wrote that it is not running. Please explain.

muhammadimranfarooqi:
```
[root@dbsmigration-57cfd98d78-9jrsw data]# /data/srv/current/config/dbsmigration/manage status
dbsmigration is NOT RUNNING.
```
I was checking manage script to check service status.

vkuznet:
So, and under which account your running manage script? ANd, under which account dbsmigration process is running?

```
sudo -u _dbsmigration /data/srv/current/config/dbsmigration/manage status "'I did read documentation'"
dbsmigration is RUNNING, PID 135
```

Please pay closer attention within pod which account do you use and which account is used by the script/service.

To me everything is fine with a service.

muhammadimranfarooqi:
Ok. Got it now. I was running with root. Indeed everything is fine then. This PR is ready to go in. Thanks

Cheers

vkuznet:
ok, glad that on our side everything is fine.

Now, we need @yuyiguo to confirm the change. As I wrote on this ticket the deployment procedure in manage script is fragile and we need to know if we can use the proposed solution and use json files which can be stored in gitlab k8s configuration area.

yuyiguo:
@vkuznet @muhammadimranfarooqi 

"_which seems to me two opposite things. The file you refer resides in gitlab area for k8s deployment. Therefore it is k8s configuration file and I think it should have k8s part._"

I understood that it is for k8s, but we will have prod, k8s-dev in near future. All of them are going to use this file. Are we going to create three versions of this file and the only difference is VARIANT="k8s"/"prod"/"k8s-dev"? This will be a maintaining nightmare.

"_Now, the /etc/secrets area in k8s contains all configurations and secrets. We only use one area. The service configuration files are supplied via gitlab repository and we use another gitlab for secrets. This area also contains robot certificates etc. This makes deployment management easy since we only look-up one area which we mount as /etc/secrets in all k8s deployments._"

I really don't like the idea to put view_instances_k8s_migration.json or Migration.py  in /etc/secrets/. DBS secrets had not been changed for last 6-7 years. We recently added k8s db into it. This /etc/secrets should not be access unless necessary. When I logged in on cmsweb VMs, I could not see the secrets. I think we should do the same for k8s too. However, Migration.py and view_instances_k8s_migration.json are public info. They  may change with each deployment. So I prefer they will state in the /data/srv/current area. I don't care where you put them in /data/srv/current , but they belong to there.  It would be nice to have them in the same place instead of separating them. In addition, I think it is also desired that move view_instances_k8s_migration.json  into gitlab with Migration.py.  When we build, they both can be included in the rpm.

 "_Now, regarding the actual issue here. You have curl calls in your service manage script (which I think bad as is since curl may fail), but even that you download the JSON configuration and validate it with concrete name, see
https://github.com/dmwm/deployment/blob/master/dbsmigration/manage#L91-L102
(see view_instances.json at line 93), while in k8s deployment we need to use view_instances_k8s_migration.json file._"

Valentin, Please don't say this anymore. As I told you many times that I did not write any of these CMSWEB deployment scripts and I don't maintain them in the past and NOW. 

Thanks,
Yuyi

muhammadimranfarooqi:
@yuyiguo 

We have three branches in services_config repository https://gitlab.cern.ch/cmsweb-k8s/services_config. The three variants for k8s  such as `k8s`, `prod`, `k8s-dev` can be placed in `preprod`, `prod` and `test` branches respectively.  Each branch will have specific configuration files for its relevant deployment. 

Regarding `/etc/secrets`, all secrets and config files are first copied to `/etc/secrets/` area. Then the relevant config files from `/etc/secrets/` area are copied to `/data/srv/current/config/` area during deployment of docker images in the k8s. The `run.sh` script copies config files into config area. For example, for dbsmigration see here https://github.com/dmwm/CMSKubernetes/blob/master/docker/dbsmigration/run.sh#L44:L67
In this way, in the end, you'll find all config files in `/data/srv/current/config/dbsmigration/`. 

I have put `view_instances_k8s_migration.json` with `Migration.py` in gitlab repo. https://gitlab.cern.ch/cmsweb-k8s/services_config/-/tree/preprod/dbsmigration

This PR first copies `view_instances_k8s_migration.json` file from `/etc/secrets` into `/data/srv/state/dbsmigration` and then replaces `view_instances.json` string with `view_instances_k8s_migration.json` in the `manage` script when the pod starts. In this way, we end up using new view instances for k8s for `dbsmigration`. I hope this is clear now. 

Cheers

vkuznet:
Yuyi, since we maintain the scripts we agreed on the structure, please follow it. I'm accepting this PR and you'll need to provide all DBS configuration files in gitlab repository in different branches.

yuyiguo:
@muhammadimranfarooqi 

Thanks for putting dbsmigration view file into gitlab.  Could you please point me the script copy the config files into /etc/secrets?  Since  `view_instances_k8s_migration.json` is in gitlab, why not just change its name to view_instances.json instead of later to replace it in the `manage `?  Could you please rename the file to view_instance?  This will make code more readable and supportable for later.

yuyiguo:
@vkuznet 

I understood there is needs to have separate  three branches for cmsweb as a whole. DBS don't need the three branches, but I still follow the common structures. However, I would make the content in the three branches to be the same.   As you said you maintain the deployment scripts, you make the final decision.

muhammadimranfarooqi:
@yuyiguo 

When we deploy secrets, they are mounted at the mount point defined in the manifest file. For instance, you can see here. 

https://github.com/dmwm/CMSKubernetes/blob/master/kubernetes/cmsweb/services/dbs-global-r.yaml#L84

I tried with the same name, but in `manage` script this JSON URL https://github.com/dmwm/deployment/blob/master/dbsmigration/manage#L37 points to original view_instaces.json file in DBS repository and replace file again with the original one. Thats why I decided to use different name.

vkuznet:
@yuyi, let's clarify all pieces:
- neither you or us wrote deployment scripts in dmwm/deployemnt github repository, but someone needs to maintain them
- the deployment scripts are used in both infrastructures, the VM deployment and k8s one
- the DBSMigration manage script takes JSON from dbsmigration repository, see
https://github.com/dmwm/deployment/blob/master/dbsmigration/manage#L91
- the DBSMigration deploy script provides two different json files, the 
https://github.com/dmwm/DBS/raw/master/view_instances.json
and
https://github.com/dmwm/DBS/raw/master/view_instances_k8s_migration.json
So, these files are maintained/created by you. We follow the logic of deploy script
- we don't copy anything to `/etc/secrets` area, instead, during k8s deployment we **mount** some area to this path. This area we take from each service configuration area on gitlab. At the end whatever is present in gitlab will be used in k8s. If files are identical for different k8s deployment you'll commit them in different branches.

Since we need to maintain both deployments and you provided two different JSON files we need to keep them. If you can think about how to organize it better please let us know.

yuyiguo:
@muhammadimranfarooqi @vkuznet 

1. for k8s, which view file is used during the deployment? `https://gitlab.cern.ch/cmsweb-k8s/services_config/-/blob/preprod/dbsmigration/view_instances_k8s_migration.json` or `https://github.com/dmwm/DBS/blob/master/view_instances_k8s_migration.json`?

2. for the /etc/secrets, which areas are mounted to  different DBS pods? I did not see there are /etc/fstab files. I guess pods are different from regular OS.

3. Because VM and k8s are running at the same time and you made the k8s servers different from VM so we have to maintain two different configures. I am trying to figure out what is the best way to run the two system at the same time and the system is supportable in the  future. That is why I suggest that not use something like replace "A" to  "B" in the deployment scripts.

vkuznet:
Yuyi,
your items outline the main point I always make, the developers know better than operators what to supply to their code and how to run their code. So
1. you commit to both repository, and you should ensure that they contain proper config. I already explained that due to transition we need separate gitlab repository since github repository contains both deployment scripts and configuration which are embedded into the k8s images. Moreover, for every system I don't (literally) know which config is right or wrong, Muhammad put some config to gitlab and we asked all developers to x-check them and ensure that they are ok. 
2. pods are not OSes, pod is an instance of OS which run in an image which by itself runs in underlying host OS. Therefore, the pods are configured through their manifest files on top of normal Linux. We supply k8s manifest files where we can specify what to mount from external area to the pod. What you need to know and understand that via manifest files we can provide proper configuration to your service instances. WIthout that your service instance will use configs which are taken from github and which are by default used in VM cluster which may or may not work out of the box in k8s pod.
3. We strive to make consistent deployment, but VM is not k8s by definition. Some changes are required to make things work. We use the same deployment procedure via dmwm/deployment manage/deploy scripts. But we also provide ability to provide k8s configurations which can overwrite image defaults.

That's why we need developer's help to align properly deployment procedure because only developers know what is required for the service (even if we need to maintain the scripts).

If you need more detailed help with details of k8s deployment please say so that we can guide you directly through the process.

yuyiguo:
@vkuznet 

Can you just point me where is the DBS manifest files ? Even better, point out the mount config in the file.

I am making a new DBS release to remove the migration view in gouthub since it is already on gitlab.

PR #243, opened 2020-07-01
Adjusting Dockerfile for cmsweb

muhammadimranfarooqi:


muhammadimranfarooqi:
With this adjustment, the size is reduced from 3.03 G to 2.76GB

PR #255, opened 2020-07-16
Set more realistic resource requirements for WMCore services

amaltaro:
I compared the resource requirements from this yaml files with what is actually used in the CMSWEB backends, and I came up with these numbers (for a some cases, I added a safe margin).

I also think that alertscollector is no longer used: https://github.com/dmwm/CMSKubernetes/blob/master/kubernetes/cmsweb/services/alertscollector.yaml
and workqueue-tasks (AFAIR we put it all under workqueue, right Imran?): https://github.com/dmwm/CMSKubernetes/blob/master/kubernetes/cmsweb/services/workqueue-tasks.yaml

Please let me know if you want me to delete them as well in this PR.

@muhammadimranfarooqi @vkuznet please review
FYI @todor-ivanov @khurtado

### DIFF
--- a/kubernetes/cmsweb/services/couchdb.yaml
+++ b/kubernetes/cmsweb/services/couchdb.yaml
@@ -68,11 +68,11 @@ spec:
               - sudo chmod 0777 /data/srv/state/couchdb/database; sudo chown _couchdb:_couchdb /data/srv/state/couchdb/database
         resources:
           requests:
-            memory: "256Mi"
-            cpu: "500m"
+            memory: "512Mi"
+            cpu: "1000m"
           limits:
             memory: "3Gi"
-            cpu: "2000m"
+            cpu: "4000m"
         ports:
         - containerPort: 5984
           protocol: TCP
--- a/kubernetes/cmsweb/services/reqmon-tasks.yaml
+++ b/kubernetes/cmsweb/services/reqmon-tasks.yaml
@@ -48,7 +48,7 @@ spec:
           periodSeconds: 10
         resources:
           requests:
-            memory: "256Mi"
+            memory: "5Gi"
             cpu: "300m"
           limits:
             memory: "8Gi"
--- a/kubernetes/cmsweb/services/reqmon.yaml
+++ b/kubernetes/cmsweb/services/reqmon.yaml
@@ -67,6 +67,8 @@ spec:
           initialDelaySeconds: 300
           periodSeconds: 10
         resources:
+          # This service also relies on the DataCache thread, so the
+          # memory footprint can be very high
           requests:
             memory: "5Gi"
             cpu: "1"
--- a/kubernetes/cmsweb/services/reqmgr2-tasks.yaml
+++ b/kubernetes/cmsweb/services/reqmgr2-tasks.yaml
@@ -52,8 +52,8 @@ spec:
           initialDelaySeconds: 300
         resources:
           requests:
-            memory: "256Mi"
-            cpu: "300m"
+            memory: "512Mi"
+            cpu: "500m"
           limits:
             memory: "3Gi"
             cpu: "1000m"
--- a/kubernetes/cmsweb/services/reqmgr2.yaml
+++ b/kubernetes/cmsweb/services/reqmgr2.yaml
@@ -79,7 +79,7 @@ spec:
           periodSeconds: 10
         resources:
           requests:
-            memory: "256Mi"
+            memory: "512Mi"
             cpu: "300m"
           limits:
             memory: "3Gi"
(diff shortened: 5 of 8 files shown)


todor-ivanov:
Thanks @amaltaro !

kubernetes/cmsweb/services/reqmon.yaml
```diff
@@ -58,11 +58,11 @@ spec:
...
         resources:
           requests:
-            memory: "256Mi"
-            cpu: "300m"
+            memory: "5Gi"
+            cpu: "1000m"
           limits:
             memory: "8Gi"
-            cpu: "1000m"
+            cpu: "1100m"
```
vkuznet, line 65: Can reqmon use more than 1 core? If not, there is no need to assign 1100m here, you can up 1000m instead.

kubernetes/cmsweb/services/couchdb.yaml
```diff
@@ -67,11 +67,11 @@ spec:
...
         resources:
           requests:
-            memory: "256Mi"
-            cpu: "500m"
+            memory: "512Mi"
+            cpu: "1000m"
           limits:
             memory: "3Gi"
-            cpu: "2000m"
+            cpu: "4000m"
```
vkuznet, line 74: Do we use couchdb in k8s setup anymore? I thought that we separated them and put couchdb node separate from k8s. If so, there is no need to change anything here.

kubernetes/cmsweb/services/reqmon.yaml
amaltaro: No, it's meant to run in a single process. I'll set request and limit to the same value then (1000m).

kubernetes/cmsweb/services/couchdb.yaml
amaltaro: Yes, but in case we decide to deploy it again in k8s, we at least have a better hardware description.

kubernetes/cmsweb/services/reqmon.yaml
amaltaro: BTW, can't we have issues if a pod requires 1000m CPU, but also say that this is the limit?

muhammadimranfarooqi: @amaltaro 
No, there is no issue. I have checked this. The `requests` value is used to schedule the service if the resources are available. Once the service is scheduled, it may consume less than the `requests` value. The `limits` are used to make sure the pod never consumes greater than this value. Of course `limit` can never be lower than the `request` but these can be equal.

amaltaro: Thanks for the clarification, Imran. I'm going to update this PR

kubernetes/cmsweb/services/reqmon-tasks.yaml
```diff
@@ -38,7 +38,7 @@ spec:
         name: reqmon-tasks
         resources:
           requests:
-            memory: "256Mi"
+            memory: "5Gi"
```
vkuznet, line 41: Alan, how we jup from 256MI to 5Gi limit? This limited defines how much RAM required to start the up. Does reqmon and reqmon-tasks really need 5GI to start?

kubernetes/cmsweb/services/reqmon.yaml
```diff
@@ -58,8 +58,8 @@ spec:
         name: reqmon
         resources:
           requests:
-            memory: "256Mi"
-            cpu: "300m"
+            memory: "5Gi"
```
vkuznet, line 61: The same comments as with reqmon-stats. This limit defines how much RAM required to start the app. I doubt we need 5Gi for that.

kubernetes/cmsweb/services/reqmon-tasks.yaml
amaltaro: This is one of the issues we spotted this year (8GB of memory footprint).
The DataCache thread contains too much data and 5GB was the average of memory footprint a couple of weeks back when I logged into the cmsweb backends and checked the service's memory consumption.

kubernetes/cmsweb/services/reqmon.yaml
amaltaro: same comment as above (the service itself runs with the DataCache thread enabled, thus high memory consumption).

vkuznet: ok, could you please add an appropriate comment over there, otherwise we'll forget about it.

amaltaro: comment added.

amaltaro:
@vkuznet @muhammadimranfarooqi I have applied all the suggested changes. Shall we merge it now?

amaltaro:
I just noticed it hasn't ever been merged, and there are conflicts now. I will resolve them later today.

amaltaro:
Code updated. If resource requirements make sense to you, please merge it.

PR #266, opened 2020-07-27
Adjusting hpa for frontend

muhammadimranfarooqi:


muhammadimranfarooqi:
apiVersion: autoscaling/v2beta2  apis support multiple metrices. By this PR, both CPUs and memory based hps work at the same time. The hpa selects maximum value from these two metrices and adjust replicas accordingly.

PR #335, opened 2020-10-08
Fixing check-metric for dbs

muhammadimranfarooqi:


vkuznet:
@muhammadimranfarooqi , how about we'll switch again to http://promethes-service.monitoring.svc.cluster.local:8080
Now with the fix for check-metric we don't really need external hostname:port

vkuznet:
please try out with internal URI and see if it works then we'll use.

muhammadimranfarooqi:
Yes, worked. closing this PR

PR #347, opened 2020-10-21
Adding liveness probe crabserver

muhammadimranfarooqi:


muhammadimranfarooqi:
@belforte 

FYI.

muhammadimranfarooqi:
I tested this on a pod of crabserver `crabserver-57c98d8cb5-lsmsx`  and it restarted as can be seen here.
```
% kubectl get pods -n crab
NAME                          READY   STATUS      RESTARTS   AGE
crabcache-f84c5b796-sq8zz     1/1     Running     0          7h26m
crabserver-57c98d8cb5-lsmsx   1/1     Running     1          14m
crabserver-57c98d8cb5-szzdj   1/1     Running     0          14m
crabserver-57c98d8cb5-vqzps   1/1     Running     0          14m
crabserver-57c98d8cb5-xcrw8   1/1     Running     0          14m
crabserver-57c98d8cb5-xr5mc   1/1     Running     0          14m

**Comment by belforte** [comment]
congrats !

File: kubernetes/cmsweb/services/crabserver.yaml [hunk 1]
```diff
@@ -58,6 +58,16 @@ spec:
       #- image: sbelforte/crabserver:3.3.1911.rc3
       - image: cmssw/crabserver #imagetag
         name: crabserver
+        livenessProbe:
+          exec:
+            command:
+            - cmsweb-ping
+            - "--url=http://localhost:8270/crabserver/prod/info"
```
vkuznet, line 65: we should be careful here since in testbed I think we need /crabserver/preprod/info while in production it would be /crabserver/prod/info. Again we need template in place, but so far we may use substitute in deploy.sh script.

vkuznet:
Muhammad, I'll merge this PR, but we need another adjustment to handle preprod vs prod instances.

muhammadimranfarooqi:
@belforte 

Does it make any difference ? `/crabserver/preprod` and `/crabserver/prod` exists both in testbed and production clusters.

belforte:
currently we have configured both prod and preprod instance of crabserver to be able to talk both to prod and preprod oracle database instances ! Sorry for overloading the "instance" word here. We have been so confused ourselves that in the end making the "pre" not very relevant help.
So currently it will work. But I agree with Valentin that we'll need a better way. In particular the two Oracle DB instances are actually on different Oracle clusters (cmsr vs. devdb11) so testing one or the other matters. If crabserver can't talk to oracle this ping will fail, which all in all is good.

Intially I suggest to check `/crabserver/prod` everywhere for simplicity. This acutally means: check that the crabserver service running at the host in the URL can talk to Oracle PROD DataBase.  All calls to crabserver are handled via WMCore/REST/DatabaseRESTApi class. If DB access fails, everything tails.

muhammadimranfarooqi:
@belforte 

Thanks for the explanation.

belforte:
In the end the best way I found was to use two parameters to characterize one crabservice instance: RestHostName and DbInstance. I think the same will be needed e.g. for DBS which has even more DB instances. Keeping a 1:1 relation between host and dbInstance was the old CMSWEB way, but in this age of multiple clusters it was getting very confused, so I decided to make things explicit. Of course if someone thinks of a better way, I am all ears !!

vkuznet:
@muhammadimranfarooqi , the best way to deal with variety of service attributes is to define extra parameter to pass to deploy script which will specify a file where we may define different mapping for services, e.g. the file may contain
```
crabserver /crabserver/prod/info /crabserver/preprod/info
dbs /dbs/global /dbs/int
....
```
Then in all our manifest files we'll have production settings like /crabserver/prod/info or dbs/global and if necessary we can supply to deploy script a file which may be used to read necessary mapping changes we may need to apply. Then in deploy script, we'll know which service we use, and grep in provided file a mapping to use, if found we will apply it if not nothing will be done to k8s manifest file.

PR #348, opened 2020-10-22
Adjustings scripts/files for FE daemonsets

muhammadimranfarooqi:


vkuznet:
@muhammadimranfarooqi  I merged this PR but I noticed that you comment out deploy_ingress. The better way is to make deploy_ingress conditional based if ds is provided or not. This way we can have uniform deployment for both cases when we use nginx/ingress and when we use custom ds. Please make adjustments in next PR.

PR #350, opened 2020-10-23
Adding livenessprobe to cmsweb services

muhammadimranfarooqi:


muhammadimranfarooqi:
@vkuznet 

How can I test this for das-server, wmarchive and exitcodes? I mean is there any alternative commands similar to this? 

`sudo -H -u _das bashs -l -c '/data/srv/current/config/das/manage status'`

Thanks

kubernetes/cmsweb/services/das-server.yaml
```diff
@@ -42,6 +42,15 @@ spec:
         env:
         - name: X509_USER_PROXY
           value: /etc/proxy/proxy
+        livenessProbe:
```
vkuznet, line 45: @muhammadimranfarooqi it seems to me you duplicated livenessProbe in das-server which I added before and now it appears now twice. Pease check and fix.

kubernetes/cmsweb/services/wmarchive.yaml
```diff
@@ -60,6 +60,12 @@ spec:
         name: wmarchive
         args:
           - /data/run.sh
+        livenessProbe:
```
vkuznet, line 63: You duplicated livenessProbe in wmarchive which I added before, see line 76. Please check and fix.

vkuznet (changes requested):
And, there is no manage status command. If you want to test das-server or wmarchive, you can use plain curl against /das/status or /wamrchive/data/ end-points. curl with 200 OK tells you that they are fine.

muhammadimranfarooqi:
@vkuznet 

I have applied requested changes.  

Actually I wanted to ask the alternative of this command. 

`sudo -H -u _das bashs -l -c '/data/srv/current/config/das/manage stop "I did read documentation"'`

If I stop the process inside the pod, k8s kills the container and exits the user from the pod. How can we stop das-server without being killed. In this way, livenessprobe can be tested.

vkuznet:
SO, in case of das-server you may login to the pod
```
/ # ps
PID   USER     TIME  COMMAND
    1 root      0:00 {run.sh} /bin/sh /data/run.sh
    7 root      0:37 /data/das2go_exporter -address :18217
    8 root     36:26 /data/das2go -config /etc/secrets/dasconfig.json
   51 root      0:00 sh
   56 root      0:00 ps
/ # cat /data/run.sh
#!/bin/sh
mkdir -p /data/logs
nohup /data/das2go_exporter -address ":18217" 2>&1 1>& /data/logs/das2go_exporter.log < /dev/null &
/data/das2go -config /etc/secrets/dasconfig.json
```
So the main process is /data/run.sh which run exporter and das2go server. So, you may kill das2go process and I would expect that pod will not restart as exporter and run.sh are running.

muhammadimranfarooqi:
@vkuznet 

Actually, the container is killed with this. 

```
/ # ps aux
PID   USER     TIME  COMMAND
    1 root      0:00 {run.sh} /bin/sh /data/run.sh
    8 root      3:25 /data/das2go -config /etc/secrets/dasconfig.json
   45 root      0:00 sh
   53 root      0:00 ps aux
/ # kill -9 8
/ # command terminated with exit code 137
```

vkuznet:
@muhammadimranfarooqi , in this case the test can be performed by simply changing the end-point in livenessProbe, e.g. take `/das/bla-bla-bla` which does not exist and you'll get non 200 OK code.

PR #351, opened 2020-10-23
Adding livenessprobe to dbs

muhammadimranfarooqi:


vkuznet (changes requested):
Here we have the same problem as with crabserver. The DBS has its instance name in its URL, see dbs/int/global/.... The instance name in DBS is int/global, or prod/global, or int/dev Therefore we need to setup in manifest file the prod name and use substitute at the deployment on testbed or dev cluster. The substitute should be done I think via external mapping file.

muhammadimranfarooqi:
@vkuznet 

I have applied the requested changes.

PR #356, opened 2020-10-28
Adjusting livenessprobe

muhammadimranfarooqi:


vkuznet:
can we assign delay to match minute(s) interval, e.g. either use 60 or 120 (one or two minutes).

PR #367, opened 2020-11-09
Add Rumble docker image files and k8s configs

mrceyhun:
@vkuznet 
I added rumble files to related directories. Tested and working good. 
When I run resource consuming queries (like in 1 month of wmarhive data), I got out of memory error from Spark.
Because of that reason, I added cronjob to rerun Rumble server when I got Spark side error.


docker/rumble/Dockerfile
```diff
@@ -0,0 +1,34 @@
+FROM cern/cc7-base:20200504-1.x86_64
+MAINTAINER Ceyhun Uzunoglu ceyhunuzngl@gmail.com
+
+ADD hadoop.repo /etc/yum.repos.d/hadoop.repo
+
+# hadoop related RPMs
+RUN yum install -y cern-hadoop-config spark-bin-3.0 hadoop-bin-2.7.5 java-1.8.0-openjdk-devel gcc git
+RUN yum clean all
```
vkuznet, line 8: put this action to line above with `&& yum clean all`. The smaller number of layers we have the faster the docker image build is

docker/rumble/Dockerfile
```diff
@@ -0,0 +1,34 @@
...
+# Setup cron
+CMD ["crond", "-n", "&"]
+ADD cronjobs.txt $WDIR
+RUN crontab $WDIR/cronjobs.txt
+
+# add files
+ADD run.sh $WDIR
+ADD daemon.sh $WDIR
+RUN chmod 755 $WDIR/daemon.sh
+RUN chmod 755 $WDIR/run.sh
```
vkuznet, line 32: this can be combined with step above as `chmod 755 $WDIR/daemon.sh && chmod 755 $WDIR/daemon.sh`, again reducing number of steps reduces number of docker layers and its build time.

docker/rumble/client_scripts/rumble_client.py
```diff
@@ -0,0 +1,83 @@
+"""
```
vkuznet, line 1: Every python script should start with the following statement (choose python or python3)
```
#!/usr/bin/env python
```
This statement allows choice of proper python version and allow to run script as `/path/script.py`, i.e. withiout python command in any shell. Also make sure you change permission of the python file to have executable bit

vkuznet (commented):
I also suggest to rewrite rumble_client.py in Golang. Since it has simple functionality to make HTTP call and parse, Go will be much better choice since it will eliminate python dependencies in docker image. It will be faster, smaller and more portable.

mrceyhun:
@vkuznet  I added go client for rumble and edited code.

vkuznet (changes requested):
Everything looks fine now to me but I still want you to make few changes:
- please put all k8s related stuff into rumble namespace, this include nginx and rumble itself. So far you have them in default namespace. This is useful if we'll decide (this is my intention) to run it within CMS monitoring cluster. 
- for go code, please use `go fmt` (please read about it on internet, e.g. https://golang.org/cmd/gofmt/) which will format your code to standard Go formatting syntax. I use vim plugin for that but it can be done directly via go fmt command, please read `go doc cmd/gofmt`
- please adjust python code formatting as well, usually it uses 4 spaces for indentation, and you use only two. We follow pep8 guidelines, see https://www.python.org/dev/peps/pep-0008/#indentation

In general, if you use vim, you may use google vim-codefmt (https://github.com/google/vim-codefmt) to format your code in different languages.

mrceyhun:
@vkuznet applied all change requests. 
Service, deployment, secret, service in k8s "rumble" namespace.
Formatted go and python codes. I used to use Sublime (had not configured pep8 in new computer), now I'm using vim.

docker/rumble/scripts/go/rumble_go.go
```diff
@@ -0,0 +1,133 @@
...
+	if verbose > 0 {
+		log.Println("Response Headers:", resp.Header)
+	}
+	if resp.StatusCode == http.StatusOK {
+		write(resp, output)
+	}
+}
+
+func main() {
+	defaultUrl := "http://test-cluster-jnbxujghdusq-node-0.cern.ch/jsoniq?materialization-cap=-1&"
```
vkuznet, line 101: this is really bad default, I rather prefer to not having a default in this case. Plus default URL should not contains parameters. Please either use https://cms-monitoring.cern.ch as default URL or use empty string. Then, move end-point and parameters into function which will use default URL.

docker/rumble/scripts/python/rumble_client.py
```diff
@@ -0,0 +1,92 @@
...
+import os
+import json
+import argparse
+
+# third party libs
+import requests
+
+
+class OptionParser:
+	def __init__(self):
```
vkuznet, line 17: Again, the indentation here is non standard, you use now 7 white spaces, while it should be 4. Please fix the python code with pep8 guidelines.

vkuznet (changes requested):
Ceyhun, more changes are required, and I specified them in a code. But we also need to take out go, python and ipynb files and put them into CMSMonitoring repository. The CMSKubernetes should contain only docker and k8s files. The go and python code can be used outside of k8s and in this case it is better to keep it in code repository instead of kubenretes one. I suggest that you take out go code and put it into CMSMonitoring/src/go/Rumble, while you take out python/ipy code and put it into CMSMonitoring/src/python/Rumble areas. The example rumble scripts should also go to CMSMonitoring/doc/Rumble or CMSMonitoring/static/Rumble areas.  It is probably also good idea to move run.sh script into CMSMonitoring/scripts (and call it run_rumble.sh) such that it can be used w/o k8s as well.

Please rename rumble_go.go as rumble.go as well, and make it consistent with python, i.e. rename rumble_client.py to rumble.py (or call both as rumble_clieng.go and rumble_client.py)

Please create a separate PR for CMSMonitoring which we can merge before this one such that you can use all necessary pieces directly here.

Then you'll clone CMSMonitring repo in your dockerfile and use appropriate pieces to build the image.

mrceyhun:
@vkuznet 
Now I get run_rumble.sh script from CMSMonitoring repo. Docker build files and other scripts completely separated.
Tested and works fine.

PR #384, opened 2020-11-16
Add reqmgr2ms-ruleCleaner.yaml.

todor-ivanov:


muhammadimranfarooqi:
@todor-ivanov 

Please replace  `ms-rulecleaner` with `ms-ruleCleaner`

kubernetes/cmsweb/services/reqmgr2ms-ruleCleaner.yaml
```diff
@@ -0,0 +1,143 @@
...
+metadata:
+  labels:
+    app: ms-rulecleaner
+  name: ms-rulecleaner
+  namespace: dmwm
+spec:
+  selector:
+    matchLabels:
+      app: ms-rulecleaner
+  replicas: 1
```
vkuznet, line 43: Todor, we usually use `1 #PROD N` syntax here where N means how many replicas you'll need on production cluster. If you foresee that your service may require more then one replicas please specify via this syntax the max amount of replicas you may need.

kubernetes/cmsweb/services/reqmgr2ms-ruleCleaner.yaml
```diff
@@ -0,0 +1,143 @@
...
+        - mountPath: /host/etc/grid-security
+          name: etc-grid-security
+          readOnly: true
+        - name: setup-certs-and-run
+          mountPath: /opt/setup-certs-and-run
+#PROD#  - name: logs-cephfs
+#PROD#    mountPath: /data/srv/logs/reqmgr2ms
+        securityContext:
+          privileged: true
+#      initContainers:
```
vkuznet, line 118: This blocks shows how to start your service, if it needs to wait for another one, e.g. MongoDB, you may uncomment this block and specify which service it should wait for.

todor-ivanov:
Hi @muhammadimranfarooqi the service name is  `ms-rulecleaner` and can't follow the camelCase syntax conventions becuase it breaks the service deployment scripts. There is one place in the old perl code (do not ask me where exactly was it)  that is converting all service names to lower case and then the name `ms-ruleCleaner` becomes `ms-rulecleaner` and following that change everything was getting quite a mess.  I will could rename  the file name to `reqmgr2ms-rulecleaner` if it is what you are asking me to do, but then it will be in inconsistency with the `service-config` file names.

kubernetes/cmsweb/services/reqmgr2ms-ruleCleaner.yaml
todor-ivanov: Hi @vkuznet , Thanks for mentioning it. So far we do not foresee more than a single instance of this  service in prod as well as in preprod. I hope we may change this  value in the far  future in case  we decide to do so, but in a near future I do not see a way to make it run in more than a singe instance without significant changes in the code.

kubernetes/cmsweb/services/reqmgr2ms-ruleCleaner.yaml
todor-ivanov: Thanks again! The `MSRuleCleaner` service is meant to be totally independent locally. Meaning, it does not depend on any other service inside the POD. While it would be good if it could run after the `MSOutput` service since the `MSRuleCleaner` tries to contact it, so we avoid several  void iterations of the service while waiting for the other one  to start. But this is not a must. We have other means to protect ourselves from such situation.

todor-ivanov:
Thanks @vkuznet !

todor-ivanov:
Thanks @muhammadimranfarooqi !

PR #394, opened 2020-11-22
Adjusting frontend alerts monitoring

muhammadimranfarooqi:


muhammadimranfarooqi:
Hi @vkuznet 

I noticed there does not exist any job with `frontend-prod` . Despite a cpu being 0 for a couple of pods in prod, I did not get any notification. So adding rules for apache-cpuload.

PR #407, opened 2020-12-04
Adjusting filebeat configuration for FE

muhammadimranfarooqi:


docker/frontend/monitor.sh
```diff
@@ -16,17 +16,14 @@ filebeat.inputs:
   backoff: 5s
   max_backoff: 10s
   tags: ["frontend"]
-# disable internal monitoring, uncomment line below, default is true
```
vkuznet, line 19: I rather prefer to keep commented lines since they become handy when we start debugging since I don't need to scratch my head again to find these useful options. Please put them back.

vkuznet (changes requested):

PR #424, opened 2020-12-17
Rucio consistency monitor

ivmfnal:
Please add Rucio consistency monitor Docker and Kubernetes files


vkuznet:
Igor, sorry but you committed entire application to the CMSKubernetes, this is wrong. We only keep Docker and aux file used by docker in docker area, everything else it is better to keep in your own repository. I suggest that you take entire app area and put it in your own github repository under whatever name you choose. Then in your Docker file you'll clone the repository to build the image, e.g.
```
RUN git clone <your_github_here>
```
You can organize your own github repo as you like and have all test files there. The docker/rucio-con-mon should basically only contains your Dockerfile.

ivmfnal:
Valentin, yes, I was wondering about that.

Can I use my own Docker image from dockerhub instead ?

Igor
________________________________

vkuznet:
yes, of course you can use and maintain your own image. The only thing would be agreement on specific tag to use. So far it is sufficient but you'll need to provide stable tag at some point. I'll commit this PR for now.

ivmfnal:
If I use my own image from dockerhub, do I still need to add anything to the CMSKubernetes/docker ?

vkuznet:
no, we can use your docker image. This area only require if packages does not provide their images.

PR #426, opened 2020-12-17
Use rucio-cm namespace, removed docker/rucio-con-mon

ivmfnal:
I changed the namespace from "http" to "rucio-cm" and removed the Dockerfile


vkuznet:
Igor, can we eliminate dashes in namespace, please either use rucio or ruciocm.

ivmfnal:
Done

PR #427, opened 2020-12-17
Add service for ms-output mongodb

goughes:
Add a deployment for ms-output mongodb and expose as a service. My only open question on this one is if we need to persist the mongodb data directory somewhere. @amaltaro

### DIFF
--- a/kubernetes/cmsweb/services/ms-output-mongo.yaml
+++ b/kubernetes/cmsweb/services/ms-output-mongo.yaml
@@ -1,0 +1,73 @@
+kind: Service
+apiVersion: v1
+metadata:
+  name: ms-output-mongo
+  namespace: dmwm
+spec:
+  selector:
+    app: ms-output-mongo
+  ports:
+    - port: 8230
+      targetPort: 8230
+      name: ms-output-mongo
+    - port: 18230
+      targetPort: 18230
+      name: ms-output-mongo-mon
+---
+kind: Deployment
+apiVersion: apps/v1
+metadata:
+  labels:
+    app: ms-output-mongo
+  name: ms-output-mongo
+  namespace: dmwm
+spec:
+  selector:
+    matchLabels:
+      app: ms-output-mongo
+  replicas: 1
+  template:
+    metadata:
+      labels:
+        app: ms-output-mongo
+        env: k8s #k8s#
+# should be enabled once we'll have exporter running with this sercice
+# we should also add prometheus.io/port in that case
+#       annotations:
+#         prometheus.io/scrape: 'true'
+    spec:
+      securityContext:
+        runAsUser: 0
+        runAsGroup: 0
+      containers:
+      - image: cmssw/ms-output-mongo #imagetag
+        name: ms-output-mongo
+        tty: true
+        stdin: true
+        resources:
+          requests:
+            memory: "500Mi"
+            cpu: "2000m"
+          limits:
+            memory: "5Gi"
+            cpu: "4000m"
+        ports:
+        - containerPort: 8230
+          protocol: TCP
+          name: ms-output-mongo
+        - containerPort: 18230
+          protocol: TCP
+          name: ms-output-mongo-mon
+        volumeMounts:
+        - name: secrets
+          mountPath: /etc/secrets
+          readOnly: true
+#PROD#  - name: ms-output-mongo-data
+#PROD#    mountPath: /data/mongodb
+      volumes:
+      - name: secrets
+        secret:
+          secretName: ms-output-mongo-secrets
+#PROD#- name: ms-output-mongo-data
+#PROD#  persistenVolumeClaim:
+#PROD#      claimName: ms-output-claim


vkuznet:
Erik, PR is fine and I can merge. But your question about persistency of mongodb is valid. I think you should define a concrete path where you MongoDB data will be located. Then you'll need to complement this PR with persistent volume which will be mounted on the path you choose and will use in your mongodb configuration. Please see any other cmsweb services how to use persistent volume claim (PVC), e.g. 
https://github.com/dmwm/CMSKubernetes/blob/master/kubernetes/cmsweb/services/reqmgr2ms-transferor.yaml#L114-L115
https://github.com/dmwm/CMSKubernetes/blob/master/kubernetes/cmsweb/services/reqmgr2ms-transferor.yaml#L141-L143
but choose proper names

Then request from @muhammadimranfarooqi to create PVC and deploy it on testbed.

vkuznet:
@goughes Erik, I posted requested actions and still awaits you to complete those. I hope you didn't miss this and it is on your TODO list.

goughes:
Hi @vkuznet, yes this is on my list. Do we have options for persistent volumes (like ceph RBD) or is everything based on the ceph filesystem? I'm not sure what to make of the commented out lines. Do you manually put something else on those lines when you deploy in preprod vs. prod? Is the expectation that I just commit similar commented lines updated for my desired mount path and PVC name?

Thanks!

vkuznet:
You just need to add similar lines (but use proper path and names for PVC). Our deploy script strip out `#PROD#` prefix when we deploy services on prod/pre-prod clusters. This allows us to avoid PVC's in test clusters while have them in place on prod/pre-prod clusters. But in addition to those lines in your manifest we need actual storage yaml file, you can find examples in `storages` directory of CMSKubernetes/kubernetes/cmsweb repo. These files first create PVC which you can use in your service yaml file. Please coordinate create of PVC with @muhammadimranfarooqi for your needs. We need to know which type (cephfs or cinder) and how much space you need. The cephfs allows to cross-mount storage across pods/nodes (but we have limited quota on it), while cinder has lots of storage capacity but will only be mounted to a single node (no cross-mounts).

amaltaro:
> The cephfs allows to cross-mount storage across pods/nodes (but we have limited quota on it), while cinder has lots of storage capacity but will only be mounted to a single node (no cross-mounts).

this MSOutput microservice is meant to run as a single instance, but given that the amount of required storage is pretty small (likely < 2GB per year), I'd suggest to adopt CEPHFS to have the ability to cross-mount it, in case we decide to do so in the future. Could we have 25GB available for this microservice?

Valentin, Imran, another question, is there any large performance difference between CEPHFS and Cinder (read/write)? Do those technologies provide any (transparent) data replication, to avoid data losses?

muhammadimranfarooqi:
@amaltaro 

I have noticed performance issue in CEPHFS. I am not sure about Cinder since it is not being used in cmsweb. Yes, both cinder and cephfs, we can detach and attach to different cluster / nodes.

vkuznet:
@amaltaro both cephfs and cinder relies on CERN network. Their performance characteristics is defined by the type used in storage, e.g. critical storage or not. Therefore, I doubt there is a big difference between the two since both is network storage. Said that we used both in k8s, but originally had some problems with Cephfs which was resolved by CERN IT. I'm not sure what do you mean by transparent replication, it is a storage and any tool, e.g. rsync, will work with it. The only difference that Cephfs is cross-mounted while cinder is only single node mounted. But both can be remounted within a cluster.

amaltaro:
That's exactly my concern, given that it's a storage-as-service, over network, I'm trying to assess what are the chances of data loss in case one of the storage disks go away. 
I don't remember now if it is EOS storage in general, or if it's the "old" CEPH volumes we have attached for CouchDB backends and WMAgents that had a replication factor equals 2 (meaning, we would have to lose 2 disks to actually completely lose our data).
Are you able to grab this information from CERN IT?

Imran, once we have this information (even though I'm inclined to cephfs technology), we can then go ahead and allocate such volume to MSOutput in kubernetes.

@goughes Erik, please correct me and/or complement on anything else that I might be missing.

muhammadimranfarooqi:
We have reply from Ricardo,

```
cinder offers block storage and you'll need to manage the filesystem on top. it is usually used with single attach (only one instance attaching the volume at a time) but volumes can be detached/reattached - kubernetes does this for you, though it can be error prone sometimes. more info and the available storage profiles (iops, etc) here: https://clouddocs.web.cern.ch/details/block_volumes.html and https://clouddocs.web.cern.ch/containers/tutorials/cinder.html#kubernetes

cephfs offers a network accessible filesystem, which can be mounted multiple times and works very well with kubernetes. more info here: https://clouddocs.web.cern.ch/file_shares/README.html and https://clouddocs.web.cern.ch/containers/tutorials/cephfs.html#kubernetes. If you expect high load you might want to check with the CEPH team in advance (there's a note about performance in the first link)

depending on what database you want to use you can also check DBOD (https://dbod.web.cern.ch/) for a centrally managed solution. and for completeness and depending on the db you want/need you can also consider relying on a kubernetes operator (if there's one available for that db) which should handle replication, redundancy and backups
```

goughes:
Thanks for the information everyone. After some discussion with Alan, we decided to go with the cephfs storage type. Our storage needs are small for this service, so I think Alan's suggestion of 25 GB would be enough. Is it easy to add more storage in the future if necessary?

I took a look at the `storages` folder to see if I could figure out if I could just add my own yaml there, but after looking figured I should get some assistance from @muhammadimranfarooqi on that one. Can you help by adding the PVC? Then I will update the service to include mounting it.

muhammadimranfarooqi:
@amaltaro @goughes 

I have added ms-output cephfs shares i.e. 5GB for testbed and 25 GB for production clusters. The relevant PR is https://github.com/dmwm/CMSKubernetes/pull/433

The shares can be added in the manifest file with PVC name `ms-output-claim`

goughes:
Thanks @muhammadimranfarooqi! I have added the PVC and mount location to the yaml. Let me know if things look ok.

muhammadimranfarooqi:
@goughes 

You may add `#PROD#` prefix similar to this https://github.com/dmwm/CMSKubernetes/blob/master/kubernetes/cmsweb/services/reqmgr2ms-output.yaml#L141-L143

goughes:
Thanks, done!

Just for everyone's info, this mount point will contain both the db and log dirs for mongodb.

muhammadimranfarooqi (changes requested):
@goughes 

The indentation/spaces is not correct. Please make it similar to other files.

goughes:
@muhammadimranfarooqi Ok, see if that looks right to you.

muhammadimranfarooqi:
+1

PR #468, opened 2021-02-18
[cmsmonit-185] Create docker and k8s confs for condor cpu eff cronjob

mrceyhun:
@vkuznet 
- For now, I created new directories under kubernetes and docker.
- It needs 2 ports to speak with Analytix cluster (spark.driver.port and spark.driver.blockManager.port), so it requires NodePorts. 
- For now, condor-cpu-eff.yaml points to test output directory. I will change it to "/eos/home-c/cmsmonit/www/cpu_eff" if everything is fine.
- Namespace suggestion would be good, currently it uses default ns.


vkuznet:
@mrceyhun could you please use `hdfs` namespace since the workflow is run over there.

PR #474, opened 2021-02-23
Adding S3 configurations for all services in testbed

muhammadimranfarooqi:


kubernetes/cmsweb/s3/fluentd-conf.yaml
```diff
@@ -57,27 +57,113 @@ data:
...
-      aws_sec_key <sec_key> 
-      s3_bucket cmsweb
+      aws_key_id <aws_key_id>
+      aws_sec_key <aws_sec_key>
+      s3_bucket cmsweb-testbed
       s3_rerion cern
       s3_endpoint https://s3.cern.ch
       s3_endpoint_url https://s3.cern.ch
-      path cms-logs/
+      path frontend-logs/
```
vkuznet, line 68: Are you sure that path should point to only frontend-logs? I thought the previous path `cms-logs` was more correct? The fluentd will push logs to S3, right, and it should do this for all services. Please clarify.

muhammadimranfarooqi: The `frontend-logs` is directory name inside `cmsweb-testbed` bucket. Similarly, there are other directories like `dmwm-logs`, `das-logs` etc

PR #512, opened 2021-03-27
Add resource limit to condor-cpu-eff

mrceyhun:
@vkuznet  Optimum values are given as resource limits after a few tests.

### DIFF
--- a/kubernetes/condor-cpu-eff/README.md
+++ b/kubernetes/condor-cpu-eff/README.md
@@ -1,5 +1,7 @@
 ### Deploy
 
+`kubectl create namespace hdfs`
+
 `kubectl create secret generic condor-cpu-eff-secrets -n hdfs --from-file=secrets/condor-cpu-eff/keytab`
 
 `kubectl apply -f condor-cpu-eff.yaml`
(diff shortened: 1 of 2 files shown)


mrceyhun:
@vkuznet If this works fine, we may migrate all acronjobs of cmsmonit to k8s?

vkuznet:
@mrceyhun , even though I merged the PR, I didn't realize until now that it resides in kubernetes area. Please note, that under kubernetes area we put directories which contains manifest for entire cluster, like cmsweb or monitoring clusters. Please relocate  condor-cpu-eff under kubernetes/monitoring/services area where it belongs.

mrceyhun:
@Valentin Kuznetsov ***@***.***> Should I keep it as a folder
or keep just the yaml file? If yes and we deploy it to the monitoring
cluster, should I edit deploy.sh and deploy-secrets.sh?

vkuznet:
it depends how we'll deploy it. If we'll deploy it within monitoring cluster then you'll only need to move yaml file into services and adjust deploy script. But if we'll decide to deploy it into separate `spider` cluster then we'll create new dir for that cluster and setup there services and deploy script. So far I suggest to move it under monitoring. Once we'll gain experience how much resources it needs we can decide if we'll run it in separate cluster or not.

PR #513, opened 2021-03-29
Fixing das-server manifest file

muhammadimranfarooqi:


vkuznet:
ohh, thanks, those are leftover I had when I was testing it on testbed.

PR #530, opened 2021-04-09
Adjusting k8s-test env

muhammadimranfarooqi:


vkuznet:
Muhammad, I don't think it is correct solution. I suggested that we should properly setup k8s-testX env for individual testX cluster, e.g. if we deploy to test1 cluster the env should be k8s-test1, etc. For prod cluster it should be k8s-prod, for preprod it should be k8s-preprod. But if we deploy to yet another cluster it should be default k8s one.

muhammadimranfarooqi:
I have adjusted deploy-srv script with necessary adjustments

PR #559, opened 2021-04-29
Adjusting cmsweb-auth

muhammadimranfarooqi:


vkuznet:
Even though I merged this PR I also suggest to remove explicit tag for images since we should use `# imagetag` comment as well to supply the image tag at the deployment.

PR #586, opened 2021-05-24
Adjusting victoria-metrics for vmbackup

muhammadimranfarooqi:


kubernetes/monitoring/services/victoria-metrics.yaml
```diff
@@ -51,7 +51,17 @@ spec:
           volumeMounts:
           - name: cinder-volume
             mountPath: /tsdb
+        - name: vmbackup
+          image: muhammadimranfarooqi/vmbackup-utility:v00.00.02
```
vkuznet, line 55: I prefer to use cmssw image, so I created new PR for that, see https://github.com/cms-sw/cms-docker/pull/129 Once merged let's push image under cmssw repo on docker hub and adjust this line to use it form cmssw repo.

vkuznet (changes requested):

muhammadimranfarooqi:
I just noticed that the manifest file is in `services/agg` directory for VM agg cluster. So, adjusted accordingly.

PR #595, opened 2021-05-31
Adding vmrestore script

muhammadimranfarooqi:


kubernetes/monitoring/scripts/victoria-metrics-restore.sh
```diff
@@ -0,0 +1,64 @@
...
+#ls $tmpDir
+if [[ "$backup_service"  == "victoria-metrics" ]] ; then
+
+	echo "testing"
+  cat vmrestore.yaml | sed -e "s,cms-monitoring/vmbackup/2021/05/31/00:00,$bucket_path,g" > vmrestore.yaml.new
+  
+  curl -ksLO https://raw.githubusercontent.com/dmwm/CMSKubernetes/master/kubernetes/monitoring/services/agg/victoria-metrics.yaml
+
+  kubectl delete -f victoria-metrics.yaml
+  sleep 60s
```
vkuznet, line 31: I suggest to not rely on sleep as it is unreliable condition. Any service for whatever reason may take longer. Instead, please add a loop with some interval, e..g 10s and use kubectl to check if service/deployment is gone, then if it is the case you may exit the loop. Make loop finite too to avoid infinite conditions.

kubernetes/monitoring/scripts/victoria-metrics-restore.sh
```diff
@@ -0,0 +1,64 @@
...
+
+  kubectl apply -f vmrestore.yaml.new
+
+  sleep 60s
```
vkuznet, line 35: replace sleep with loop to check conditions of the service. We should use API to check if service is started, or even better kubectl logs to read certain pattern. Here again, make finite loop with some interval, check the conditions and exit appropriately.

For instance, here is what I see from VM startup:
```
2021-05-31T16:45:34.314Z        info    VictoriaMetrics/app/victoria-metrics/main.go:62 started VictoriaMetrics in 0.894 seconds
```
So, the pattern to check would be `.*info.*started VictoriaMetrics.*`

vkuznet (changes requested):
Please change all sleeps statements with for loop logic. The loop should make N iterations (make it finite) and either exit the script (with non zero exit code) or break the loop to continue to next logical step.

vkuznet (changes requested):
After each while loop you should check that action within a loop actually happen. If it never executed you should exit the script with non-zero exit status.

PR #597, opened 2021-06-07
Arrange dockerfile and k8s confs for rumble go server

mrceyhun:
@vkuznet 
Dockerfile and k8s confs are edited for rumble go server. 

I have one problem: I forked Rumble repo, changed a few lines and compiled this forked version to produce custom rumble jar. Docker image needs that JAR but it's size is 12Mb. I did not push it to this repo. Do we have a common place to store static files?
Adding compilation process to Dockerfile is another option but it will increase the image build time.


docker/rumble/run.sh
```diff
@@ -0,0 +1,18 @@
...
+principal=`klist -k "$keytab" | tail -1 | awk '{print $2}'`
+echo "principal=$principal"
+kinit $principal -k -t "$keytab"
+if [ $? == 1 ]; then
+    echo "Unable to perform kinit"
+    exit 1
+fi
+klist -k "$keytab"
+
+git clone https://github.com/mrceyhun/CMSMonitoring.git && \
```
vkuznet, line 14: why do you need to build rumble server in run.sh script?

vkuznet (changes requested):

vkuznet:
regarding static file, yes, I agree compiling it every time is not ideal. There are few places you can use, but it will require either invoke scp or HTTP to fetch the static content. You may use EOS, or gist. But the most convenient place will be put it in repo directly. You can try to gzip'ed though before the push to the repo and then only unzip it in place where it will be required.

docker/rumble/run.sh
mrceyhun: In order to get latest CMSMonitoring changes in each k8s deployment. When we build rumble_server.go  in Dockerfile, we need to rebuild docker image to get latest changes in CMSMonitoring.

mrceyhun:
@vkuznet I compressed with tar and zip, the best one reduced 12Mb to 11Mb. Current CMSKubernetes repo size is 35Mb. I'm close to your suggestion of putting it in cmsmonit EOS. What do you think?

vkuznet:
Ok, the jar is almost compressed. So, create a new repo, rumble_jar, and put only jar over there. Doing this way, we can only update a jar and keep code and jar in github, but in different repos.

docker/rumble/run.sh
vkuznet: No, it is wrong. If you need new code it means you should build new image. Moreover each build should come from a branch and not head of the master. Doing this way you can reproduce the behaviour of the image. Please remove code build from run script. Add a tag into Dockerfile and use it. See example in [here](https://github.com/dmwm/CMSKubernetes/blob/master/docker/das-server/Dockerfile). Then in CI/CD the tag can be substituted to a new one and everything can be automated, see [das CI/CD](https://github.com/dmwm/das2go/blob/master/.github/workflows/build.yml#L73)

mrceyhun: @vkuznet 
I removed go build from run script, tested and deployed my test cluster. If it's fine, I would like to postpone CI/CD implementation for a few days. I can open another PR for CI/CD.

vkuznet:
ok, what should I do with this PR then, should I wait or merge it?

mrceyhun:
Please merge. I'll create another PR for CI/CD.

PR #600, opened 2021-06-08
Bug fix rumble dockerfile, bump go to v1.6.5

mrceyhun:
fyi @vkuznet 
- Rumble jar is fetched from https://github.com/mrceyhun/rumble_jar
- Docker file and run.sh bugs are fixed


docker/rumble/Dockerfile
```diff
@@ -16,10 +15,10 @@ RUN yum install -y cern-hadoop-config spark-bin-3.0 hadoop-bin-2.7.5 git git-cor
 ENV PATH $PATH:/usr/hdp/hadoop-2.7.5/bin:/usr/hdp/spark3/bin
 
 WORKDIR $WDIR
-RUN curl -k -L -O https://golang.org/dl/go1.16.4.linux-amd64.tar.gz && \
-    tar xfz go1.16.4.linux-amd64.tar.gz && \
+RUN curl -k -L -O https://golang.org/dl/go1.16.5.linux-amd64.tar.gz && \
```
vkuznet, line 18: you should not download go tarball, instead start from go docker image like das-server, see https://github.com/dmwm/CMSKubernetes/blob/master/docker/das-server/Dockerfile#L1

docker/rumble/Dockerfile
```diff
@@ -16,10 +15,10 @@ RUN yum install -y cern-hadoop-config spark-bin-3.0 hadoop-bin-2.7.5 git git-cor
...
+    tar xfz go1.16.5.linux-amd64.tar.gz && \
     mkdir $WDIR/gopath && \
```
vkuznet, line 20: you don't need to setup gopath, instead create go.mod and go.sum in CMSMonitoring/src/go/Rumble which will keep all dependencies in place. Just visit this area and do:
```
go mod init github.com/dmwm/CMSMonitoring/src/go/Rumble
go mod tidy
```
Read about go.mod on internet!

docker/rumble/Dockerfile
```diff
@@ -29,5 +28,14 @@ RUN curl -k -L -O https://golang.org/dl/go1.16.4.linux-amd64.tar.gz && \
...
 ENV PATH $PATH:$GOROOT/bin:$WDIR
-# Set rumble jar location
+
+RUN git clone https://github.com/mrceyhun/rumble_jar.git && \
+    cp rumble_jar/spark-rumble-cmsmonit.jar $WDIR/spark-rumble.jar && \
+    rm -rf rumble_jar && \
+    git clone https://github.com/mrceyhun/CMSMonitoring.git && \
+    cp -R CMSMonitoring/src/go/Rumble/* $WDIR && \
+    rm -rf CMSMonitoring && \
+    go build rumble_server.go && \
```
vkuznet, line 38: When building go executable always build static one, it can be done using `CGO_ENABLED=0 go build ...`. Doing this way you may build go executable in one image and then copy it to another without worrying about dependencies (and dynamic libs).

vkuznet (changes requested):
I suggest the following:
- start from go builder image, e.g. like
`FROM golang:latest as go-builder` which will always guarantee latest go release
- make multi-stage build, see how it is done in other docker files which uses Go, e.g. das-server
- build static server executable
- copy it to another image which can have hadoop from CC7 image

docker/rumble/Dockerfile
mrceyhun: go.mod file is already in place under github.com/mrceyhun/CMSMonitoring/src/go/Rumble, PR is in review.

docker/rumble/Dockerfile
vkuznet: I suggest to use additional flags to build static (compressed) executable, e.g.
```
CGO_ENABLED=0 go build -o rumble_server -ldflags="-s -w -extldflags -static" rumble_server.go
```

vkuznet (changes requested):

muhammadimranfarooqi:
Hi @mrceyhun 

Do you need this PR or should it be closed or merged?

mrceyhun:
> Hi @mrceyhun
> 
> Do you need this PR or should it be closed or merged?

Hi @muhammadimranfarooqi , sorry this PR is idle since ankle fracture. I would like to merge this PR to not lost reviews of @vkuznet . I don't know when I will have time to improve this but last I remember is that a somehow working service.

PR #606, opened 2021-06-15
Fix Rucio urls in workqueue/reqmgr2ms just before running the service

amaltaro:
Fixes https://github.com/dmwm/WMCore/issues/10600

Just before we start the service via `run.sh` script, fixup the `rucio.cfg` file to make sure the correct default values are in place.
I took advantage and made the same fix for the `reqmgr2ms` service, which could potentially have the same faith (even though we do not initialize Rucio over there with default values...)

@muhammadimranfarooqi @vkuznet could you please review this? I have done some interactive tests and it seems to be working fine.

### DIFF
--- a/docker/reqmgr2ms/run.sh
+++ b/docker/reqmgr2ms/run.sh
@@ -85,6 +85,15 @@ for fname in $files; do
     fi
 done
 
+# before running the service, we need to make sure rucio.cfg has the correct URLs
+SERVICE_CONFIG=${cdir}/config-*.py
+RUCIO_CONFIG=${cdir}/etc/rucio.cfg
+MATCH_RUCIO_URL=`cat ${SERVICE_CONFIG} | egrep '^RUCIO_URL =' | awk -F'=' '{print $2}' | sed 's/"//g'`
+MATCH_RUCIO_AUTH_URL=`cat ${SERVICE_CONFIG} | egrep '^RUCIO_AUTH_URL =' | awk -F'=' '{print $2}' | sed 's/"//g'`
+# now replace it in the rucio.cfg file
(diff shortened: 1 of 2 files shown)


vkuznet (commented):
Alan, changes look fine to me, but you should double check the location of your configs. If they are in secrets area you can't change them. But if they are in another location where your account has proper rights then everything is fine.

My suggestion is to take you docker image, login to it and apply these changes. Doing this way you'll be sure that they are working. You can do it as following:
```
docker pull <your_image>
docker run --rm -h YOUR_HOST.cern.ch -i -t <your_image> /bin/sh
```
This will allow you to run your image on a node which has docker daemon. Then you can go to /data area, change your run.sh accordingly and start it.

vkuznet:
@amaltaro I only made my comment and will wait for your confirmation. If you think that everything is fine, please let me know, and I can merge the PR.

amaltaro:
@vkuznet Valentin, thanks for this hint. I have just tested it in both `ms-transferor` and `workqueue` production PODs.
From my side, it's good to go.

Once it's merged, do we need to build a new image to get this script in the image? Or it comes from a different/shared area?

vkuznet:
Merging, @muhammadimranfarooqi please create new image with these changes.

muhammadimranfarooqi:
@vkuznet @amaltaro 

Images created and deployed in production cluster.

amaltaro:
Thanks, Imran and Valentin! I confirm new images are working fine!

vkuznet:
@amaltaro , next time I suggest that you build and deploy your images. If you never try, you'll never learn. Stop the fear, it is very easy :)

amaltaro:
I agree! But it's not fear, it's just time and priorities management :-P

PR #612, opened 2021-06-24
Add parameters to reqmgr2ms/run.sh && Add more pods for MSUnmerged

todor-ivanov:
This PR is related to solving the WMCore issue [1], for scaling the MSUnmerged micro service, based on a predefined subsets of RSEs per pod.

[1]
https://github.com/dmwm/WMCore/issues/10599


todor-ivanov:
Hi @vkuznet can you please take a quick look to this one, just to confirm that the *.yaml files have/include anything needed for the 3 different flavors of the MSUnmerged to start. Thanks! 

FYI @amaltaro

docker/reqmgr2ms/run.sh
```diff
@@ -94,6 +148,12 @@ MATCH_RUCIO_AUTH_URL=`cat ${SERVICE_CONFIG} | egrep '^RUCIO_AUTH_URL =' | awk -F
 sed -i -e "s,rucio_host.*,rucio_host =${MATCH_RUCIO_URL},g" ${RUCIO_CONFIG}
 sed -i -e "s,auth_host.*,auth_host =${MATCH_RUCIO_AUTH_URL},g" ${RUCIO_CONFIG}
 
+# Edit the MSUnmerged configuration file (msConfig for the service) at runtime
+# based on the set of arguments passed to the run.sh script by the service
+# statrtup command defined in the .yaml file of the currently starting service
+[[ -n $rseExpr ]] && sed -e "s/^[[:blank:]]*RSEEXPR.*/RSEEXPR = \"${rseExpr}\"/g" ${cdir}/config-unmerged.py
```
amaltaro, line 154: Aren't you missing this `RSEEXPR` variable in the configuration files?

docker/reqmgr2ms/run.sh
```diff
@@ -1,4 +1,58 @@
...
+
+### Searching for the mandatory and optional arguments:
+# export OPTIND=1
+while getopts ":e:h" opt; do
+    case ${opt} in
+        e)
+            rseExpr=${OPTARG}
+            # escaping all possible special characters in the RSE expression:
+            # NOTE: One should always start escaping the escaping chars first
+            rseExpr=${rseExpr//\\/\\\\}
```
amaltaro, line 37: I'm afraid I do not follow this logic, but from your note, I assume you have extensively tested it.

amaltaro (commented):
These changes look good to me, Todor. However, I left a comment in the code on something that might be missing in the service configuration.

Just a note for future-us, as Dario would say, I believe this RSE distribution is not well balanced. The T1 pod (excluding FNAL) will likely have half or less work than the other two PODs.

docker/reqmgr2ms/run.sh
todor-ivanov: It is there in my personal machine ... unfortunately I cannot push my changes neither to github nor to gitlab `service_config` because none the previous PR related to the changes in `config-unmerged.py` file are yet merged.

docker/reqmgr2ms/run.sh
todor-ivanov: Hi Alan yes, I did test it. Because Rucio uses `\` as a set compliment operator it collides with it's shell purpose of an special character escape operator and because of that if start putting escape characters in front of the rest of the Rucios set operators `&` `|` first then you end up with random number of `\` operators in the string out of which you cannot tell the purpose of each other (was it an already existing Rucio set operator, was it a shell escape operator from the previous iteration). let me give you an example: 

If you do NOT escape the `\` operator first:
```
$ rseExpr='((cms_type=real&rse_type=DISK&(tier=2|tier=3))\country=US)'
$ rseExpr=${rseExpr//&/\\&}
$ rseExpr=${rseExpr//|/\\|}
$ rseExpr=${rseExpr//\\/\\\\}
]$ echo $rseExpr 
((cms_type=real\\&rse_type=DISK\\&(tier=2\\|tier=3))\\country=US)
```

If you do escape the `\` operator first: 
```
$ rseExpr='((cms_type=real&rse_type=DISK&(tier=2|tier=3))\country=US)'
$ rseExpr=${rseExpr//\\/\\\\}
$ rseExpr=${rseExpr//&/\\&}
$ rseExpr=${rseExpr//|/\\|}
$ echo $rseExpr 
((cms_type=real\&rse_type=DISK\&(tier=2\|tier=3))\\country=US)
```

todor-ivanov:
thanks for the review @amaltaro 

> The T1 pod (excluding FNAL) will likely have half or less work than the other two PODs.

I am open to suggestions .... even though I'd lett it run as a first iteration of manual load splitting and monitor it how would it go for a while and then we can reorganize the pods as we wish.... at the end this is a matter of a single `run.sh -e '$rsrExpr'` parameter change.

docker/reqmgr2ms/run.sh
amaltaro: Unless I'm missing your point, why can't you edit your PR/MR and add the RSEEXPR to the current changes? It looks much cleaner than making yet another 3 PRs to provide one more service configuration parameter.

docker/reqmgr2ms/run.sh
amaltaro: Thank you for these details, Todor. The escaping part is know to me, what I got lost on was the syntax of the normal slashes `/`, so this bash expression: `${rseExpr//&/\\&}`. I will google for that though ;)

amaltaro:
I see you're planning for 3 PODs, so a small variation of what I have already suggested would be:
POD 1: all T1s
POD 2: all T2/T3 US RSEs
POD 3: all the rest (thus, T2/T3 non-US RSEs)

But I agree that we can see how it goes and we can change it in the future, if needed.

amaltaro:
Another suggestion, maybe we should make the name of these yaml files more generic. Otherwise, if we change their RSE expression, it will likely be inconsistent with the yaml file name. Perhaps it could be named like `reqmgr2ms-unmerged-pod1.yaml` and so on. Just an observation though.

docker/reqmgr2ms/run.sh
todor-ivanov: No need to google @amaltaro, `man bash` tells everything. Here is the regular shell variable expansion section from the manual with the relevant pattern for this case:
```
 ${parameter/pattern/string}
              Pattern substitution.  The pattern is expanded to produce a pattern just as in pathname expansion, Parameter is expanded and the longest match of pattern against its value
              is  replaced  with  string.   The match is performed using the rules described under Pattern Matching below.  If pattern begins with /, all matches of pattern are replaced
              with string.  Normally only the first match is replaced.  If pattern begins with #, it must match at the beginning of the expanded value of parameter.  If  pattern  begins
              with %, it must match at the end of the expanded value of parameter.  If string is null, matches of pattern are deleted and the / following pattern may be omitted.  If the
              nocasematch shell option is enabled, the match is performed without regard to the case of alphabetic characters.  If parameter is @ or *, the substitution operation is ap‐
              plied to each positional parameter in turn, and the expansion is the resultant list.  If parameter is an array variable subscripted with @ or *, the substitution operation
              is applied to each member of the array in turn, and the expansion is the resultant list.
```

just a little help on the peculiarities for the example at hand: starting with double `//` tells the expansion engine to substitute the pattern everywhere it have been found not just the first occurrence. Then is the `\\` escape of the escape operator itself so that  the shell may avoid expanding the operator in the `${rseExpr//\\/\\\\}` statement itself. then comes the long chain of escapes `\\\\` which need accumulate levels of escapes based on the consequent steps down the road until it reaches the final destination i.e.:
   * first level of escapes is again the expansion of the current expression `${rseExpr//\\/\\\\}`, just as above
   * second level of escapes is for the expansion that happens for passing the variable to `sed`
   * third level of escape is the internal `sed` expansion, which cannot be avoided 
   * and the forth `\` is not an escape it is just the symbol itself (the Rucio set operator)

todor-ivanov: Sorry, you meant this expression : `${rseExpr//&/\\&}` ... but any way the logic and the sequence of `/` and escapes is exactly the same.

docker/reqmgr2ms/run.sh
todor-ivanov: Let me try fitting it in those commits ... may need to tell @muhammadimranfarooqi  to wait for a minute or two before merging though.

todor-ivanov:
hi @amaltaro 

> Otherwise, if we change their RSE expression, it will likely be inconsistent with the yaml file name.

That is correct too. It depends on us whether we like to have an expressive naming convention and substitute/rename the whole file every time we change the RSE expression or we want to have a generic naming convention and we just change the RSE expression every time we  want to re purpose the pod.

I am fine both ways. You chose.

todor-ivanov:
> POD 1: all T1s
> POD 2: all T2/T3 US RSEs
> POD 3: all the rest (thus, T2/T3 non-US RSEs)

This is an achievable schema yes..... Let me make the relevant changes.

amaltaro:
>That is correct too. It depends on us whether we like to have an expressive naming convention and substitute/rename the whole file every time we change the RSE expression or we want to have a generic naming convention and we just change the RSE expression every time we want to re purpose the pod.

@muhammadimranfarooqi Imran, does the POD gets named after the yaml file name? Or it's defined inside the yaml file?

docker/reqmgr2ms/run.sh
todor-ivanov: done

todor-ivanov:
Done

amaltaro:
This looks great to me, thanks, Todor.
There is just one remaining question - maybe you know the answer - which is how the pods will be named. It would be great to have the pod1 / pod2 / pod3 in the pod names, if possible.

amaltaro (approved):

vkuznet:
@todor-ivanov the changes are fine, but I have few suggestions:

- please avoid names like pod since it is the k8s acronym, the name should represent the service. Therefore, instead of `reqmgr2ms-unmerged-pod2.yaml ` you should use `reqmgr2ms-unmerged-tier2.yaml ` or something which will identify the service. This will allow operator to "know" what running pod means. In other words, the pod is an instance of the service, that's why it is a bad idea to put podId into the service yaml file. 
- When service is created the k8s will create a pod for you using your service name and not manifest file name. Therefore you should adjust your manifest to use proper name as well. So far you have
```
metadata:
  name: ms-unmerged
```
Therefore, you pods will be `ms-unmerged-<some-id>` and you actually want them to be called `ms-unmerged-tier2-<some-id>`. I suggest that you not only change manifest file name, but also adjust metadata attributes within manifest file to represents properly name of your service. Please note, that the metadata name should be consistent across manifest file since k8s uses name matching during deployment.
- There are two ways to run your service, the first approach you used in this line: ems to be complicated, e.g. instead of
``` 
   su  -c "cd /data && /data/run.sh -e '((cms_type=real&rse_type=DISK&(tier=2|tier=3))&country=US)' " --preserve-environment _reqmgr2ms
```
where you pass parameters to `run.sh` script. The second approach is to add multiple run scripts to your image, e.g. `run-tier1.sh`, `run-tier2.sh`, etc., which will implement logic how you want to run your service. This will give you more control and use better syntax. Each individual `run-tierX.sh` file may call common `run.sh` file with proper set of parameters. I don't have preference which approach is better, but if set of parameters become more and more complex, may be you may consider using second option.

vkuznet:
@todor-ivanov , could you please separate docker and k8s changes in two different PRs, it will allow to review and merge things easily. For instance, we can merge already docker part but still discuss how to make k8s one.

todor-ivanov:
Thanks @vkuznet !  I am renaming the yaml files and splitting the docker changes in  separate PR now . just give me few minutes and we can merge them both.

todor-ivanov:
Thanks @vkuznet I just did the requested change. The docker related changes are in a separate commit  now. The manifest files are not renamed in a way to reflect every service flavor. Also added in the metadata name of the instance. 

When it comes to the `run.sh` changes - I'd rather stick to the approach with variables for the time being. If we find it failing or generating troubles, then we may consider  multiple `run.sh` scripts too.

PR #615, opened 2021-06-25
Install gfal2-python3 dependecies

todor-ivanov:
Fixes https://github.com/dmwm/WMCore/issues/10613

This is a PR intended to bring the `gfal2-python3` dependency for the `reqmgr2ms` docker image. This is needed by `MSUnmerged` for using GRID services directly.

### DIFF
--- a/docker/reqmgr2ms-unmerged/Dockerfile
+++ b/docker/reqmgr2ms-unmerged/Dockerfile
@@ -1,4 +1,4 @@
-FROM reqmgr2ms:latest
+FROM cmssw/reqmgr2ms:HG2106g-v2
 MAINTAINER Valentin Kuznetsov vkuznet@gmail.com
 
 ENV WDIR=/data
--- a/docker/reqmgr2ms-unmerged/Dockerfile
+++ b/docker/reqmgr2ms-unmerged/Dockerfile
@@ -1,4 +1,4 @@
-FROM cmssw/cmsweb:20210621 AS reqmgr2ms
+FROM reqmgr2ms:latest
 MAINTAINER Valentin Kuznetsov vkuznet@gmail.com
 
 ENV WDIR=/data
@@ -15,6 +15,18 @@ USER ${USER}
 # start the setup
 RUN mkdir -p $WDIR
 WORKDIR ${WDIR}
+
+# Install gfal2-python3 dependecies
+RUN yum install -y \
+    gfal2-python3 \
+    gfal2-plugin-file \
+    gfal2-plugin-gridftp \
+    gfal2-plugin-http \
+    gfal2-plugin-srm \
+    gfal2-plugin-xrootd \
+    xrootd-client \
+    yum clean all && \
+    rm -rf /var/cache/yum
 
 # pass env variable to the build
 ARG CMSK8S
--- a/docker/reqmgr2ms-unmerged/README.md
+++ b/docker/reqmgr2ms-unmerged/README.md
@@ -1,7 +1,7 @@
 
 
 ### build image
-docker build -t cmssw/reqmgr .
+docker build -t cmssw/reqmgr2ms-unmerged .
 
 ### list images
 docker images
--- a/docker/reqmgr2ms-unmerged/README.md
+++ b/docker/reqmgr2ms-unmerged/README.md
@@ -1,0 +1,31 @@
+
+
+### build image
+docker build -t cmssw/reqmgr .
+
+### list images
+docker images
+
+### list of running containers
+docker ps --no-trunc -aq
+
+### remove all running containers
+docker rm -f `docker ps --no-trunc -aq`
+
+### run given image
+docker run --rm -h `hostname -f` -v /tmp/vk:/etc/secrets -i -t cmssw/reqmgr /bin/bash
+
+### remove existing image
+docker rmi cmssw/reqmgr
+
+### inspect running container
+docker ps # find docker id
+docker inspect <docker_id> | grep IPAddress
+
+### push image to docker.com
+docker push cmssw/reqmgr
+
+### references
+https://stackoverflow.com/questions/18497688/run-a-docker-image-as-a-container#18498313
+https://stackoverflow.com/questions/17236796/how-to-remove-old-docker-containers#17237701
+http://goinbigdata.com/docker-run-vs-cmd-vs-entrypoint/
--- a/docker/reqmgr2ms-unmerged/install.sh
+++ b/docker/reqmgr2ms-unmerged/install.sh
@@ -1,0 +1,91 @@
+#!/bin/bash
+
+ARCH=slc7_amd64_gcc630
+VER=HG2107a
+REPO="comp"
+AREA=/data/cfg/admin
+PKGS="admin backend reqmgr2ms"
+SERVER=cmsrep.cern.ch
+
+cd $WDIR
+git clone git://github.com/dmwm/deployment.git cfg
+mkdir $WDIR/srv
+
+cd $WDIR/cfg
+git reset --hard $VER
+
+# adjust deploy script to use k8s host name
+cmsk8s_prod=${CMSK8S:-https://cmsweb.cern.ch}
+cmsk8s_prep=${CMSK8S:-https://cmsweb-testbed.cern.ch}
+cmsk8s_dev=${CMSK8S:-https://cmsweb-dev.cern.ch}
+cmsk8s_priv=${CMSK8S:-https://cmsweb-test.web.cern.ch}
+sed -i -e "s,https://cmsweb.cern.ch,$cmsk8s_prod,g" \
+    -e "s,https://cmsweb-testbed.cern.ch,$cmsk8s_prep,g" \
+    -e "s,https://cmsweb-dev.cern.ch,$cmsk8s_dev,g" \
+    -e "s,https://\`hostname -f\`,$cmsk8s_priv,g" \
+    reqmgr2ms/deploy
+
+# Deploy services
+# we do not use InstallDev script directly since we want to capture the status of
+# install step script. Therefore we call Deploy script and capture its status every step
+cd $WDIR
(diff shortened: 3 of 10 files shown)


todor-ivanov:
Hi @vkuznet  here is the PR for the Docker image dependency from `gfal2-python3`

amaltaro:
@todor-ivanov @vkuznet how about we create a new docker image named something like `reqmgr2ms-unmerged`, which will be based on the reqmgr2ms image, and deploy the gfal tools on top of it? That way we can keep all these extra dependencies (and apparently a different python (3.6) version isolated to the MSUnmerged service only.

vkuznet:
@amaltaro that's another way to do it and I'm fine with that. But this implies that this PR is irrelevant. You need to decide how you want to proceed. Bottom line, the layers allow to reduce image size and copy necessary libraries from one image to another.

todor-ivanov:
ok, let me check if I can save this poor PR's life ;) If I can't .. .well I will create another one.

todor-ivanov:
@vkuznet in order to give it a try, could we merge this PR https://github.com/dmwm/CMSKubernetes/pull/612 so that I may change the docker image which `MSUnmerged`'s manifest refer to?

vkuznet:
ok, #612 is merged now

todor-ivanov:
Thanks @vkuznet

docker/reqmgr2ms-unmerged/Dockerfile
```diff
@@ -0,0 +1,43 @@
+FROM reqmgr2ms:latest
```
vkuznet, line 1: @todor-ivanov it is bad practice to start from latest since it does not allow to reproduce a potential issue with image since latest can change any time. I suggest that you apply specific tag of the image you start with. See how it is done for reqmgr2ms image which use specific tag of cmsweb one.

todor-ivanov:
Hi @vkuznet here is my try to achieve the following image dependency chain:
```
cmsweb -> reqmgr2ms -> reqmgr2ms-unmerged + gfal2 external dependencies
```

I hope I am not missing anything. I also reverted the changes to `run.sh` in the main `reqmgr2ms` image so that all the rest of the microservices continue rolling as they were. Please let me know in case I am missing something.

docker/reqmgr2ms-unmerged/install.sh
```diff
@@ -0,0 +1,91 @@
+#!/bin/bash
```
vkuznet, line 1: We found that bash may have vulnerabilities and if you don't really use bash specific features it is better to switch to simpler /bin/sh.

docker/reqmgr2ms-unmerged/monitor.sh
```diff
@@ -0,0 +1,33 @@
+#!/bin/bash
```
vkuznet, line 1: the same argument here, bash may not be required and simpler /bin/sh can be used. Doing this way bash may not need to be installed in an image at all.

docker/reqmgr2ms-unmerged/run.sh
```diff
@@ -0,0 +1,107 @@
+#!/bin/bash
```
vkuznet, line 1: the same argument here, may be you can replace bash with sh

docker/reqmgr2ms-unmerged/slc7-cernonly.repo
```diff
@@ -0,0 +1,9 @@
+[slc7-cernonly]
+name=CentOS-7 - CERN Only [20190324]
```
vkuznet, line 2: I don't know where you get it but it seems very outdated, e.g. cmsweb uses `name=CentOS-7 - CERN Only [20200928]`, and as such it is newer. I suggest that you take whatever latest from lxplus.

vkuznet (changes requested):

docker/reqmgr2ms-unmerged/Dockerfile
todor-ivanov: thanks @vkuznet I know how to put the tag . Can you save me some time digging and point me how to tag the `reqmgr2ms` or to take it's tag in case it is already done.

docker/reqmgr2ms-unmerged/install.sh
todor-ivanov: Based on my last night experience losing hours to debug shell versions - I am hardly inclined not to touch this one right now.

docker/reqmgr2ms-unmerged/monitor.sh
todor-ivanov: same reply as above

docker/reqmgr2ms-unmerged/run.sh
todor-ivanov: And here already I am 100% sure that I not only need bash, but also that If I could have a specific version of bash, more than half of the variable expansion hacking in the `run.sh` which took me hours of debugging, would not have happened at all :)

docker/reqmgr2ms-unmerged/slc7-cernonly.repo
todor-ivanov: this is a copy  from `reqmgr2ms` - nothing new. I am going to change it to the one that  is used in `cmsweb` though

docker/reqmgr2ms-unmerged/Dockerfile
vkuznet: All cmsweb images appear on docker hub, just use it to search for your favorite service and see which tags are there and what was the latest build, e.g. https://hub.docker.com/repository/docker/cmssw/reqmgr2ms

docker/reqmgr2ms-unmerged/slc7-cernonly.repo
vkuznet: end-to-end deployment approach implies that you manage your images as you like. If reqmgr2ms is outdated it is still DMWM group issue, I just pointed to outdated version based on provided PR.

todor-ivanov: done

docker/reqmgr2ms-unmerged/Dockerfile
todor-ivanov: super  ! Thanks! I just did it please  take a look at my latest commit

todor-ivanov:
Hi @vkuznet I made the changes requested (except the shell swap, which I consider for extremely dangerous on this stage). Please take another look. Thanks again!

docker/reqmgr2ms/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM cmssw/cmsweb:20210621
+FROM cmssw/cmsweb:20210621 AS reqmgr2ms
```
vkuznet, line 1: I"m not sure why did you make this change. Since you are not changing anything in reqmgr2ms image you don't need it

docker/reqmgr2ms-unmerged/monitor.sh
```diff
@@ -0,0 +1,33 @@
...
+
+# start process exporter
+for p in "config-monitor" "config-output" "config-transferor" "config-ruleCleaner"; do
```
vkuznet, line 4: It is unclear to me if this applies to this service. Does it have any of this configs? If not, the process monitor will not start, but if it is then the port may be identical to what is inside of this block for other services. Since you blindly copy-and-paste this script I want to make sure that you are not forgetting anything specific to this particular service.

docker/reqmgr2ms-unmerged/README.md
```diff
@@ -0,0 +1,31 @@
+
+
+### build image
+docker build -t cmssw/reqmgr .
```
vkuznet, line 4: If you copy-and-paste instructions then you should reflect them to the service you apply them to, i.e. in instruction document it uses `cmssw/reqmgr` but here your service is `cmssw/reqmgr2ms-unmerged`. Please adjust accordingly to avoid wrong instructions.

vkuznet (changes requested):
@todor-ivanov I made minor comments within necessary blocks, but I also want to point out here that you to a problem this PR will create. You'll use `cmssw/reqmgr2ms:HG2106g-v2` image to start with, but your install script uses `VER=HG2107a`, and every time the new version will be used you/cmsweb operator should remember that there are two different tags, one is used to start from specific image and another to use specific version of the deployment. I want to make sure that you understand this, and will keep this in mind for future updates.

docker/reqmgr2ms/Dockerfile
todor-ivanov: ok, removing it than

docker/reqmgr2ms-unmerged/monitor.sh
todor-ivanov: that is a correct statement - none of those configs are present for the serivce itself while the one that is for the current service `config-unmerged` is  actually missing in this list. fixing this one too.

docker/reqmgr2ms-unmerged/README.md
todor-ivanov: ok. fixing this one too

todor-ivanov:
Hi @vkuznet could you please  take another look. I am pretty sure thee could be also other things I may have forgotten. Thank sou again!

todor-ivanov:
BTW I just noticed this comment:
>You'll use cmssw/reqmgr2ms:HG2106g-v2 image to start with, but your install script uses VER=HG2107a, and every time the new version will be used you/cmsweb operator should remember that there are two different tags, one is used to start from specific image and another to use specific version of the deploymen

Yes, this is good that you point this out. 
So I used whatever was the latest tag uploaded in dockerhub for this service. And I am not sure if it have ever been build with any of the HG2107 tags yet. And I actually was not aware of this difference that we have here. Thanks you.

vkuznet:
merged, enjoy :)

todor-ivanov:
Thanks @vkuznet

PR #653, opened 2021-07-16
Update sqoop image tags

mrceyhun:
@vkuznet @leggerf 
New sqoop image and it's new secret deployed to cms-monitoring cluster.
To be compatible, I also deployed new image for "hdfs-exporter-eos" in all 3 monitoring clusters. 

Related with hdfs size issue in CMSMONIT-398 ticket.


mrceyhun:
I also updated spider images with new tag.

PR #657, opened 2021-07-19
Write cron logs to stdout

mrceyhun:
Write cronjob logs to stdout, so logs can be red from k8s pod logs.


mrceyhun:
@muhammadimranfarooqi could you review and merge if it's okay

PR #661, opened 2021-07-20
Adding bash in exporters

muhammadimranfarooqi:


vkuznet:
why do we need bash? Can we use sh instead?

vkuznet:
if we'll rely on sh then we can use alpine as a image (which is minimal and also secure).

muhammadimranfarooqi:
First, I tried without bash and was getting error for bash in `process_monitor.sh` script for configuring exporter as sidecar.

vkuznet:
so, instead of jumping to change docker file, just fix process_monitoring.sh to work with plain sh instead of bash. Then you can safely switch in dockerfile from bash to sh.

PR #662, opened 2021-07-20
Removing exporter from wmarchive

muhammadimranfarooqi:


muhammadimranfarooqi:
@vkuznet 

```
Starting: process_exporter -pid=13 -prefix wma_exporter
time="2021-07-20T18:48:36Z" level=info msg="Starting Server: :18200" source="process_exporter.go:415"
time="2021-07-20T18:48:39Z" level=error msg="ERROR: exit status 1" source="process_exporter.go:309"
Found existing process_exporter:    41 root      0:00 process_exporter -pid 13 -prefix wma_exporter -address :18200
time="2021-07-20T18:48:54Z" level=error msg="ERROR: exit status 1" source="process_exporter.go:309"
Found existing process_exporter:    41 root      0:00 process_exporter -pid 13 -prefix wma_exporter -address :18200
```
I guess this is exptected behaviour of exporter. First time, exporters are started and then after interval there are errors?

muhammadimranfarooqi:
@vkuznet 
Plz review. Thanks

vkuznet:
@muhammadimranfarooqi , even though the PR is ok, the exporter behavior is not ok. The error is real and if you'll inspect process_exporter.go code you'll see that it comes from `top` function. So, you need to add to alpine the Linux `top` function which process exporter will use. Please see if alpine provides it, if not then we should switch from alpine to `debian:stretch` image and only install additional `top` utility. Please investigate and make sure that the exporter error will be gone.

muhammadimranfarooqi:
@vkuznet 

I checked from inside exporter pod and top is working. 

```
Mem: 7034392K used, 413984K free, 4912K shrd, 80K buff, 4520564K cached
CPU:   4% usr   1% sys   0% nic  93% idle   0% io   0% irq   0% sirq
Load average: 0.26 0.31 0.46 2/1036 35825
  PID  PPID USER     STAT   VSZ %VSZ CPU %CPU COMMAND
   38    16 root     S     696m  10%   3   0% process_exporter -pid 11 -prefix wma_exporter -address :18200
   11     6 root     S     694m   9%   3   0% /data/wmarchive -config=/data/wmarch_go.json
  578     0 root     S     1688   0%   0   0% sh
35789     0 root     S     1660   0%   3   0% sh
   16     0 root     S     1604   0%   0   0% {process_monitor} /bin/sh /data/process_monitor.sh .*wmarch_go.json wma_exporter :18200 15
    6     0 root     S     1592   0%   1   0% {run.sh} /bin/sh /data/run.sh
35796 35789 root     R     1592   0%   0   0% top
35825    16 root     S     1580   0%   0   0% sleep 15
    1     0 root     S     1020   0%   1   0% /pause
```

muhammadimranfarooqi:
But -p flag is not there in top for alpine. 

```
top -b -n 1 -p 11
top: unrecognized option: p
BusyBox v1.33.1 () multi-call binary.

Usage: top [-bmH] [-n COUNT] [-d SECONDS]

Provide a view of process activity in real time.
Read the status of all processes from /proc each SECONDS
and display a screenful of them.
Keys:
        N/M/P/T: show CPU usage, sort by pid/mem/cpu/time
        S: show memory
        R: reverse sort
        H: toggle threads, 1: toggle SMP
        Q,^C: exit
Options:
        -b      Batch mode
        -n N    Exit after N iterations
        -d SEC  Delay between updates
        -m      Same as 's' key
        -H      Show threads
```

muhammadimranfarooqi:
With debian:stretch and cern base image, there is syntax error here
https://github.com/dmwm/cmsweb-exporters/blob/master/process_monitor.sh#L44

For example, instead of pid, it returns user. 

```
ps auxw | grep ".*wmarch_go.json" | grep -v grep | grep -v process_exporter | grep -v process_monitor | grep -v rotatelogs | tail -1 | awk '{print $1}'
root
```
So, either need adjustment in code for debian/centos or for alpine to resolve these errors.

muhammadimranfarooqi:
I have adjusted exporters to debian:stretch. Now it is working fine. 

```
Starting: process_exporter -pid=12 -prefix wma_exporter
time="2021-07-21T06:44:32Z" level=info msg="Starting Server: :18200" source="process_exporter.go:415"
Found existing process_exporter: root          40  0.0  0.1 712612  8712 ?        Sl   06:44   0:00 process_exporter -pid 12 -prefix wma_exporter -address :18200
```

The relevant PR is https://github.com/dmwm/CMSKubernetes/pull/664
However,  debian:stretch has bash by default.

vkuznet:
well, the bash vs sh, you can always uninstall stuff if necessary in any distribution. Therefore, to make decision to use or not bash you simply need to inspect new image for vulnerabilities. If bash is still listed there, just uninstall it in debian:stretch, and use sh instead. If it is not the case you can proceed with bash.

PR #666, opened 2021-07-21
Add dedicated ports to every service instance.

todor-ivanov:


muhammadimranfarooqi:
@todor-ivanov 

Please also adjust service and deployment names to `ms-unmerged-T1` , `ms-unmerged-T2-T3-US` and `ms-unmerged-T2-T3-noUS`

todor-ivanov:
@muhammadimranfarooqi please let me now to which lines should the name be changed. because there are way to omany fieleds called 'name` and lat tie when we discussed that  IIRC we decided that  those  that we already have should be the bare minimum.

muhammadimranfarooqi:
@todor-ivanov 

You can adjust all occurrences.

PR #686, opened 2021-07-28
Add vmalerts services

mrceyhun:
- Add Cluster Ip to listen vmalerts internal metrics, see [vmalert ref1](https://docs.victoriametrics.com/vmalert.html#web), [-httpListenAddr](https://docs.victoriametrics.com/vmalert.html#configuration)
- Fix alertmanager port

@muhammadimranfarooqi could you review?
fyi @vkuznet

### DIFF
--- a/kubernetes/monitoring/services/vmalert-1h.yaml
+++ b/kubernetes/monitoring/services/vmalert-1h.yaml
@@ -1,4 +1,18 @@
 ---
+apiVersion: v1
+kind: Service
+metadata:
+  name: vmalert-1h
+spec:
+  type: ClusterIP
+  selector:
+    app: vmalert-1h
+  type: ClusterIP
+  ports:
+  - port: 31880
+    protocol: TCP
+    name: http
+---
 apiVersion: apps/v1
 kind: Deployment
 metadata:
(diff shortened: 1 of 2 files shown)


vkuznet:
@mrceyhun , and why do we need an HTTP service for vmalert?

mrceyhun:
@vkuznet AFAIU from the vmalert [documentation](https://docs.victoriametrics.com/vmalert.html#web), the metrics `vmalert_remotewrite_sent_rows_total` and `vmalert_iteration_total` which VM developers asked in [issue](https://github.com/VictoriaMetrics/VictoriaMetrics/issues/1499#issuecomment-887717475) are served from `http://<vmalert-addr>/metrics` and we need to expose `-httpListenAddr` port. `curl/yum/...` commands are not exist in`victoriametrics/vmalert` docker image, so I cannot get these metrics via `curl http://localhost:8880/metrics` inside the pod. 

We need this only for debugging purpose of course. Btw, they are already exposing that HTTP port(i.e. [dockerfile](https://hub.docker.com/layers/victoriametrics/vmalert/v1.63.0/images/sha256-a345814e83af575418ae7d702cc0ff6168e812aaf6b56f12124a0cef3346ad6d?context=explore)) and serving internal metrics in `http://<vmalert-addr>/metrics`

PR #696, opened 2021-07-30
Adding helm templates for crab namespace.

arooshap:


crabcache/templates/configMap2.yaml
```diff
@@ -0,0 +1,14 @@
...
+  labels:
+    app: crabcache
+  namespace: crab
+data:
+  setup-certs-and-run.sh: |
+    #!/bin/bash
+    cp /host/etc/grid-security/* /etc/grid-security
+    echo 'INFO Files in /etc/grid-security'
+    ls -lahZ /etc/grid-security
+    su  -c "cd /data && /data/run.sh" --preserve-environment  _crabcache
```
muhammadimranfarooqi, line 14: We don't need extra file for configmap. Can be included in a single file after `---` lines.

crabcache/templates/deployment.yaml
```diff
@@ -0,0 +1,182 @@
...
+        image: "{{ .Values.image.repository }}:{{ .Values.image.tag | default .Chart.AppVersion }}"
+        imagePullPolicy: {{ .Values.image.pullPolicy }}
+        ports:
+        - name: {{ .Values.service.name }}
+          containerPort: {{ .Values.service.port }}
+          protocol: {{ .Values.service.protocol }}
+        - name: {{ .Values.serviceMon.name }}
+          containerPort: {{ .Values.serviceMon.port }}
+          protocol: {{ .Values.serviceMon.protocol }}
+        livenessProbe:
```
muhammadimranfarooqi, line 57: livenessprobe will be in exporters image only if it uses `cmsweb-ping` command.

crabcache/templates/deployment.yaml
```diff
@@ -0,0 +1,182 @@
...
+        - name: crabcache-storage
+          mountPath: /data/srv/state/crabcache/files
+        - name: logs
+          mountPath: /data/srv/logs/crabcache
+      - name: {{ .Values.image2.name}}
+        image: {{  .Values.image2.path }}
+        args: [
+          "-c", "/etc/filebeat.yml",
+          "-e",
+         ]
```
muhammadimranfarooqi, line 91: `args` values should come from `values.yaml` file

crabcache/templates/deployment.yaml
```diff
@@ -0,0 +1,182 @@
...
+        hostPath:
+          path: /etc/grid-security
+      - name: setup-certs-and-run
+        configMap:
+          name: crabcache
+{{- if or .Values.environment.prod .Values.environment.preprod }}
+      - name: crabcache-storage
+        persistentVolumeClaim:
+{{- if .Values.environment.preprod }}
+            claimName: logs-cephfs-claim-preprod-crab
```
muhammadimranfarooqi, line 157: claimName is `crabcache-claim` for both preprod and prod for `crabcache-storage`

crabcache/values.yaml
```diff
@@ -0,0 +1,151 @@
...
+
+image2:
+  name: crabcache-filebeat
+  path: docker.elastic.co/beats/filebeat:7.12.0
+
+image2resources:
+  requests:
+    memory: "50Mi"
+    cpu: "50m"
+livenessProbe:
```
muhammadimranfarooqi, line 85: livenessprobe should be in exporters only here.

muhammadimranfarooqi: @arooshap 

Apply same suggestions for crabserver as well.

crabserver/templates/deployment.yaml
```diff
@@ -0,0 +1,180 @@
...
+metadata:
+  name: {{ include "crabserver.fullname" . }}
+  labels:
+    {{- include "crabserver.labels" . | nindent 4 }}
+  namespace: crab
+spec:
+  {{- if not .Values.autoscaling.enabled }}
+  {{- if or .Values.environment.prod .Values.environment.preprod}}
+  replicas: 5
+  {{- end}}
```
muhammadimranfarooqi, line 12: Put else here with if condition. By default replica is 1 from values.yaml file, for prod and preprod it is 5.

crabserver/values.yaml
```diff
@@ -0,0 +1,186 @@
...
+        - sudo chmod 0777 /data/srv/state/crabcache/files; sudo chown _crabcache:_crabcache /data/srv/state/crabcache/files
+
+livenessProbe:
+  exec:
+    command:
+      - cmsweb-ping
+      - "--url=http://localhost:8271/crabcache/info"
+      - "--authz=/etc/hmac/hmac"
+      - -verbose
+      - "0"
```
muhammadimranfarooqi, line 75: I think this is not need anymore because it is there in exporter.

PR #709, opened 2021-08-05
Fixing ms-unmerged rules

muhammadimranfarooqi:


amaltaro:
@muhammadimranfarooqi this seems to fix the query string `rse_type`. But I fail to see where the fix for the `info` API is, since all those paths are hard-wired to: `/ms-unmerged/data/status`.

muhammadimranfarooqi:
@amaltaro 

As discussed, nginx does not support redirection for query string in the `path`. So, meanwhile, I have ajdusted rules to redirect all requests to `t1`. Once, we have something else, I'll update.

vkuznet:
@amaltaro , can you add REST functionality to the service which will use full explicit paths, i.e. instead of `/ms-unmerged/data/status?rses=t1` it will serve data from `/ms-unmerged/data/status/t1`? This will be truly REST API and it will solve the problem with redirect rules since we can use full path to redirects.

amaltaro:
@vkuznet `status` or `info` are the resources we are accessing, the API itself. While `rse_type` defines the search criteria to be applied by the server. To my knowledge, this is a true REST implementation and a common practice (even though your alternative is valid as well).

Regarding the problem we are trying to solve here, it only got exposed because we had to make some specific decisions and because we have also faced some issues along the way, such as:
* MSUnmerged had to be implemented with a very tight schedule, so there is no (yet) database behind this service persisting data;
* to avoid scalability issues, we had to create multiple instances/replicas of the service;
* we want to expose some basic information for each of those replicas;
* the k8s testbed and production clusters have a different architecture, thus making it impossible to test some features in a pre-production environment
* and the limitation spotted today with Nginx (which is not a problem with our pre-production setup)

Anyhow, we are planning to plug a database (likely MongoDB) to this service, where we will be able to persist data from multiple replicas, as well as retrieve them transparently, thus, removing the unnecessary complication of frontend rules to forward requests to the "correct" backend.

vkuznet:
@amaltaro my concern is that you rely on a structure of FE layer which we may replace at any time with different technology stack, e.g. cmsweb-auth will not use neither apache or nginx. And, if your application can't behave properly behind simple redirect rules I would suggest to pay attention to this problem. The purpose of FE (reverse proxy) is to redirect HTTP request based on its path (and not on parameters).

Anyway, I don't know if you (@muhammadimranfarooqi ) want to merge the PR, I only see that different backends are removed and test is changed to use  new parameters. If this is what you need I'm fine with a merge.

muhammadimranfarooqi:
Yes, this can be merged. I'll update ingress rules when @amaltaro updates the app to use path instead of parameters,

PR #733, opened 2021-08-17
dbs-migrate service helm chart

arooshap:


helm/cmsweb/dbs-migrate/values.yaml
```diff
@@ -0,0 +1,154 @@
...
+    - -verbose
+    - "0"
+  initialDelaySeconds: 120
+  periodSeconds: 10
+
+livenessProbePreProd:
+  exec:
+    command:
+    - cmsweb-ping
+    - "--url=http://localhost:8257/dbs/init/global/DBSMigrate/"
```
muhammadimranfarooqi, line 65: The URL is not correct. It should be `http://localhost:8257/dbs/int/global/DBSMigrate/`

helm/cmsweb/dbs-migrate/values.yaml
```diff
@@ -0,0 +1,154 @@
+# Default values for dbs-migrate.yaml.
+# This is a YAML-formatted file.
+# Declare variables to be passed into your templates.
+
+shareProcessNamespace: true
+
+replicaCount: 
+  prod: 5
```
muhammadimranfarooqi, line 8: The prod replica should be 3

PR #740, opened 2021-08-17
Rucio DIDS table dump for last access time of datasets

mrceyhun:
@vkuznet @ericvaandering @leggerf 

I created the sqoop job. When you give green light, I'll run it manually.

### DIFF
--- a/docker/sqoop/scripts/rucio_dids.sh
+++ b/docker/sqoop/scripts/rucio_dids.sh
@@ -1,0 +1,24 @@
+# Imports CMS_RUCIO_PROD.DIDS table for last access time of datasets
+BASE_PATH=/project/awg/cms/rucio_dids/
+JDBC_URL=jdbc:oracle:thin:@cms-nrac-scan.cern.ch:10121/CMSR_CMS_NRAC.cern.ch
+if [ -f /etc/secrets/rucio ]; then
+  USERNAME=$(cat /etc/secrets/rucio | grep username | awk '{print $2}')
+  PASSWORD=$(cat /etc/secrets/rucio | grep password | awk '{print $2}')
+else
+  echo "Unable to read Rucio credentials"
+  exit 1
+fi
+LOG_FILE=log/$(date +'%F_%H%m%S')_$(basename $0)
+TABLE=CMS_RUCIO_PROD.DIDS
+TZ=UTC
+START_DATE=$(date +'%Y-%m-%d' -d "3 months ago")
+
+/usr/hdp/sqoop/bin/sqoop import \
+  -Dmapreduce.job.user.classpath.first=true -Ddfs.client.socket-timeout=120000 --direct --connect $JDBC_URL \
+  --fetch-size 10000 --username $USERNAME --password $PASSWORD -m 1 \
+  --as-avrodatafile -z --direct --target-dir $BASE_PATH$(date +%Y-%m) \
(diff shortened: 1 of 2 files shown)


leggerf:
+1

docker/sqoop/scripts/rucio_dids.sh
```diff
@@ -0,0 +1,24 @@
...
+LOG_FILE=log/$(date +'%F_%H%m%S')_$(basename $0)
+TABLE=CMS_RUCIO_PROD.DIDS
+TZ=UTC
+START_DATE=$(date +'%Y-%m-%d' -d "3 months ago")
+
+/usr/hdp/sqoop/bin/sqoop import \
+  -Dmapreduce.job.user.classpath.first=true -Ddfs.client.socket-timeout=120000 --direct --connect $JDBC_URL \
+  --fetch-size 10000 --username $USERNAME --password $PASSWORD -m 1 \
+  --as-avrodatafile -z --direct --target-dir $BASE_PATH$(date +%Y-%m) \
+  --query "SELECT * FROM ${TABLE} WHERE accessed_at<=to_date('${START_DATE}','YYYY-MM-DD') AND scope='cms' AND project='Production' AND did_type='D' AND deleted_at IS NULL AND hidden=0 AND \$CONDITIONS" \
```
ericvaandering, line 20: I don't think there is any need to restrict it by Project.

ericvaandering (commented):
I''m curious if this gives you enough information. Do you need the sizes? The sites? In here, you don't check if these blocks are on disk somewhere.

docker/sqoop/scripts/rucio_dids.sh
mrceyhun: Thanks Eric, I'll remove Project condition.

mrceyhun:
@ericvaandering I need sizes and I assumed "bytes" field provides this data? 
The only condition of Danilo about Sites is "The HI data will not have to be considered if residing at Vanderbilt.". It would be very helpful if you know how can I filter out HI datasets in Vanderbilt.  Beside that,  I think I don't need site information.

ericvaandering:
Bytes might be good. You should double check on at least a few blocks that the sum of the files in the block matches the rucio dataset size. It may not.

As for locations, the very first line of the issue you are trying to solve says "It is important to know what datasets are on disk and are not read for a certain amount of time."

The DIDs table in Rucio has every single dataset ever known to CMS including things which have been on tape for years. (I suppose its missing things deleted from tape as well). 

There is another complication which is that Rucio will keep data around on disk even though it's not required any more until it needs space on that disk. So you should clarify with Danilo what the parameters of his request really are. If you also need datasets with a Rucio rule which requires them to be present, then that's more complicated. It probably also involves the locks you were looking at earlier.

davidlange6:
based on past experience with the reviews - the set of things actually on (central) disk and not read is what is relevant.

mrceyhun:
Thanks @ericvaandering and @davidlange6 

So, I think we're looking for only datasets on disk AFAIU. @ericvaandering What would be your suggestions if we do not account rucio rules for now. I mean, which db tables(I can join them if it's required) should I use to find access times of datasets on disks only?

ericvaandering:
I think you want the replicas table (filtered for disk RSEs) joined with the dids table on scope and name for FILES, not datasets. Then look at the accessed_at field of the DID table. This gets you away from caring about blocks or containers (CMS datasets). 

Keep in mind, "accessed_at" is very incomplete now and even more incomplete in the past, so I think you'll get garbage results for the next few months.

davidlange6:
One mainly cares about CMS datasets in the end so starting from files just means aggregating. - thats what the code shared with @ceyhun some months back does (not for the usage part, but for the on-disk tally aspect)

PR #765, opened 2021-08-23
Adjusting changes for crab namespace.

arooshap:


helm/cmsweb/crabcache/templates/configMap.yaml
```diff
@@ -34,7 +34,9 @@ metadata:
 data:
   setup-certs-and-run.sh: |
     #!/bin/bash
-    cp /host/etc/grid-security/* /etc/grid-security
+    sudo cp /host/etc/grid-security/* /etc/grid-security
     echo 'INFO Files in /etc/grid-security'
     ls -lahZ /etc/grid-security
     su  -c "cd /data && /data/run.sh" --preserve-environment  _crabcache
```
muhammadimranfarooqi, line 40: @arooshap 
This line  `su  -c "cd /data && /data/run.sh" --preserve-environment  _crabcache` should be commented.

helm/cmsweb/crabcache/values.yaml
```diff
@@ -106,25 +113,7 @@ serviceMon:
   protocol: TCP
   name: crabcache-mon
  
-command:
-- /bin/bash
-- /opt/setup-certs-and-run/setup-certs-and-run.sh
```
muhammadimranfarooqi, line 111: @arooshap 

This command (` - /opt/setup-certs-and-run/setup-certs-and-run.sh`) should not be deleted. This is in main container.

arooshap: It wasnt deleted, it was shifted upwards.
image

helm/cmsweb/crabcache/templates/configMap.yaml
arooshap: https://github.com/arooshap/CMSKubernetes/blob/a70bf64fb9597f12e8e5cddb0b166c0a7d2f2f9d/kubernetes/cmsweb/services/crabcache.yaml#L57

This needs to be updated as well.

muhammadimranfarooqi: yes, I have fixed there as well. Thanks

PR #774, opened 2021-08-30
Adding helm/cmsweb as a repo along with all the packaged charts.

arooshap:


arooshap:
@ericvaandering can you please specify why the changes are automatically happening to your charts? we tried to find a workaround, but it keeps updating the version number of your charts.

ericvaandering:
I made all those changes to test the chart generation process. But you should not be merging things into gh-page. Just let the actions do that.

ericvaandering:
OK, I see now what you're doing. I don't think that's what you want. I assume you are building these charts manually and then taging them into this branch. Rather check out #776 which is trying to add a workflow to build your charts. I'm not sure it's working. Try increasing the version number on one of your charts and see if it runs.

If it doesn't then I think you need to take all the contents of helm/cmsweb and move it to helm/ (no cmsweb) and put all the charts on equal footing. Then a single action can keep everything up to date

muhammadimranfarooqi:
@ericvaandering 

Can you adjust rucio area to `helm/rucio` as we have `helm/cmsweb` instead of just `helm`. Then we'll have similar scenario as of rucio.

ericvaandering:
I can try that, but I'm not certain it will work with the chart-releaser action.

PR #810, opened 2021-09-08
Update tag

muhammadimranfarooqi:


vkuznet:
@muhammadimranfarooqi in a future you may merge this kind of PRs as they only contains version change you apply to all images.

PR #814, opened 2021-09-10
crabserver alert if high numbr of fds

mapellidario:
As part of the efforts addressing https://github.com/dmwm/CRABServer/issues/6708, we would like to restart a crabserver pod if the number of open fds gets too high.

Implementing the restarting procedure requires creating an alert in AlertManager and then making PodManager react to this alert. This PR takes care of the first step. 

This is only a first test for us: at this moment, we only want to make sure we understand the machinery and try to restart a pod. This is why
- we are creating the alert as soon as `expr: avg_open_fds > 5`
- we would like to ask @muhammadimranfarooqi if it is possible to deploy PodManager so that it reacts to this alert only in the `test2` cluster. We checked the metrics and the pod in test2 has 9 open fds since at least 30min, so we should be able to trigger the alert.

After having made sure that we properly restart the test2 cluster, we plan to
- create a new PR with proper real-world metrics for `prod` (or only `testbed` first?)
- ask Imran to deploy PodManager so that it reacts to this alert in `prod` (or `testbed` cluster first?)

I order to implement our changes here, we took inspiration from [dbs.rules](https://github.com/dmwm/CMSKubernetes/blob/master/kubernetes/cmsweb/monitoring/prometheus/rules/dbs.rules) and [dbs.test](https://github.com/dmwm/CMSKubernetes/blob/master/kubernetes/cmsweb/monitoring/prometheus/rules/dbs.test). We have two question on this:
- since we want to trigger the alert when the average of open fds is over a certaing threshold for a somewhat long time, we are averaging over 15m. Is is ok to test this alert with `- eval_time: 15m` in `crabserver.test`?
- if we want to test the alert works, and having set the alert as `avg_open_fds > 5`, it is ok to set the input series as `    values: '1+5x1000'`, having set the eval time to 15min? We noticed that in `dbs.rules` every time that there is `values: M+Nx100`, then the eval time is 10m and value reported in the alert description is `M+10*N`, while when `values: M+Nx10` the eval time is 5min and the value in the description is `M+5*N`. We also noted that a 15min eval time is correlated to an input series containing `x1000`, but i could not find any example to copy how to set M and N in `M+Nx1000` from.

Every comment and correction is very much appreciated! :)

fyi: @ddaina @belforte

### DIFF
--- a/kubernetes/cmsweb/monitoring/prometheus/rules/crabserver.rules
+++ b/kubernetes/cmsweb/monitoring/prometheus/rules/crabserver.rules
@@ -1,7 +1,8 @@
 groups:
 - name: crabserver
   rules:
-
+  - record: avg_open_fds
+    expr: avg_over_time(crabserver_process_open_fds[15m])
   - alert: CRAB server is down
     expr: crabserver_num_cpus == 0
     for: 5m
@@ -14,3 +15,14 @@ groups:
     annotations:
       summary: "crabserver {{ $labels.env }} is down"
       description: "{{ $labels.env }} has been down for more than 5m"
+  - alert: CRAB server service has high number of fds
+    expr: avg_open_fds > 5
+    for: 1m
+    labels:
+      severity: high
+      tag: cmsweb
+      service: crab
+      action: Please check CRAB server {{ $labels.instance }} and possibly restart it
+    annotations:
+      summary: "CRAB {{ $labels.env }} environment"
+      description: "{{ $labels.env }} has high level of fds {{ $value }} for more than 1m"
--- a/kubernetes/cmsweb/monitoring/prometheus/rules/crabserver.test
+++ b/kubernetes/cmsweb/monitoring/prometheus/rules/crabserver.test
@@ -23,3 +23,22 @@ tests:
               exp_annotations:
                  summary: "crabserver prod is down"
                  description: "prod has been down for more than 5m"
+- interval: 1m
+  input_series:
+  - series: 'avg_open_fds{env="prod",instance="test-instance",host="k8s-test"}'
+    values: '6+1x100'
+  alert_rule_test:
+      - eval_time: 10m
+        alertname: CRAB server service has high number of fds
+        exp_alerts:
+            - exp_labels:
+                 severity: high
+                 tag: cmsweb
+                 service: crab
+                 host: k8s-test
+                 action: Please check CRAB server test-instance and possibly restart it
+                 instance: test-instance
+                 env: prod
+              exp_annotations:
+                 summary: "CRAB prod environment"
+                 description: "prod has high level of fds 16 for more than 1m"


muhammadimranfarooqi:
@vkuznet 

Do we need to redeploy monitoring with these changes in test2 cluster or another PR https://gitlab.cern.ch/cmsmonitoring/cmsmon-configs/-/merge_requests/11 in `cmsmon-config` should also be deployed ? Or we need both of them deployed?

vkuznet:
I merged the cmsmonitoring changes, and yes we need to redeploy monitoring in
test2 cluster to take them into effect. The cmsmonitoring/cmsmon-configs
are applied to CMS monitoring cluster, in cmsweb though we rely on
rule stored in:
https://github.com/dmwm/CMSKubernetes/tree/master/kubernetes/cmsweb/monitoring/prometheus/rules

You'll need to adjust crab rules in that location and then redeploy
prometheus (monitoring) in test2 cluster.

vkuznet:
Dario,
please read AlertManager documentation here:
https://prometheus.io/docs/alerting/latest/overview/

In particular you'll need:
https://www.prometheus.io/docs/prometheus/latest/configuration/unit_testing_rules/
https://prometheus.io/docs/alerting/latest/configuration/
to understand various parameters.

Now to answer your questions:
- you can use any threshold, e.g. avg_open_fds>5 or avg_open_fds>500,
as it will be used to fire the alert
- we can setup PodManager only on test2 cluster for your testing purposes
- for testing you can use any averaging threshold 5m or 15m as it depends
on how you decide to test your alerts. To better understand how alerts
work please refer to slide 8 in my presentation about AM:
https://www.dropbox.com/s/8p2qlr58vduey06/CMSMonitoring_AlertManager.pdf?dl=0

After you'll review all materials and may still have questions
please ask again.

Best,
Valentin.

mapellidario:
Thanks @vkuznet !

The official documentation for alert unit testing is definitely clear and I updated the commit with some more meaningful numbers.

I still have a question though: When you specify the `eval_time` at

https://github.com/dmwm/CMSKubernetes/blob/df0f6ce424908bbaf6023e5076780a4659c8cc8b/kubernetes/cmsweb/monitoring/prometheus/rules/dbs.test#L47-L52 

is it a coincidence that it is exactly double the time that you average `avg_open_fds` over

https://github.com/dmwm/CMSKubernetes/blob/df0f6ce424908bbaf6023e5076780a4659c8cc8b/kubernetes/cmsweb/monitoring/prometheus/rules/dbs.rules#L6-L7

? 

Can I use an `eval_time` that is equal or even lower than the time i average `avg_open_fds`? I do not think this would be a problem since i suppose that the values specified for example at https://github.com/dmwm/CMSKubernetes/blob/df0f6ce424908bbaf6023e5076780a4659c8cc8b/kubernetes/cmsweb/monitoring/prometheus/rules/dbs.test#L50 already simulate the average and not the values that should be averaged, so i do not find any strong argument for using a long `eval_time`.  However I may be missing something.

Moreover, do you think this PR is ready or am I missing something else?

Lastly, since the values I provided for the alert unittest in https://gitlab.cern.ch/cmsmonitoring/cmsmon-configs/-/merge_requests/11 are not the proper ones, if this causes some problems I can open a new merge request straight away. However, if this is not critical, I would correct those only when I will provide the final real-world metrics. Does this make sense?

Thanks again!

vkuznet:
Dario, 
honestly I don't remember how we end-up with settings for service rules. At some point I was writing all alert rules and test them. Therefore, I suggest that you perform unit test for your rule, see https://github.com/dmwm/CMSKubernetes/blob/master/kubernetes/cmsweb/monitoring/prometheus/rules/test_rules.sh
The beauty of AM/Prometheus that we can apply unit tests using promtool. The promtool is available on cvmfs, see /cvmfs/cms.cern.ch/cmsmon area where we put all monitoring related tools. Just use `promtool check rules your.rules` and `promtool test rules your.test` and see if it will pass. The only documentation about all AM/Prometheus parameters we have I already provided to you, if something is missing we'll need to open source code and check for details there (and you can do that too). The AM/Prometheus are written in Go and it is easy to read the code.

But, the 5m interval is very common for alerts and we'll use it everywhere. The 10m interval was (most likely) chosen to ensure that alert will be fired. As I pointed already to my slides the AM fires alert based on specific time interval triggered by different parts of the system, and I used this diagram to pick up the ranges.

Bottom line:
- perform unit test on your alert
- adjust this PR if necessary
- fix cmsmon-configs with proper alert rules

Once everything is ready and tested I can merge in both repos.

mapellidario:
Thanks Valentin! The `promtool` is wonderful, I caught a type with it!

Now the unittests look good

```plaintext
> /cvmfs/cms.cern.ch/cmsmon/promtool check rules kubernetes/cmsweb/monitoring/prometheus/rules/crabserver.rules
Checking kubernetes/cmsweb/monitoring/prometheus/rules/crabserver.rules
  SUCCESS: 3 rules found

> /cvmfs/cms.cern.ch/cmsmon/promtool test rules kubernetes/cmsweb/monitoring/prometheus/rules/crabserver.test  
Unit Testing:  kubernetes/cmsweb/monitoring/prometheus/rules/crabserver.test
  SUCCESS
```

I think this is ready to be merged. Thanks again for the information!

vkuznet:
@muhammadimranfarooqi now the PR is merged and you should apply it to testbed monitoring, if everything is fine, please put it to production too.

muhammadimranfarooqi:
@mapellidario 

I have deployed it in `test2` and `testbed` cluster.

PR #821, opened 2021-09-16
New rucio template 1.21

muhammadimranfarooqi:


ericvaandering:
I explicitly don't want their prometheus, so no "monitoring enabled" Maybe that make it the same as your CMSWeb 1.21 which would be great.

muhammadimranfarooqi:
Ok Thanks. So, this PR is not required. Closing then

PR #822, opened 2021-09-19
Adding ingress functionality to helm charts.

arooshap:


muhammadimranfarooqi:
@arooshap 

We need to adjust host which is cmsweb-test8 in this PR. It may be adjusted with environment or from argument

PR #851, opened 2021-10-28
WIP: Cert manager

Panos512:
I made a few changes but this is still not ready to be merged. 

The run.sh script now takes a directory as the argument and checks all the certs in this directory one by one.
The alert message includes the dn of the certificate and the expiration date.
I made it so that it pushes an alert if the certificate will expire in the next 15 days. This should give us enough time to react. 
The alert expiration is also set 15 days in the future.
The cmsweb tag is used in the alert to notify the correct set of people (TODO: Check if it actually works ) 

I experimented with deploying the container as a sidecar in the frontend pods but I think it's not a good strategy. Too many containers are deployed and even if we managed to renew certificates automatically we could do it from one central place. 

I have also managed to automatically renew cern host certificates but there are still a few things I need to figure out.

0. How to make this more configurable. Right now we need to specify the secrets to be mounted one by one in the yaml file. This means that we would need a different manifest file for each cluster, which should be maintained by the cluster operator. 
1. How to update the k8s secrets from within the pod, once I get the renewed certificate (this should be easy by passing the cluster config and installing kubectl)
2. How to restart the pods that are using a certificate when we renew it (This I don't know yet).

### DIFF
--- a/docker/cert-manager/run.sh
+++ b/docker/cert-manager/run.sh
@@ -1,5 +1,5 @@
 #!/bin/bash
-cert=$1
+cert_path=$1
 if [ -n  "${AMURL}" ] ; then
     amurl=$AMURL
 else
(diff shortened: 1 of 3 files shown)


vkuznet:
@Panos512 , the flexible configuration can be easily achieved through helm where you'll write template which later can be filled in with specific values. But we should decide if we'll move to helm in cmsweb. Meanwhile, what you have is sufficient since we'll only need to specify which secrets to use. The secrets with certs we can create in each cluster and they will be unique if we'll put them in dedicated namespace, e.g. certs. Therefore, you only need to adjust your yaml to use dedicated namespace (certs) and then we can try it out with different secrets in there.

For your questions 1 and 2 you should look at how we renew our proxies. The main idea is that you add kubectl to your container and use it from there. See [proxy](https://github.com/dmwm/CMSKubernetes/tree/master/docker/proxy) docker area, its Dockerfile and [proxy.sh](https://github.com/dmwm/CMSKubernetes/blob/master/docker/proxy/proxy.sh) script where at line 46 you'll see how we call kubectl within the pod which itself create a new secrets and apply them. In your case you'll create secrets you use in your manifest such that those will be re-created. Once you'll figure out that we can decide how to recreate secrets in other namespaces.

muhammadimranfarooqi:
Hi @Panos512 
Do you need this PR or it should be closed?

Panos512:
Let's close it and we can revive it if we want to continue working on that

PR #857, opened 2021-11-05
start.sh -d sets crabserver in single thread

belforte:


belforte:
@muhammadimranfarooqi  please merge, this only introduces a small change to crabserver /data/start.sh command to make it easier to run it with debugger

PR #867, opened 2021-11-18
crabserver docker image: change scripts for new py3 env

mapellidario:
With the migration of the CRAB REST interface to the py3 environment, we need to adapt the scripts that ends up in the container.

I tested the changes when building both a py2 [1] and a py3 [2] docker image.

I attach here the file where we get the version of WMCore [3] and of CRABInterface [4] from.

fyi: @belforte

[1]:

```plaintext
Step 15/17 : RUN ./addGH.sh
 ---> Running in 248b3cf9106b
/data/repos /data
Cloning into 'CRABServer'...
Cloning into 'WMCore'...
Note: checking out 'v3.210701'.
[...]
HEAD is now at e02a437... fix #6686 (#6687)
Note: checking out '1.4.6.crab1'.
[...]
HEAD is now at aaeabc2... 1.4.6.crab1
/data
```

[2]:
```plaintext
Step 15/17 : RUN ./addGH.sh
 ---> Running in dbb31cc24351
python3 is /data/srv/HG2111d-comp16/sw.dmapelli/slc7_amd64_gcc630/external/python3/3.8.2-comp/bin/python3
/data/repos /data
Cloning into 'CRABServer'...
Cloning into 'WMCore'...
Note: checking out 'py3.211104patch1'.
[...]
HEAD is now at 4ff1cc8... complete removal of unused taskbuffer
Note: checking out '1.5.5'.
[...]
HEAD is now at 80c63c9... 1.5.5
/data
```

[3]

```python
[_crabserver@d9b9df5b5010 data]$ cat $CRABServerDir/lib/python*/site-packages/WMCore/__init__.py
#!/usr/bin/env python3
"""
_WMCore_

Core libraries for Workload Management Packages

"""

__version__ = '1.5.5'
__all__ = []

```

[4]

```python
[_crabserver@d9b9df5b5010 data]$ cat $CRABServerDir/lib/python*/site-packages/CRABInterface/__init__.py
#!/usr/bin/env python3
"""
_RestInterface_

RestInterface component
"""

#the __version__ will be automatically changed when building RPMs
__version__ = 'development'


__version__ = "py3.211104patch1"#Automatically added during RPM build process

```

### DIFF
--- a/docker/crabserver/addGH.sh
+++ b/docker/crabserver/addGH.sh
@@ -13,9 +13,8 @@ CRABServerDir=`realpath /data/srv/current/sw*/*/cms/crabserver/*`
 # 1. find which GitHub tags were installed
 # beware that directory name in $CRABServerDir may not be the GH tag but rather
 # name assigned at build time like v3.230303-comp4
-export PYTHONPATH=${CRABServerDir}/lib/python2.7/site-packages
-CRABServerGHTag=`python -c "from CRABInterface import __version__; print __version__"`
-WMCoreGHTag=`python -c "from WMCore import __version__; print __version__"`
+CRABServerGHTag=$(grep __version__ $CRABServerDir/lib/python*/site-packages/CRABInterface/__init__.py | tail -n 1 | cut -f1 -d"#" | cut -f2 -d"=" |  tr -d " '\"")
+WMCoreGHTag=$(grep __version__ $CRABServerDir/lib/python*/site-packages/WMCore/__init__.py | cut -f2 -d"=" |  tr -d " '\"")
 
 # 2. create directories for repositories and clone
 mkdir /data/repos
--- a/docker/crabserver/start.sh
+++ b/docker/crabserver/start.sh
@@ -39,4 +39,5 @@ if [ "$MODE" = fromGH ]; then
 fi
 
 #==== START THE SERVICE
+export CRYPTOGRAPHY_ALLOW_OPENSSL_102=true
 /data/srv/current/config/crabserver/manage start 'I did read documentation'
--- a/docker/crabserver/run.sh
+++ b/docker/crabserver/run.sh
@@ -65,6 +65,7 @@ done
 sleep 5
 
 # start the service
+export CRYPTOGRAPHY_ALLOW_OPENSSL_102=true
 /data/srv/current/config/$srv/manage start 'I did read documentation'
 
 # run monitoring script


belforte:
thanks Dario. Since this only affects crabserver I think that we can merge w/o bugging Imran. Do you have permission to do ? Otherwise I will

docker/crabserver/addGH.sh
```diff
@@ -13,9 +13,16 @@ CRABServerDir=`realpath /data/srv/current/sw*/*/cms/crabserver/*`
 # 1. find which GitHub tags were installed
 # beware that directory name in $CRABServerDir may not be the GH tag but rather
 # name assigned at build time like v3.230303-comp4
-export PYTHONPATH=${CRABServerDir}/lib/python2.7/site-packages
-CRABServerGHTag=`python -c "from CRABInterface import __version__; print __version__"`
-WMCoreGHTag=`python -c "from WMCore import __version__; print __version__"`
+source `realpath /data/srv/HG*/apps/crabserver/etc/profile.d/init.sh`
```
belforte, line 16: Dario, can we avoid sourcing init.sh and this "if" by simply doiing some grep in `${CRABServerDir}/lib/python*/site-packages` ?

docker/crabserver/addGH.sh
```diff
@@ -13,9 +13,16 @@ CRABServerDir=`realpath /data/srv/current/sw*/*/cms/crabserver/*`
...
+if type python3; then
```
vkuznet, line 17: This will not give you what you want, e.g.
```
type bla
bla not found

type python3
python3 is /opt/local/bin/python3
```
Much better is to use `command` bash tool to check for existence of some program, see https://stackoverflow.com/questions/592620/how-can-i-check-if-a-program-exists-from-a-bash-script#677212

docker/crabserver/start.sh
```diff
@@ -39,4 +39,5 @@ if [ "$MODE" = fromGH ]; then
 fi
 
 #==== START THE SERVICE
+export CRYPTOGRAPHY_ALLOW_OPENSSL_102=true
```
vkuznet, line 42: Should you check openssl version as well here which in fact supports cryptography?

vkuznet (changes requested):

docker/crabserver/addGH.sh
mapellidario: I like this option! the new commit contains this approach

docker/crabserver/addGH.sh
mapellidario: As per Stefano suggestion, this is not needed anymore.

In any case, it relied on `python3` not being present in our py2 docker image, so that `type python3` returned non-zero exit code. 

Thanks for pointing `command -v`, always better to respect POSIX when we can, i will use it next time i need it!

docker/crabserver/start.sh
mapellidario: So, this export is here because otherwise when cherrypy starts, openssl raises an exception. 
With this export, we convert the exception to a warning (which allows the server to run)

> /data/srv/HG2111d-comp16/sw.dmapelli/slc7_amd64_gcc630/external/py3-cryptography/3.2.1-comp/lib/python3.8/site-packages/cryptography/hazmat/bindings/openssl/binding.py:175: CryptographyDeprecationWarning: OpenSSL version 1.0.2 is no longer supported by the OpenSSL project, please upgrade. The next version of cryptography will completely remove support for it.

I imagine that this is coming from https://github.com/cms-sw/cmsdist/blob/comp_gcc630/py3-cryptography.spec .The [openssl.spec](https://github.com/cms-sw/cmsdist/blob/comp_gcc630/openssl.spec) point to an even earlier version of openssl, so i think that out openssl is coming from the system (cc7, if i am not mistaken).

> Should you check openssl version as well here which in fact supports cryptography?

What woud be the advantage of checking the openssl version? I am pretty sure we are using the system one, why should i check it every time i start the crab REST erver?

The proper solution would be to have a newer version of openssl, but i do not know if this is worth the effort, if anybody else in our group already pt some effort in this direction. Any suggestion is very well appreciated! :)

Thanks for the review, by the way!

mapellidario:
Tested the build of crabserver docker image, rebased onto master, squashed 4 commits into a single one. I consider this ready to be merged.

fyi @belforte

belforte:
@vkuznet @muhammadimranfarooqi we (CRAB) think that this is OK now and have tested it already. Can you kindly merge ?

PR #872, opened 2021-11-23
New script to fix high number of dns queries

muhammadimranfarooqi:


kubernetes/cmsweb/scripts/fix_highdns.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
+
```
vkuznet, line 2: Please add here a doc string describing use-case, e.g. message we get from network team, etc.

vkuznet (changes requested):

PR #879, opened 2021-11-29
crabserver - test env - sidecar container for persistent logging

mapellidario:
When developing the crabserver rest on `cmsweb-test1` and `cmsweb-test2` environment, it would be handy to have persistent logging in case the `crabserver` pod is automatically restarted.

Taking inspiration from [1] , we can use a sidecar container whose sole purpose is to read the file where the crabserver is writing the log and stream it to its stdout. In this way:
- the logs from the crabserver are available with `kubectl -n crab logs $(kubectl -n crab get pods | tee | grep crabserver | grep "2/2" | awk '{print $1}') crabserver-sidecar | less`, even without entering in the crabserver container with bash
- the logs from the crabserver are not lost after it is restarted by the liveness probe.

I tested this on `cmsweb-test1` and it is working.

Any suggestion is very welcome! :)

fyi: @belforte 

[1]: https://kubernetes.io/docs/concepts/cluster-administration/logging/

### DIFF
--- a/kubernetes/cmsweb/scripts/deploy-srv.sh
+++ b/kubernetes/cmsweb/scripts/deploy-srv.sh
@@ -147,8 +147,8 @@ elif  [ "$cmsweb_env" == "k8s-preprod" ] ; then
 	fi
 
 elif [ "$srv" == "crabserver" ]; then
-            cat $srv.yaml | sed -e "s, #imagetag,$cmsweb_image_tag,g" |  sed -e 's+crabserver/prod+crabserver/preprod+g' | sed -e "s,k8s #k8s#,$cmsweb_env,g"  >  $srv.yaml.new
-            cat $srv.yaml | sed -e "s, #imagetag,$cmsweb_image_tag,g" |  sed -e 's+crabserver/prod+crabserver/preprod+g' | sed -e "s,k8s #k8s#,$cmsweb_env,g" | kubectl apply -f -
+            cat $srv.yaml | sed -e "s,1 #TEST#,,g" | sed -e "s,#TEST#,      ,g" | sed -e "s, #imagetag,$cmsweb_image_tag,g" |  sed -e 's+crabserver/prod+crabserver/preprod+g' | sed -e "s,k8s #k8s#,$cmsweb_env,g"  >  $srv.yaml.new
+            cat $srv.yaml | sed -e "s,1 #TEST#,,g" | sed -e "s,#TEST#,      ,g" | sed -e "s, #imagetag,$cmsweb_image_tag,g" |  sed -e 's+crabserver/prod+crabserver/preprod+g' | sed -e "s,k8s #k8s#,$cmsweb_env,g" | kubectl apply -f -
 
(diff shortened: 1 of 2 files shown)


muhammadimranfarooqi:
@mapellidario 

We have already logs available in stdout with `crabserver-filebeat` sidecar container. If you want them enable in test clusters, then may be better to remove #PROD# prefix from this sidecar instead of writing a new sidecar

mapellidario:
> We have already logs available in stdout with crabserver-filebeat sidecar container. If you want them enable in test clusters, then may be better to remove #PROD# prefix from this sidecar instead of writing a new sidecar

Sounds good to me, i assumed there was some reason not enable filebeat on test1/test2 clusters.

muhammadimranfarooqi:
The logs in stdout are temporary and obviously use the disk space of nodes but these automatically wash out. I don't know the exact size/time for this operation. 

We don't use filebeat sidecar in test clusters in order to avoid extra operation and also we don't use fluentd in test clusters to scrap logs to S3. Nothing else

belforte:
let's give it a try then !!

belforte:
@mapellidario is this PR really needed ? I thought that Imran already have things setup and can make this happen by simply removing a couple comments https://github.com/dmwm/CMSKubernetes/pull/879#issuecomment-981720134
I am not sure what the goal is.

vkuznet:
@mapellidario I doubt these changes are needed. WE only use `#PROD#` directive which is removed by our scripts when we deploy stuff to production environment. Here you removed explicitly all of them and introduced yet another directive `#TEST#` which reverse the `#PROD#`. Please do not do that and put back `#PROD#` and organize your changes around it.

belforte:
I suspect that Dario is pushing things to his branch for his own testing/learning, unaware of the noise created. I suggest to close this PR and look forward to a new one once dust has settled.

mapellidario:
> @mapellidario I doubt these changes are needed. WE only use #PROD# directive which is removed by our scripts when we deploy stuff to production environment. Here you removed explicitly all of them and introduced yet another directive #TEST# which reverse the #PROD#. Please do not do that and put back #PROD# and organize your changes around it.

I saw how you use `#PROD#`, i removed it because we want to use the same configuration for our test cluster. I know I introduced `#TEST#`, and I updated the script `kubernetes/cmsweb/scripts/deploy-srv.sh` to properly use it.

vkuznet:
@mapellidario instead of adding additional stuff to existing deployment procedure I rather suggest that you embrace helm charts which can provide you flexibility to deal with different environments. We already decided to move forward with charts and @muhammadimranfarooqi will prepare migration plan. Therefore, it would be better if you'll adopt it earlier and avoid duplicating work. The helm chart for crabserver already exists in [here](https://github.com/dmwm/CMSKubernetes/tree/master/helm/crabserver)

belforte:
this is a very interesting and useful discussion, but does not belong in a PullRequest.  I suggest again to close and to debate in the proper channel "what's the best way for CRAB developers to look at crab logs of their test container which just crashed"

mapellidario:
This is a very good suggestion, thanks Valentin!

Sorry for the noise, I will close this PR.

PR #921, opened 2022-01-26
Fix httpgo

mrceyhun:
Already deployed to ha1 (others are fine). 

@vkuznet  Should we add HA deployments to `deploy.sh`? 
`deploy.sh` needs some improvements IIUC: secrets deployments are not same with `deploy-secrets.sh` and no HA deployments. I'll implement it in another PR if you give green light.

### DIFF
--- a/kubernetes/monitoring/services/httpgo.yaml
+++ b/kubernetes/monitoring/services/httpgo.yaml
@@ -48,11 +48,3 @@ spec:
               command: ["/bin/sh", "-c", "ls"]
         ports:
         - containerPort: 8888
-        volumeMounts:
-        - name: proxy-secrets
-          mountPath: /etc/proxy
-          readOnly: true
-      volumes:
-      - name: proxy-secrets
-        secret:
-          secretName: proxy-secrets


vkuznet:
@mrceyhun , I merged this PR. Here is answers to your questions:
- of course `deploy.sh` and other scripts require maintenance. I did my best to create them in a first place and keep up-to-date with evolving infrastructure. I think it is your turn now to adjust them to be up-to-date
- I suggest to now put secrets in `deploy.sh` per-se and keep `deploy-secrets.sh` separate. But it requires that order should be preserved to first do secrets and then deployment. Therefore, it must be documented.
- probably it is better to have different `deploy.sh` scripts for main and HA clusters, but it is up to you to decide. It also depends on who will be in charge of deploying services on clusters in a future. You should evaluate long-term maintenance, i.e. when you may not be around and someone else should perform the deployment.
- Eventually, you should start thinking to migrate `deploy.sh` to helm charts. We already did that work in cmsweb (using summer student) and plan to start this migration during this year.

Said that, I want you to evaluate all pieces which involves deployment procedure, and moreover, try it out on fresh cluster from scratch. Of course, full documentation should be in place. Please have a look at [cmsweb deployment](https://cms-http-group.docs.cern.ch/k8s_cluster/cmsweb-deployment/) and made up something similar for monitoring cluster.

PR #944, opened 2022-02-14
Arrange monitoring storage configs

mrceyhun:
Fyi @leggerf @brij01


muhammadimranfarooqi (approved):

mrceyhun:
Thanks to Muhammad, we changed the storage classes definitions to declarative way. "cmsmonitnew" is already declarative. So, we applied new storage definitions to all 3 clusters: ha1, ha2, vm-agg. The services that use mentioned volumes are stopped and started to apply new storage configurations. 
Everything seems fine for now.

PR #947, opened 2022-02-16
Updating deploy secrets for client secrets substitution

muhammadimranfarooqi:


kubernetes/cmsweb/scripts/deploy-secrets.sh
```diff
@@ -42,13 +42,27 @@ conf=$3
...
         files=""
+
+    if [ "$srv" == "auth-proxy-server" ] || [ "$srv" == "x509-proxy-server" ] || [ "$srv" == "scitokens-proxy-server" ] ; then
+        if [ -d $secretdir ] && [ -n "`ls $secretdir`" ] && [ -f $secretdir/client.secrets ]; then
+           export CLIENT_SECRET=`grep CLIENT_SECRET $secretdir/client.secrets | head -n1 | awk '{print $2}'`
+           export CLIENT_ID=`grep CLIENT_ID $secretdir/client.secrets | head -n1 | awk '{print $2}'`
+           export IAM_CLIENT_ID=`grep IAM_CLIENT_ID $secretdir/client.secrets | head -n1 | awk '{print $2}'`
+           export IAM_CLIENT_SECRET=`grep IAM_CLIENT_SECRET $secretdir/client.secrets | head -n1 | awk '{print $2}'`
+	   if [ -f $secretdir/config.json ]; then
+              echo "$(jq '.client_id=env.CLIENT_ID' $secretdir/config.json)" > $secretdir/config.json
```
vkuznet, line 58: I would avoid using `jq` tool since it may not present on a node. Instead it is much safer to rely on `sed`. For instance, you can replace a secret as following:
```
sed -i -e "s,CLIENT_ID,$CLIENT_ID,g" $secretdir/config.json
```
etc.

Here we only need to ensure that our `config.json` has these lines:
```
    "client_id": "CLIENT_ID",
    "client_secret": "CLIENT_SECRET",
    "iam_client_id": "IAM_CLIENT_ID",
    "iam_client_secret": "IAM_CLIENT_SECRET",
...
    "http://COUCHDB_USER:COUCHDB_PASSWORD@,,,,"
```

So, in configs we will use values for our secret attributes as name of env variables which will hold the actual values.

muhammadimranfarooqi:
Thanks @vkuznet 
Could you please review again.

vkuznet (changes requested):
@muhammadimranfarooqi  now structure looks good to me, but I want to request an additional clean-up step. When someone call `sed -i` (in-place changes) the sed leave an original file with `-e` extension, therefore I suggest that you test `sed` calls and add additional if statement to remove backup file, e.g.
```
if [ -f $secretdir/config.json-e ]; then
   rm $secretir/config.json-e
fi
```

muhammadimranfarooqi:
Hi @vkuznet 

I checked there is no such file generated with `sed -i -e`. However, if we use `sed -i-e` then it is generated. So, we don't need this additional check.

vkuznet:
ok, then it is only macOS thing :) since I do see it on a mac using mac version of sed. So, I'm merging and closing this issue. Let's adjust configs in gitlab then and we'll be done.

PR #950, opened 2022-02-17
reduce memory limit to 1.5Gi and incerase hpa to 10

novicecpp:
Reduce mem limit to 1.5Gi to use with new crabserver config: 10 threadpool, scale out pod to 10.

[memory usage of 5(green)/10(orange)/15(magenta)/25(sky) threadpools](https://monit-grafana.cern.ch/d/cmsweb_crabserver/crabserver-cmsweb?orgId=11&from=1644434324136&to=1645026172547&viewPanel=29)
Screenshot from 2022-02-17 10-25-49

Increase hpa number in case anyone accidentally apply hpa manifest.

### DIFF
--- a/kubernetes/cmsweb/hpa/crabserver-hpa.yaml
+++ b/kubernetes/cmsweb/hpa/crabserver-hpa.yaml
@@ -4,8 +4,8 @@ metadata:
   name: crabserver-hpa
   namespace: crab
 spec:
-  maxReplicas: 8
-  minReplicas: 5
+  maxReplicas: 10
+  minReplicas: 10
   scaleTargetRef:
     apiVersion: apps/v1
     kind: Deployment
--- a/kubernetes/cmsweb/services/crabserver.yaml
+++ b/kubernetes/cmsweb/services/crabserver.yaml
@@ -69,7 +69,7 @@ spec:
   selector:
     matchLabels:
       app: crabserver
-  replicas: 1 #PROD# 5
+  replicas: 1 #PROD# 10
   template:
     metadata:
       labels:
@@ -92,7 +92,7 @@ spec:
             memory: "256Mi"
             cpu: "200m"
           limits:
-            memory: "3Gi"
+            memory: "1.5Gi"
             cpu: "1500m"
         livenessProbe:
           exec:


novicecpp:
@belforte @mapellidario please review.

vkuznet:
@novicecpp could you please justify why hpa should be increased to 10? Please note, we can't just take any number and apply it since it affects resource planning for entire cluster which has other services. Instead, we need a proof how much load each service can sustain. You may generate load to your service via different tools and use monitoring dashboards to understand how much resource your service really need.

If you are not experience how to generate load I suggest that you get familiar with tools we use. On cvmfs we have hey tool,
```
/cvmfs/cms.cern.ch/cmsmon/hey -h
...
```
You can easily test your server by running the tool with basic set of parameters, e.g.
```
# generate your proxy
voms-proxy-init -voms cms -rfc

# run hey with 200 requests and 10 concurrent clients against cmxweb-testbed
/cvmfs/cms.cern.ch/cmsmon/hey -n 200 -c 10 https://cmsweb-testbed.cern.ch
All requests done.

Summary:
  Total:        0.4799 secs
  Slowest:      0.1949 secs
  Fastest:      0.0009 secs
  Average:      0.0080 secs
  Requests/sec: 416.7173
  Total data:   565000 bytes
  Size/request: 2825 bytes

Status code distribution:
  [200] 200 responses
....
```

Using such tools you can generate load against your server APIs and figure out how much load your service can sustain. There are other tools on a market such as apache bench, etc. We use hey because it supports X509 proxy we need to access the cluster. Once you'll figure out the load (which should imitate your usual service usage) you can consciously define your service limits, show us plots and tell us which limit we need to use.

belforte:
I think this was supposed to be done together with decreasing number of pool slaves in each crabserver from 25 to 10 or 15. In other words, trying to answer @vkuznet on why we did not do that study which you indicate yet, the idea is to have more pods with fewer slaves in each. This configuration appears to be more stable as far as memory usage and should have same overall impact on CMSWEB.
Current configuration is working, but not very well. Since we moved to python3 stack something has changed in memory management and memory usage by CRABServer is much higher than before and continuously growing, current HPA settings now means that we always run with 8 pods instead of 5. So some change even experimental, is due.

As you know it is not easy to setup a realistic stress test with a servcie which handles thousands of calls, many of them being write or modity of DB records which can't be really replayed from logs. I am not saying that it should not be done, it was one of first goals for @mapellidario as he joined in September, but we have not got to do it yet. Still is in the plan.

kubernetes/cmsweb/hpa/crabserver-hpa.yaml
```diff
@@ -4,8 +4,8 @@ metadata:
   name: crabserver-hpa
   namespace: crab
 spec:
-  maxReplicas: 8
-  minReplicas: 5
+  maxReplicas: 10
+  minReplicas: 10
```
muhammadimranfarooqi, line 8: @novicecpp 
minimum and maximum replicas should not be same. Otherwise there is no point of having hpa. Minimum replicaas should be less than 10

novicecpp:
Sorry for not providing the contexts and thanks Stefano for explaining what is going on here.

* We use much more memory in py3 version (~300 MB in py2 vs ~1.8GB in py3, 6 times). [grafana](https://monit-grafana.cern.ch/d/cmsweb_crabserver/crabserver-cmsweb?orgId=11&from=1642514562853&to=1645106562853&viewPanel=29)
Screenshot from 2022-02-17 15-02-58_2
* We only know that the number of threads directly impacts memory usage (both test and production environment), but we do not know why.
* Based on previous config (25 threads per instance, 5 pods). We reduce the number of threads to 10 to make it consume memory lower than 1 GB, scale it out to 10 pods to match the number of threads that have been reduced. Let Kubernetes spreading loads/resources for us.

kubernetes/cmsweb/hpa/crabserver-hpa.yaml
novicecpp: The new minimum number of pods will be 10 and we do not want to use hpa. I change this value in case someone accidentally applies it.

vkuznet:
@belforte you should improve your server metrics by adding those which may lead you to discovery. For instance, I provided generic [ProcessStats](https://github.com/dmwm/WMCore/blob/master/src/python/Utils/ProcessStats.py) WMCore module in a past. It has basic example of what you can obtain within your server. You may either expand or write your own class which will collect metrics like, number of HTTP request, number of ORACLE conenctions, all metrics from [psutil](https://psutil.readthedocs.io/en/latest/), etc. It is your server and only you know what's going on within your code. What you need is a new end-point, e.g. `/crab/metrics`, which you can expose from your server. Then we can get prometheus exporter to scrape your metrics and integrate them into monitoring.

I also advise that you enable cherrypy stats, see [WMCore REST Server](https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/REST/Server.py#L364). If CRAB server uses WMCore REST server then you already have `/<crabserver>/stats` end-point. Then, you may add cherrypy exporter within your docker image and it will provide all cherrypy metrics of your server, see example [dbs monitor.sh](https://github.com/dmwm/CMSKubernetes/blob/master/docker/dbs/monitor.sh) which runs different exporters (one for process stats and another for cherrypy).

Bottom line, increase your server monitoring metrics which will give you a clue about your server issue.

mapellidario:
While we are here, I have a quick question for @vkuznet. I had a look at the [cmsweb doc](https://cms-http-group.docs.cern.ch/k8s_cluster/benchmark/#usage-of-hey-tool) about hey. I put two urls in `urls.txt` and asked to perform 10 requests [1]. Do you know how are these requests randomized? Can we assume they are evenly distributed among the provided urls? 

In our case, the request that most stresses the memory consumption is a GET with basically only one parameter. In order to perform a memory stress test of crabserver, we can create a list of ~1k urls, all calling the same endpoint but with different data and have hey randomize the test requests over such a list. Most of the responses are [small](https://monit-grafana.cern.ch/d/wIybNUV7z/user-dmapelli-crab-timber-pods?orgId=11&from=now-3h&to=now&var-system=crabserver&var-method=All&var-api=All&var-code=All&var-metadataType=All&var-Filters=data.code%7C%3E%7C1&var-Filters=data.api%7C!%3D%7Cinfo&var-copy_of_system=All&var-cluster=All&var-env=All&var-client=All&var-bin=100s&var-dnFilter=.*&viewPanel=29) (~20k), but few are [huge](https://monit-grafana.cern.ch/d/wIybNUV7z/user-dmapelli-crab-timber-pods?orgId=11&from=now-3h&to=now&var-system=crabserver&var-method=All&var-api=All&var-code=All&var-metadataType=All&var-Filters=data.code%7C%3E%7C1&var-Filters=data.api%7C!%3D%7Cinfo&var-copy_of_system=All&var-cluster=All&var-env=All&var-client=All&var-bin=100s&var-dnFilter=.*&viewPanel=62) (~150MB). If the queries are not balanced, then we risk never making the big requests, compromising the validity of the test.

I had a look at [hey source code](https://github.com/rakyll/hey/), but the current master uses the [-U](https://github.com/rakyll/hey/blob/898582754e00405372f0686641441168f4e2f489/hey.go#L91) argument for setting the user agent, not for reading urls from a file. Do you have a private fork of hey or should I look back in hey commit history?

Thanks a lot for the suggestion of using hey! It already look really promising for our purposes.

[1]

```plaintext
/cvmfs/cms.cern.ch/cmsmon/hey -disable-keepalive -T "*/*" -A "*/*" -N 10 -n 10 -c 5 -o csv -U ./urls.txt 
0-th request : https://cmsweb-testbed.cern.ch:8443/crabserver/preprod/info
1-th request : https://cmsweb-test1.cern.ch:8443/crabserver/dev/info
2-th request : https://cmsweb-testbed.cern.ch:8443/crabserver/preprod/info
3-th request : https://cmsweb-test1.cern.ch:8443/crabserver/dev/info
4-th request : https://cmsweb-test1.cern.ch:8443/crabserver/dev/info
5-th request : https://cmsweb-testbed.cern.ch:8443/crabserver/preprod/info
6-th request : https://cmsweb-test1.cern.ch:8443/crabserver/dev/info
7-th request : https://cmsweb-test1.cern.ch:8443/crabserver/dev/info
8-th request : https://cmsweb-testbed.cern.ch:8443/crabserver/preprod/info
9-th request : https://cmsweb-testbed.cern.ch:8443/crabserver/preprod/info
[...]
```

vkuznet:
@mapellidario you should use my branch https://github.com/vkuznet/hey/tree/x509-csv-fixes which I forked form hey main branch in order to add x509 support. If I recall correctly queries are randomly used, see https://github.com/vkuznet/hey/blob/x509-csv-fixes/requester/requester.go#L103. I also changed `-U` flag, see https://github.com/vkuznet/hey/blob/x509-csv-fixes/hey.go#L106 to read from provide url file.

mapellidario:
Looks perfect, that's all I needed. Thank you very much!

belforte:
thanks for great advice @vkuznet , we will do.

belforte:
Hi @muhammadimranfarooqi and @vkuznet . Thanks for help and great suggestions. We need to progress now.
Wa, Dario and myself discussed things and we propose the following:
1. let's close this PR w/o merging
2. let's make no change to HPA policy now
3. we will make a new PR with a small change to crabserver/config.py that hope you will approve and that in our opinion may lead to better overall memory usage
4. we will keep studying memory allocaiton, will add more monitoring, will try to find server limits in test stamd etc. and hopefully will find a better metric than the current one for HPA scaling.

vkuznet:
@belforte your plan seems solid to me and we are here to help with questions. In my view understanding and fixing the root cause is the best solution and your suggestion fits into this. Since we do have dedicated cmsweb-testX clusters I suggest that your perform your studies over there, and **document** them. It is not a first time we have issues with Python servers and any documentation can help other dev-teams if they will experience similar problem.

Said that, I'm closing this PR, if necessary we can re-open it in a future.

PR #954, opened 2022-02-21
Adding new cronjob for token

muhammadimranfarooqi:


docker/token/Dockerfile
```diff
@@ -0,0 +1,25 @@
+FROM registry.cern.ch/cmsweb/cmsweb:20220201-stable
```
vkuznet, line 1: WE must replace this image since it is TOO big:
```
registry.cern.ch/cmsweb/cmsweb      20220201-stable    77a74bf8539e   2 weeks ago     2.57GB
```
and we only need shell, curl and kubectl tools to get the token. I suggest to replace cmsweb with debian, something like that (I didn't test it):
```
FROM debian:stable-slim
RUN apt-get update && apt-get -y install curl

# Install latest kubectl for using with crons
RUN curl -k -O -L https://storage.googleapis.com/kubernetes-release/release/$(curl -s https://storage.googleapis.com/kubernetes-release/release/stable.txt)/bin/linux/amd64/kubectl
RUN mv kubectl /usr/bin
RUN chmod +x /usr/bin/kubectl
```
This will give image with O(100MB) image size instead of 2.7GB.

docker/token/token.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
+
+if [ -f /etc/client/client_id ] && [ -f /etc/client/client_secret ]; then
+    export client_id=`cat /etc/client/client_id`
+    export client_secret=`cat /etc/client/client_secret`
+    curl -s -d grant_type=client_credentials -d scope="profile" -u ${client_id}:${client_secret} https://cms-auth.web.cern.ch/token | python2 -c 'import sys,json; print(json.loads(sys.stdin.read())["access_token"])' > /tmp/token
+
+    if [ -f /tmp/token ]; then
+        kubectl create secret generic token-secrets \
```
vkuznet, line 9: please add echo to indicate when token is created, this will be useful to have in k8s logs when we'll need to debug token issue and will need to know when last time token was successfully created. Please put timestamp and proper message, including size of /tmp/token file.

vkuznet (changes requested):

vkuznet (changes requested):
Please only adjust echo printouts that we'll have normal logging output. After that feel free to merge.

PR #963, opened 2022-03-07
Add node-exporter to cmsweb prometheus

mrceyhun:
As I understood from the helm chart documentations provided in CMSKUBERNETES-131, their node-exporter metrics can be scraped by the prometheus that run in the same cluster. So, here is the new config to get node metrics to the cmsweb prometheus.

fyi @muhammadimranfarooqi @vkuznet


muhammadimranfarooqi:
@mrceyhun 

We use this file for services https://github.com/dmwm/CMSKubernetes/blob/master/kubernetes/cmsweb/monitoring/prometheus/services/prometheus.yaml

PR #969, opened 2022-03-10
add new filebeat sidecar to crabserver

novicecpp:
Add new sidecar for sending crabserver cherrypy http log to es monit. For building new crabserver dashboard.
Send it to cmsweb logstash and parse with new logstash config here https://github.com/dmwm/CMSKubernetes/pull/968

### DIFF
--- a/kubernetes/cmsweb/services/crabserver.yaml
+++ b/kubernetes/cmsweb/services/crabserver.yaml
@@ -40,12 +40,12 @@ data:
       paths:
         - /data/srv/logs/crabserver/*${MY_POD_NAME}*.log
       file_identity.path:
-      ignore_older: 1h
       scan_frequency: 10s
       backoff: 5s
       max_backoff: 10s
       include_lines:
         - '^\[.+?\] crabserver-'
+      tags: ["crabserver"]
     output.logstash:
       hosts: ["logstash.monitoring:5044"]
       compression_level: 3
--- a/kubernetes/cmsweb/services/crabserver.yaml
+++ b/kubernetes/cmsweb/services/crabserver.yaml
@@ -39,6 +39,7 @@ data:
       enabled: true
       paths:
         - /data/srv/logs/crabserver/*${MY_POD_NAME}*.log
+      file_identity.path:
       ignore_older: 1h
       scan_frequency: 10s
       backoff: 5s
--- a/kubernetes/cmsweb/services/crabserver.yaml
+++ b/kubernetes/cmsweb/services/crabserver.yaml
@@ -278,4 +278,4 @@ spec:
 #PROD#    name: crabserver-filebeat-monit-config
 #PROD#- name: filebeat-cephfs
 #PROD#  persistentVolumeClaim:
-#PROD#      claimName: filebeat-cephfs-claim-default
+#PROD#      claimName: crabserver-filebeat-monit-data
--- a/kubernetes/cmsweb/services/crabserver.yaml
+++ b/kubernetes/cmsweb/services/crabserver.yaml
@@ -210,8 +210,9 @@ spec:
 #PROD#- name: crabserver-filebeat-monit
 #PROD#  image: docker.elastic.co/beats/filebeat:7.12.1
 #PROD#  args: [
-#PROD#    "-c", "/etc/filebeat.yml",
-#PROD#    "-e",
+#PROD#    "bash",
+#PROD#    "-c",
+#PROD#    "filebeat -c /etc/filebeat.yml --path.data /data/filebeat/${MY_POD_NAME}/data -e",
 #PROD#  ]
 #PROD#  env:
 #PROD#  - name: MY_POD_NAME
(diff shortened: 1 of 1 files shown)


novicecpp:
@muhammadimranfarooqi Please review.

muhammadimranfarooqi:
@novicecpp 

You did not use registry as we use in the frontend. https://github.com/dmwm/CMSKubernetes/blob/master/docker/frontend/monitor.sh#L21
If you don't use it, then restarting of sidecar container will start reading files from scratch. In order persist this info, we need to have registry..  We faced this issue in frontend.

@mrceyhun  FYI

novicecpp:
Fixed.

Thanks Imran. I do not know filebeat has marking problem with network storage like cephfs.

muhammadimranfarooqi:
@novicecpp 

I think its better to run filebeat from inside main service container using `monitor.sh` script as we run for frontends instead from running sidecar. https://github.com/dmwm/CMSKubernetes/blob/master/docker/frontend/monitor.sh

Then use this section in manifest file.
https://github.com/dmwm/CMSKubernetes/blob/master/kubernetes/cmsweb/daemonset/frontend-ds.yaml#L107-L108
and
https://github.com/dmwm/CMSKubernetes/blob/master/kubernetes/cmsweb/daemonset/frontend-ds.yaml#L170-L172

This scenario has been tested and working very well. However, I am not sure about the sidecar solution will work without any issues.

mrceyhun:
+1 @muhammadimranfarooqi 

If you still want to run Filebeat as sidecar, please do not forget to use [PersistentVolume](https://github.com/dmwm/CMSKubernetes/blob/master/kubernetes/cmsweb/daemonset/frontend-ds.yaml#L107-L108) for filebeat. If your work directory is `/data/filebeat`, filebeat stores all metadata (offsets of files, tstamps, etc) in `/data/filebeat/data/registry`. So you need to host it in PV. Otherwise, when filbeat is restarted, all metadata will be lost and filebeat will re-read all log files that are already read.

As far as I remember, [file_identity.path:](https://github.com/dmwm/CMSKubernetes/blob/master/docker/frontend/monitor.sh#L21) is for a some symbolic link problem or some Kubernetes directory path problem which occurs in Pod restarts. Its default value is `: ~`, so we removed it. This is different thing than filebeat registry.

mrceyhun:
If filebeat will read just the pod's own output data and you'll use pod name in the directory, yes, you may not need persistent volume.

belforte:
I do not understand all details here. At the moment crabserver logs which we want to read are on a CEPH shared volume, i.e. persistent. But filename contains the pod name. So when a pod is restarted (well a new pod started, IIUC one deletes a pod or creates a new one) filebeat should then read a new file (IIUC) and there is no duplication of data.
Maybe this naive description agrees with last comments by @novicecpp and @mrceyhun .
In any case, this is a new functionality, so there's nothing to break and experimenting is welcome. Give it a try and see how it goes !

muhammadimranfarooqi:
Hi,

The problem will come when only filebeat sidecar container restart while crabserver container will not restart. In this case, podname will remain the same and filebeat sidecar container will read crabserver logs from scratch. That's why I suggested to put filebeat inside crabserver container and use registry for filebeat persistency.

novicecpp:
`But filename contains the pod name` I forget this point. Thanks @belforte!

@muhammadimranfarooqi it should survive restart if I mount persistent storage to filebeat's sidecar like [you show me here](https://github.com/dmwm/CMSKubernetes/blob/master/kubernetes/cmsweb/daemonset/frontend-ds.yaml#L107-L108), isn't it? Or am I misunderstanding something?

muhammadimranfarooqi:
Hi @novicecpp 
I dont' know how you'll manage different configuration for each filebeat that runs as side car using your manifest file. 

In cephfs for frontend [here](https://github.com/dmwm/CMSKubernetes/blob/master/docker/frontend/monitor.sh#L46-L55
we have location `/data/filebeat` and in this location, every container has its own configuration

```
[_frontend@cmsweb-uiyd2lkjujqz-node-0 filebeat]$ ls /data/filebeat
frontend-24q49  frontend-2x7cn  frontend-5v5r4  frontend-85q2c  frontend-9fjbp  frontend-c4cm7  frontend-dsf7b  frontend-h5mlt  frontend-j5nnw  frontend-kjxjg  frontend-mj2sd  frontend-pcsp6  frontend-qzl77  frontend-s7t2d  frontend-ts6p2  frontend-wj9qz  frontend-z2txz
frontend-24sg9  frontend-2x9qg  frontend-6bskp  frontend-85q57  frontend-9fqrs  frontend-c5vfl  frontend-dzn7s  frontend-h62p5  ```
```
 [_frontend@cmsweb-uiyd2lkjujqz-node-0 filebeat]$ ls /data/filebeat/frontend-24q49/
data  log
```
If you want, i can mege this and you can test it in testbed?

File: kubernetes/cmsweb/services/crabserver.yaml [hunk 1]
```diff
@@ -25,6 +25,35 @@ data:
...
+      enabled: true
+      paths:
+        - /data/srv/logs/crabserver/*${MY_POD_NAME}*.log
+      file_identity.path:
+      ignore_older: 1h
+      scan_frequency: 10s
+      backoff: 5s
+      max_backoff: 10s
+      include_lines:
+        - '^\[.+?\] crabserver-'
```
mrceyhun, line 48: You need to add `crabserver` tag, same level with include_lines:
```
  include_lines:
    - '^\[.+?\] crabserver-'
  tags: ["crabserver"]
```

mrceyhun:
@novicecpp After adding tag, if you're okay with `ignore_older: 1h` setting, it seems ready to deploy.

mrceyhun:
@novicecpp After adding tag, if you're okay with `ignore_older: 1h` setting, it seems ready to deploy.

PR #985, opened 2022-03-18
Add crab logstash to cmsweb for taskworker vm

mrceyhun:
@muhammadimranfarooqi we want to run a logstash for crab in `-n crab` in one of cmsweb cluster. I edited the `deploy.sh` but we need to deploy it only one cluster. In service, there is a "NodePort" to be accessed by Taskworker VM.

fyi @novicecpp

### DIFF
--- a/kubernetes/cmsweb/monitoring/crab/README.md
+++ b/kubernetes/cmsweb/monitoring/crab/README.md
@@ -1,0 +1,53 @@
+## CRAB LOGSTASH
+
+Used for parsing CRAB Taskworker VM instance logs. Filebeat should send data with `crabtaskworker` tag.
+
+- Refs: https://monit-docs.web.cern.ch/metrics/http/
+- https://its.cern.ch/jira/browse/CMSMONIT-460
+
+#### Resorved keywords
+
+They are used in grok parsers (alpha order) !!please update the list when you update the grok!!
+
+```
+- acquiredFiles, acquiredFilesStatus, action,
+- blocks,
+- completionTime,
+- exceptionHandled,
+- files, filesPublished, functionName,
+- logMsg, log_type,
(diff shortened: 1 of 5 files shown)


mrceyhun:
Suggested `filebeat.yml` with correct paths is below. You can edit below settings according to VM. Cmsweb cluster alias will be defined after the deployment, so we need to wait its deployment. Please review my comments and , if possible, I suggest to test this conf with `output.console`. After tests, I suggest to clean registry '$FILEBEAT_HOME/data/*' before sending data to logstash.

```
filebeat.inputs:
- type: log
  enabled: true
  paths:
    - /data/srv/TaskManager/logs/twlog.txt
    - /data/srv/Publisher*/logs/log.txt
    #- /data/srv/logs/frontend/access_log_${NAME}*.txt  !I don't know this is a correct path for Taskworker VM!
  #ignore_older: 1h !You can edit this setting to not send old data! 
  scan_frequency: 10s
  backoff: 5s
  max_backoff: 10s
  tags: ["crabtaskworker"]
filebeat.config.modules:
  path: ${path.config}/modules.d/*.yml
  reload.enabled: true

# 30044 is nodeport
output.logstash:
  hosts: ["CMSWEB_DNS_ALIAS:30044"]
  compression_level: 3
  bulk_max_size: 4096

queue.mem:
  events: 65536

logging.metrics.enabled: false

processors:
  - add_host_metadata:
      when.not.contains.tags: forwarded
  - add_cloud_metadata: ~
  - add_docker_metadata: ~
  - add_kubernetes_metadata: ~
  # !add_kubernetes_metadata may produce error in a VM, you need to check!

# You may remove below lines because filebeat do not send data directly to ES 
#setup.template.settings:
#  index.number_of_shards: 1
#setup.kibana:
```

PR #1012, opened 2022-04-11
Update logstash confs for new filebeat versions and add client_name

mrceyhun:
- Tested with filebeat versions: v7.10, v7.12, v8.1.0, v8.1.2
- New default filebeat fields to remove
- Since we're using json_batch, we don't need `message` part which is confusing.
- New `client_name` is added which is extracted from `client`. `client_name` is equal to `user_agent_name` of aps/sps/xps one.
- `agent_name`, `agent_id`, `agent_version` are changed with `filebeat_` prefix since they are filebeat meta data.

@vkuznet what do you think about `client_name` parsing?

### DIFF
--- a/kubernetes/cmsweb/monitoring/aps/logstash.conf
+++ b/kubernetes/cmsweb/monitoring/aps/logstash.conf
@@ -9,15 +9,17 @@ input {
 filter {
   mutate {
       add_field => {
-          "hostname" => "%{[agent][hostname]}"
-          "agent_id" => "%{[agent][id]}"
+          "hostname" => "%{[host][name]}"
+          "filebeat_id" => "%{[agent][id]}"
+          "filebeat_name" => "%{[agent][name]}"
+          "filebeat_version" => "%{[agent][version]}"
           "ephemeral_id" => "%{[agent][ephemeral_id]}"
           "cmsweb_log" => "%{[log][file][path]}"
           "cmsweb_cluster" => "${CMSWEB_CLUSTER:NA}"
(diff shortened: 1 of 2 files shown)


vkuznet:
regarding `client_name`, I still think you should add a split between name and version for any given client since we scrape logs from different providers, e.g. apache, aps, crab, etc. Having stats on these attribute is important and knowing which version of tools are used to access is also useful information.

mrceyhun:
@vkuznet 
For a `client: WMCore.Services.Requests/v002`:`client_name` will be `WMCore.Services.Requests`. You mean to add new `client_version` for `v002` part?

vkuznet:
yes, precisely that. In other words, everything before first slash is a name, and after the first slash is a version. This seems to be consistent with most of the user-agent naming convention. In my experience it is important to know which version of tool we use, e.g. dasgoclient v1, v2, v3 where I will know immediately if users still use old version which may have non fixed bug in a code.

mrceyhun:
@vkuznet Now we'll get `client_name` and `client_version` from `client` with `'%{DATA:client_name}/%{DATA:client_version}$'` grok. Example results:
```
1st
"client" => "WMCore.Services.Requests/v002"
"client_name" => "WMCore.Services.Requests"
"client_version" => "v002"

2nd
"client" => "Python-urllib/2.7
"client_name" => "Python-urllib"
"client_version" => "2.7"

Mozilla ones also works fine with multiple slashes
```

vkuznet:
+1

PR #1014, opened 2022-04-18
Fix rucio mon crond

mrceyhun:
NodePorts 31201 and 31202 will be used in `cmsmonitoring` cluster to reach Spark Cluster (UI and Block Manager)
Related with  https://github.com/dmwm/CMSSpark/pull/84

fyi @vkuznet

### DIFF
--- a/docker/cmsmon-rucio-ds/Dockerfile
+++ b/docker/cmsmon-rucio-ds/Dockerfile
@@ -4,8 +4,7 @@ ENV WDIR=/data
 WORKDIR $WDIR
 
 ENV HADOOP_CONF_DIR=/etc/hadoop/conf
-ENV PATH="${PATH}:${WDIR}/CMSSpark/bin"
-
+ENV PATH="${PATH}:/usr/hdp/hadoop/bin/hadoop:/usr/hdp/spark3/bin:/usr/hdp/sqoop/bin:${WDIR}/CMSSpark/bin"
 # CMSMonitoring folder is in WDIR
 ENV PYTHONPATH "${PYTHONPATH}:${WDIR}:${WDIR}/CMSSpark/src/python"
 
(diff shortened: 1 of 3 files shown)


docker/cmsmon-rucio-ds/Dockerfile
```diff
@@ -33,12 +33,13 @@ RUN yum remove -y hbase-bin-2.3 && \
...
     yum clean all &&  rm -rf /var/cache/yum && \
     rm -f /usr/bin/python && ln -s /usr/bin/python3.6 /usr/bin/python && \
-    hadoop-set-default-conf.sh analytix && \
-    source hadoop-setconf.sh analytix 3.2 spark3
-
+    hadoop-set-default-conf.sh analytix && source hadoop-setconf.sh analytix 3.2 spark3
 
 WORKDIR $WDIR
+
+# Run crond
```
vkuznet, line 43: I doubt it is correct approach here since you put hard-coded logic (of crontab.txt) into image. This implies that image will only work if all proper parameters are in place which are not know a-priory for someone who will take the image. I still prefer to create config map in k8s manifest file which will specify all parameters how to run the cron.

docker/cmsmon-rucio-ds/cronjobs.txt
```diff
@@ -0,0 +1,3 @@
+#
+# postStart K8s handler fills the /etc/environment file
+45 6 * * * . /etc/environment; /data/CMSSpark/bin/cron4rucio_datasets_daily_stats.sh --keytab /etc/secrets/keytab --cmsr /etc/secrets/cmsr_cstring --rucio /etc/secrets/rucio --amq /etc/secrets/amq-creds.json --cmsmonitoring /data/CMSMonitoring.zip --stomp /data/stomp-v700.zip --eos /eos/user/c/cmsmonit/www/rucio_daily_ds_stats >> /proc/$(cat /var/run/crond.pid)/fd/1 2>&1
```
vkuznet, line 3: this are all hard-coded parameter which will not allow you to easily debug the issue. For instance, if you'll deploy the image it will only trigger crontab at 6:45, so you'll be stuck to wait till that time to debug it. If you'll keep it in k8s manifest file though you don't have this issue, since someone can easily modify k8s file to test if workflow is running at a time convenient during a test.

kubernetes/monitoring/services/cmsmon-rucio-ds.yaml
```diff
@@ -19,26 +19,6 @@ spec:
       protocol: TCP
       targetPort: 31202
 ---
-kind: ConfigMap
```
vkuznet, line 22: I suggest to keep config map instead of hard-coded parameters. All paths can be changed, include time of cronjob as well.

vkuznet (changes requested):

mrceyhun:
Thanks @vkuznet , it makes sense. Since one line crontab of the script is hard to understand, all parameters are separated in new lines in run.sh using ConfigMap.

PR #1020, opened 2022-04-21
adds decryption capability to deploy secrets script

Panos512:


kubernetes/cmsweb/scripts/deploy-secrets.sh
```diff
@@ -1,24 +1,31 @@
...
     echo "+++ namespaces   : $ns"
+    echo "+++ secretref   : $secretref"
 
     if [ ! -d $conf/$srv ]; then
 	echo "Unable to locate $conf/$srv, please provide proper directory structure like <configuration>/<service>/<files>"
   	exit 1
     fi
+   
+    # Get SOPS decryption key and put it in config directory
+    mkdir -p ~/.config/sops/age/
```
vkuznet, line 26: @Panos512 , unfortunately this is wrong. What if as a user I want to have my own sops/age keys not related to ones we will store to openstack? Therefore without my consent this script will delete my keys. Instead, either use local directory and place there `PWD/.openstack/config/sops/age` area and keys, or rely on common env parameter which should be checked upfront. Moreover, `keys.txt` will be better to replace with `openstack-keys.txt` to distinguish them from `keys.txt` which can be defined and used by end-user.

kubernetes/cmsweb/scripts/deploy-secrets.sh
```diff
@@ -1,24 +1,31 @@
...
+    rm ~/.config/sops/age/keys.txt
```
vkuznet, line 27: I'll remove this line completely, as script should not intervene to do anything in my personal areas. Instead,  put 
```
if [ ! -f $HOME/.openstack/config/sops/age/openstack-keys.txt ]; then
fi
```
block to check for specific file and if necessary download necessary file. Moreover, if we'll need to keep different `$secretref` it would be better to use this name either as part of file or directory name

kubernetes/cmsweb/scripts/deploy-secrets.sh
```diff
@@ -1,24 +1,31 @@
...
+    openstack secret get $secretref --payload_content_type application/octet-stream --file ~/.config/sops/age/keys.txt
```
vkuznet, line 28: this command should be only executed if specific file is not present (but not deleted). If such file is present we should not execute this command.

vkuznet (changes requested):

kubernetes/cmsweb/scripts/deploy-secrets.sh
```diff
@@ -1,24 +1,31 @@
 #!/bin/bash
 # helper script to deploy given service with given tag to k8s infrastructure
 
-if [ $# -ne 3 ]; then
-    echo "Usage: deploy-secrets.sh <namespace> <service-name> <path_to_configuration>"
+if [ $# -ne 4 ]; then
+    echo "Usage: deploy-secrets.sh <namespace> <service-name> <path_to_configuration> <reference_of_decryption_barbican_secret>"
```
muhammadimranfarooqi, line 5: @Panos512 

This secret reference is quite long string. i.e., ``https://openstack.cern.ch:9311/v1/secrets/35145b52-18af-47a3-90d0-1xxxxxxxxxxxxxx` users should not remember this string. Instead based on the `env`, we can use  existing key in this script?

PR #1022, opened 2022-04-22
Merge CRAB tw, publisher, crabserver data pipelne in crab logstash

mrceyhun:
- Remove "crabserver" from cmsweb Logstash and move to CRAB logstash with different MONIT producer and type.
- Arrange CRAB logstash accordingly

@novicecpp It is not tested yet, but I would like to get your ideas. For example, as it is requested in https://its.cern.ch/jira/browse/CMSMONIT-463 , "crabserver" logs type will be "crabhttpcall" and I forced to use "crabhttpcall" tag in filebeat too (same with type to differentiate easily).

After tests, i'll announce its production ready version.

PR #1028, opened 2022-04-28
Update filebeat output to crab logstash, bump filebeat version

novicecpp:
Adjust new tag and change endpoint to crab logstash, according to https://github.com/dmwm/CMSKubernetes/pull/1022
Also, bump filebeat version to 8.1.3.

cc @mrceyhun


mrceyhun:
+1

PR #1065, opened 2022-06-05
Update monitoring cc7 base images for 2022-06-01 release

mrceyhun:
Service updates and dockerfile improvements

- `cmsmon-hadoop-base` image is arranged to use python3.9. `condor-cpu-eff` and `cmsmon-rucio-ds` images arranged accordingly. 
- Other monitoring dockerfiles also updated with new `cern/cc7-base:20220601-1` base image.

fyi @leggerf @brij01

### DIFF
--- a/docker/cmsmon-hadoop-base/Dockerfile
+++ b/docker/cmsmon-hadoop-base/Dockerfile
@@ -1,12 +1,9 @@
-ARG CC7_BASE_VERSION
-
-FROM cern/cc7-base:$CC7_BASE_VERSION
+FROM registry.cern.ch/cmsmonitoring/cmsmon-py:3.9.13
 MAINTAINER Ceyhun Uzunoglu ceyhunuzngl@gmail.com
 
 ARG HADOOP_VERSION
 ARG SPARK_VERSION
 ARG SPARK_MAJOR_VERSION
-#RUN echo hadoop:$HADOOP_VERSION spark:$SPARK_VERSION $SPARK_MAJOR_VERSION
 
 ENV WDIR=/data
 WORKDIR $WDIR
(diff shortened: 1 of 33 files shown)


docker/cmsmon-hadoop-base/Dockerfile
```diff
@@ -21,43 +21,57 @@ ENV PATH $PATH:/usr/hdp/hadoop/bin:/usr/hdp/sqoop/bin
...
-    yum clean all && rm -rf /var/cache/yum &&  \
     touch /etc/hadoop/conf/topology.table.file && ln -s /bin/bash /usr/bin/bashs && \
     echo "32 */6 * * * root ! /usr/sbin/fetch-crl -q -r 360" > /etc/cron.d/fetch-crl-docker
-# add fetch-crl to fetch all sertificates
+    # add fetch-crl to fetch all sertificates
 
-# Copy entrypoint.sh according to spark version
 RUN if [ "$SPARK_MAJOR_VERSION" = "spark3" ] ; then \
-        cp /usr/hdp/spark3/kubernetes/dockerfiles/spark/entrypoint.sh /usr/bin/entrypoint.sh ; \
+# build Python3.9 for Centos7 if spark3
```
vkuznet, line 58: @mrceyhun I suggest to move this python build into separate Dockerfile such that you may re-use it across images. This will also allow to upgrade other Dockerfiles easily, since someone can rebuild python image and re-use it here. I suggest that you'll create new area, e.g. cmsmon-python, and maintain it separately. Then, over here you can use this image and copy python installation (but for that you should make python installation into custom area).

vkuznet (changes requested):

docker/cmsmon-hadoop-base/Dockerfile
mrceyhun: Thanks for the suggestion @vkuznet . I created a new docker image `cmsmon-py` and used it as base image in `cmsmon-hadoop-base`. I searched in the community if it is possible to make multistage build and copy python3 files from the builder. It seems it's not possible and not suggested by [the community](https://stackoverflow.com/questions/71053803/how-to-copy-python-from-one-stage-to-another-during-a-multi-stage-docker-build). That's why i used it as base image.

`cmsmon-py` builts latest LCG release python3 version to be compatible with cern, and current latest LCG py3 version is python3.9 which is important for spark jobs.

vkuznet (approved):

PR #1074, opened 2022-06-08
Adding manifests for global/phys03 migrate and migration servers

muhammadimranfarooqi:


vkuznet (commented):
Few things to fix:
- for migrate deployment we can use lower cpu limit 3000m instead of 8000m
- for migration deployment we can use lower memory limit from 8000Mi to 1000Mi
- for migration deployment you have port 9258 but dbsmigration server is 8251, should we align to 51? or use new port 58?

muhammadimranfarooqi:
thanks @vkuznet 

I have made proposed adjustments

PR #1075, opened 2022-06-08
Adjustment of endpoints for dbs2go migrate/migration

muhammadimranfarooqi:


muhammadimranfarooqi:
> The manifest files conains `/dbs/prod` prefix, while on testbed we have `/dbs/int` one. Does `deploy-srv.sh` script makes appropriate changes during deployment?

Yes, that's why I used `/dbs/prod`

vkuznet:
ok, merging

PR #1103, opened 2022-07-04
adding user in dbs2go Dockerfile

muhammadimranfarooqi:


muhammadimranfarooqi:
@vkuznet 

could you please review. I get permission denied error when I start any script.

vkuznet:
You copied all content into /data area which now owned by root, e.g. here is snapshot from the image I built using your changes
```
drun 4a73cc828ba2
_dbs2go@vocms0181:~$ ls -al
total 18576
drwxr-xr-x. 1 root root       64 Jul  5 18:04 .
drwxr-xr-x. 1 root root        6 Jul  5 18:08 ..
-rw-r--r--. 1 root root      257 Mar 15  2021 config.json
-rwxr-xr-x. 1 root root 10230808 Jul  5 18:04 dbs2go
-rwxr-xr-x. 1 root root      503 Oct 21  2021 monitor.sh
-rw-r--r--. 1 root root      165 Apr  1 18:15 oci8.pc
-rwxr-xr-x. 1 root root  8765440 Jul 26  2021 process_exporter
-rwxr-xr-x. 1 root root     2560 Jul 26  2021 process_monitor.sh
-rwxr-xr-x. 1 root root      979 Mar 15  2021 run.sh
drwxr-xr-x. 7 root root     4096 Jul  5 18:04 static
```
As such you need to do the following:
- change recursively ownership of `/data` area to `_dbs2go._dbs2go`
- ensure that image has access to `/etc/tnsnames`, and ORACLE libraries (this is done I think)
- ensure that `_dbs2go` account can read from `/etc/secrets` where all secret and configuration files are
- ensure that `_dbs2go` account can read `/etc/proxy/proxy`
- ensure that `_dbsgo` account can write to log area which we use in k8s manifest

Therefore, changes may be required not only for image but for k8s manifest files to ensure that `_dbs2go` account can access required files and write to log area.

vkuznet:
since we use filebeat and mount `/data/srv/logs/dbs` in side-car you need to address that. You have two possible solutions here:
- within filebeat image change permission to /data/srv (recursively) to allow anyone to read/write to it, e.g. `chmod -R a+w`,
- or, add `_dbs2go` user to filebeat image and change ownership of `/data/srv/logs/dbs` within filebeat image to `_dbs2go`.

But from your input we must address these issue as well:
```
mkdir: cannot create directory '/data/srv/state': Permission denied
mkdir: cannot create directory '/data/srv/current': Permission denied
ln: failed to create symbolic link '/data/srv/state/dbs/proxy/proxy.cert': No such file or directory
ln: failed to create symbolic link '/data/srv/current/auth/proxy/proxy': No such file or directory
```
This comes from cmsweb rpm based procedure in `run.sh` which we no longer needed in dbs2go service. So, you'll need to remove this from [run.sh](https://github.com/dmwm/CMSKubernetes/blob/master/docker/dbs2go/run.sh#L3--L9)

muhammadimranfarooqi:
Thanks @vkuznet 

Could you please check now `dbs2go-global-r` in new testbed cluster.

vkuznet:
what is the new testbed cluster public URL? https://cmsweb-testbednew.cern.ch? You can place a simple call like that: `https://cmsweb-testbed.cern.ch/dbs2go/datatiers` where you replace URL with proper one, e..g `cmsweb-testbednew.cern.ch/dbs/int/global/DBSReader/datatiers` and it should provide results. I can see `cmsweb-testbednew`, but I can't access dbs on it. Please check FE rules and ensure that we have access to BEs services.

Meanwhile, if you log into curl pod you'll see that it is working fine, eg `curl http://dbs2go-global-r.dbs:9252/dbs/int/global/DBSReader/datatiers` will return proper set of results.

Bottom line, the dbs2go works in newtestbed, but you need to fix FE since we can't access access to BE somehow.

PR #1136, opened 2022-07-29
Add new json exporter for dbs db stats

mrceyhun:
This is related with `CMSMONIT-481`

[json_exporter](https://github.com/prometheus-community/json_exporter) provides nice functionality to convert JSON data to prometheus metrics format. It scrapes [dbstats](https://cmsweb.cern.ch/dbs/prod/global/DBSReader/dbstats) and convert it according to configuration. Service uses clusterIp, so it can be tested in internal network like `curl json-exporter-dbs-db.http:17979/probe?target=https://cmsweb.cern.ch/dbs/prod/global/DBSReader/dbstats`. Prometheus will do the same and scrape it.

@vkuznet since you're in vacation, I'll merge this and apply Prometheus scrape config in monitoring clusters. Then I'll create its aggregation rules to store in VictoriaMetrics. When you're back, I can change it according to your reviews.

### DIFF
--- a/kubernetes/monitoring/services/json-exporter-dbs-db.yaml
+++ b/kubernetes/monitoring/services/json-exporter-dbs-db.yaml
@@ -1,0 +1,112 @@
+kind: Service
+apiVersion: v1
+metadata:
+  name: json-exporter-dbs-db
+  namespace: http
+spec:
+  type: ClusterIP
+  ports:
+  - port: 17979
+    protocol: TCP
+    name: http
+  selector:
+    app: json-exporter-dbs-db
+---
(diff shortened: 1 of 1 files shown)


mrceyhun:
Updating this PR. Since [cmsmon-configs/-/merge_requests/29](https://gitlab.cern.ch/cmsmonitoring/cmsmon-configs/-/merge_requests/29) is ready for vmalert rules, I added dbs-dbinfo aggregation to `vmalert.yaml` which is for 1d/7d/30d aggregation. I thought daily(1d) aggregation is enough but I can add others too if they are required.

PR #1189, opened 2022-08-21
Improve docker hadoop base image to be build in GH

mrceyhun:
Starting to use `gitlab-registry.cern.ch/db/cerndb-infra-hadoop-conf:latest-hdp3` as base image to be able to build `cmsmon-hadoop-base` image in GitHub actions. 

**Advantage**: CERN hadoop team updates this image when there is a new release. We just need to trigger GH actions in CMSSpark.
**Disadvantage**: Image size 2x compared with our previous image(1GB vs 2GB). May be we can request them to create new images like `slim`. We don't need lots of libraries they are installing like yum installed python3. Because, its version is 3.6, we need 3.9 versions to run in K8s environment which does not have access to `cvmfs` and we build our own python3. 

TL;DR: we can build all our spark/hadoop images via GH actions.  Ref https://github.com/dmwm/CMSSpark/pull/105

### DIFF
--- a/docker/cmsmon-hadoop-base/Dockerfile-spark3
+++ b/docker/cmsmon-hadoop-base/Dockerfile-spark3
@@ -1,5 +1,6 @@
 # cmsmon-hadoop-base:spark3-YYYYMMDD
-FROM gitlab-registry.cern.ch/linuxsupport/cc7-base:20220801-1.x86_64
+# Build by CMSSpark GH actions
+FROM gitlab-registry.cern.ch/db/cerndb-infra-hadoop-conf:latest-hdp3
 MAINTAINER Ceyhun Uzunoglu ceyhunuzngl@gmail.com
 
 # Should be full such that includes minor version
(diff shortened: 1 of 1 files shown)


vkuznet:
@mrceyhun , I doubt you need to use CERN IT image. Instead, you need to create your own own and upload it to registry (this should be done manually though but since you create base image very rarely I think it is good compromise. For example, you can set a reminder to rebuild base image only once per year and then use it will be used by CMSSpark all the time). Then, you can use it instead of CERN IT one. Now, about credentials. In order to access registry you need them, but we have common account for that and use it in imagebot on cmsweb. The CMSSpark repo already have them in Settings-Secrets. You just need to use `CERN_LOGIN` and `CERN_TOKEN` env variables in your GitHub Action manifest. For example, please see https://github.com/dmwm/dbs2go/blob/master/.github/workflows/build.yml#L74-L88 how it is done in dbs2go manifest. I build DBS images automatically and they are uploaded to registry.

mrceyhun:
Thanks @vkuznet , I removed base image build to manual way, updated `docker/cmsmon-hadoop-base/build-and-push.sh`. I put new tags `spark3-latest and spark2-latest` to those images that we can use. Tested and working fine.

Additionally, I added MongoDB index creation to `kubernetes/monitoring/datasetmon/cmsmon-rucio-spark2mng.yaml` using `mongosh`.

I think now PR is ready now.

mrceyhun:
Btw, @vkuznet I started to use `docker/build-push-action@v2`[1] action which provides more functionalities.

For example, for dbs2go it can be used like:
```
- name: Build image
  run: |
    curl -ksLO https://raw.githubusercontent.com/dmwm/CMSKubernetes/master/docker/dbs2go/Dockerfile
    curl -ksLO https://raw.githubusercontent.com/dmwm/CMSKubernetes/master/docker/dbs2go/oci8.pc
    curl -ksLO https://raw.githubusercontent.com/dmwm/CMSKubernetes/master/docker/dbs2go/config.json
    curl -ksLO https://raw.githubusercontent.com/dmwm/CMSKubernetes/master/docker/dbs2go/monitor.sh
    curl -ksLO https://raw.githubusercontent.com/dmwm/CMSKubernetes/master/docker/dbs2go/run.sh
    chmod +x run.sh
    sed -i -e "s,ENV TAG=.*,ENV TAG=${{steps.get-ref.outputs.tag}},g" Dockerfile
- name: Login to registry.cern.ch
  uses: docker/login-action@v1
  with:
    registry: registry.cern.ch
    username: ${{ secrets.CERN_LOGIN }}
    password: ${{ secrets.CERN_TOKEN }}
- name: Publish cmsmon-rucio-ds image to registry.cern.ch
  uses: docker/build-push-action@v2
  with:
    context: ./
    file: ./Dockerfile
    push: true
    #build-args: |
    #  TAG=${{ steps.get-ref.outputs.tag }}
    tags: registry.cern.ch/cmsweb/dbs2go:${{ steps.get-ref.outputs.tag }}, registry.cern.ch/cmsweb/dbs2go:latest
```

If you define `ENV TAG` as `ARG TAG`, you can provide it in the context as `build-args`, so you'll not need `sed`. Just a suggestion :)

[1] https://github.com/dmwm/CMSSpark/pull/105/files#diff-35fe72be94bf1af7d42235c3ae05982f07c69808da00468700e15f1a97b84e80R74

Issue #1198, opened 2022-09-01
Deploy Spark crontab jobs as K8s CronJobs


mrceyhun:
As CMS Monitoring we need to move our cron jobs running as crontab both in K8s and VirtualMachines to the K8s `CronJob` kind. All our K8s `CronJobs` that run Spark/Hadoop, should be in `kubernetes/monitoring/cron-hdfs` directory.

- Example test for `cmsmon-rucio-ds` can be found in : https://github.com/mrceyhun/CMSKubernetes/blob/f-cron-hdfs/kubernetes/monitoring/cron-hdfs/cron-test.yaml
- Helpful commands and files for tests can be found in  `/afs/cern.ch/user/c/cuzunogl/public/kyrylo`.

Assignee: @kyrylogy

PR #1203, opened 2022-09-04
Add cmsmon-rucio-ds CronJob

kyrylogy:
This is a K8s `CronJob` to execute `cron4rucio_datasets_daily_stats.sh`  from CMSSpark with all required parameters ready for production. 

#1198

### DIFF
--- a/kubernetes/monitoring/cron-hdfs/cmsmon-rucio-data-ds.yaml
+++ b/kubernetes/monitoring/cron-hdfs/cmsmon-rucio-data-ds.yaml
@@ -1,0 +1,91 @@
+# cron4rucio_datasets_daily_stats.sh for k8s
+kind: ConfigMap
+apiVersion: v1
+metadata:
+  name: cmsmon-rucio-data-ds
+  namespace: cron-hdfs
+  labels:
+    app: cmsmon-rucio-data-ds
+data:
+  run.sh: |
+    #!/bin/bash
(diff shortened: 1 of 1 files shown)


kubernetes/monitoring/cron-hdfs/cmsmon-rucio-ds.yaml
```diff
@@ -0,0 +1,113 @@
...
+      targetPort: 32502
+---
+kind: ConfigMap
+apiVersion: v1
+metadata:
+  name: cmsmon-rucio-ds
+  namespace: cron-hdfs
+  labels:
+    app: cmsmon-rucio-ds
+# Reference: https://www.ibm.com/support/pages/aix-cron-cron-environment-and-cron-job-failures
```
mrceyhun, line 30: We can add a `README.md` file to `cron-hdfs` folder and put this reference and other explanations there. What do you think?

mrceyhun (commented):
Thanks for the PR @kyrylogy , it is nice. I put a few small comments. 

Other than that, It was my suggestion but I thought it would be more convenient to use `port-1` instead of `port-0`. And `port-2` instead of `port-1`. So starting port name index from 1, instead of 0. What do you think? It will be compatible with cron script parameters (`--p1` and `--p2`) too.
If you agree, could you please make necessary name changes?

mrceyhun:
@kyrylogy looks good. So one of them is ready, merged :rocket:

PR #1204, opened 2022-09-06
Create services for cron-hdfs CronJobs

kyrylogy:
Services for cron-hdfs CronJobs are provided in the services.yaml file. New services can be added to the end of the file, but previously used **NodePort** values should be taken into account. 

All references will be provided in the README.md of this folder.

### DIFF
--- a/kubernetes/monitoring/cron-hdfs/README.md
+++ b/kubernetes/monitoring/cron-hdfs/README.md
@@ -1,0 +1,7 @@
+## CMSSpark/bin cronjobs in k8s 
+
+#### References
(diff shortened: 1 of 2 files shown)


mrceyhun:
@kyrylogy everything is perfect, I like your abbreviations. Just one suggestion, `cmsmon-rucio-data-tm` can be [cmsmon-rucio-spark2mng](https://github.com/dmwm/CMSKubernetes/blob/master/kubernetes/monitoring/services/datasetmon/cmsmon-rucio-spark2mng.yaml) which is its current name as a service and carry its purpose IMO.

PR #1206, opened 2022-09-07
Add main CMSSpark cronjobs to k8s

kyrylogy:
This PR includes several Kubernetes CronJob yaml files (not including dbs_condor, aggregation, which are still being refactored)
Schedules and outputs are set according to cronjobs which are running in production. 

Apply only after ensuring that services are created.
#1198

### DIFF
--- a/kubernetes/monitoring/cron-hdfs/cmsmon-hpc-usage.yaml
+++ b/kubernetes/monitoring/cron-hdfs/cmsmon-hpc-usage.yaml
@@ -1,0 +1,85 @@
+kind: ConfigMap
+apiVersion: v1
+metadata:
+  name: cmsmon-hpc-usage
+  namespace: cron-hdfs
+  labels:
+    app: cmsmon-hpc-usage
+data:
+  run.sh: |
+    #!/bin/bash
+    # Get env variables, since cron's process environment is different than shell's one
+    . /etc/environment
+    # Run and print all results(except for Spark logs) to stdout
+    /data/CMSSpark/bin/cron4hpc_usage.sh \
+      --keytab /etc/secrets/keytab --output /eos/user/c/cmsmonit/www/hpc_usage \
+      --url https://cmsdatapop.web.cern.ch/cmsdatapop/hpc_usage \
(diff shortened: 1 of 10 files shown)


kubernetes/monitoring/cron-hdfs/cmsmon-hpc-usage.yaml
```diff
@@ -0,0 +1,85 @@
...
+apiVersion: batch/v1
+kind: CronJob
+metadata:
+  name: cmsmon-hpc-usage
+  namespace: cron-hdfs
+spec:
+    schedule: "43 13 * * *"
+  concurrencyPolicy: Forbid
+  failedJobsHistoryLimit: 1
+  jobTemplate:
```
mrceyhun, line 28: Last 3 lines should be aligned with `schedule: "43 13 * * *"`

mrceyhun (commented):
Thanks @kyrylogy , it is very nice and neat. I put 2 small comments.

Other than that, there is one issue for `cmsmon-rucio-ds-sum.yaml`. It is running `cron4rucio_datasets_daily_stats.sh` script. However its name entails to run `cron4rucio_ds_summary.sh`. So, my suggestion is that please ignore `cron4rucio_ds_summary.sh` K8s CronJob, don't put it; because we did not give final decision about it yet. But, we need to prepare CronJob for `cron4rucio_datasets_daily_stats.sh`. Its K8s CronJob name can be `cmsmon-rucio-daily-stat` ? IMO, when you refactor the `cmsmon-rucio-ds-sum.yaml` (change the file name and all the confgimap, secret, CronJob name to `cmsmon-rucio-daily-stat`), everything will be in place.

Very good job!

kyrylogy:
Thanks a lot for reviewing, I will fix everything mentioned above, but I'm not sure about this one:

> change the file name and all the confgimap, secret, CronJob name to `cmsmon-rucio-daily-stat`
I wanted to stick with the same naming convention for rucio jobs, and it is already like that in services.yaml

mrceyhun:
Closes #1198

mrceyhun:
Close #1198

PR #1213, opened 2022-09-14
check-metric helm chart for WMA and DBS namespaces.

arooshap:


helm/check-metric/values.yaml
```diff
@@ -0,0 +1,111 @@
...
+nameOverride: ""
+fullnameOverride: ""
+
+args:
+  dbs: 
+  - /data/check-metric
+  - -url
+  - "http://prometheus-service.monitoring.svc.cluster.local:8080"
+  - -metric
+  - "dbsmigration_process_open_fds,dbs_migrate_exporter_process_open_fds,dbs_global_exporter_process_open_fds,dbs_globalW_exporter_process_open_fds,dbs_phys03R_exporter_process_open_fds,dbs_phys03W_exporter_process_open_fds"
```
vkuznet, line 23: This should be fixed since  we no longer use dbs metrics since we decommissioned dbs services and use instead dbs2go. And, for dbs2go we do not have so far any outages, therefore i doubt we need it. As such I doubt we need dbs section here per-se.

vkuznet (changes requested):
In addition to comment I put in this PR I think we should generalize this template to all services since python and go-based services have similar kind of metrics. For instance, in python based servics, like reqmgr2, ms-XXX, etc we have metrics similar to dbs like, reqmgr2_open_fds, etc. Therefore, we can apply the rules to all python based services with these metrics.

While for Go-based services, we have metrics like xxx_total_connections which we can use for all Go-based services.

This is a suggestion and it requires knowledge of all prefixes we use for python and go-based services. But once we know all prefixes, e.g. dbs_global_exporter, or wma_exporter, etc then we can generalize this template.

Therefore, so far I suggest only remove dbs (as we no longer run dbs python services) and add wma, but later add a new ticket to generalize this template to all services.

muhammadimranfarooqi:
Hi, 

I have created separate issue to generalize this chart. https://its.cern.ch/jira/browse/CMSKUBERNETES-182

arooshap:
@vkuznet I have removed the DBS section as per your suggestions, and so far, we only have args for wmarchive.

As for inculcating the other services, I will get started on that.

Issue #1214, opened 2022-09-15
Create helm charts for cms monitoring spark running cron jobs


mrceyhun:
Eventually we will automate our deployment process with FluxCD and helm charts. Currently, our k8s CronJobs are ready to run and we need to prepare their Helm charts.

### How

[CronJobs](https://github.com/dmwm/CMSKubernetes/tree/master/kubernetes/monitoring/cron-hdfs) have common definitions like `concurrencyPolicy`, `failedJobsHistoryLimit `, `backoffLimit `. And also there are changing values and they should be arranged in helm. In templates, we can have cronjob.yaml and service.yaml initially. We can evaluate to get rid of ConfigMaps and to provide commands directly in values.yaml for each cron.

We can use values.yaml structure like in this SO entry https://stackoverflow.com/a/73571078/6123088 . An example can be found in here https://github.com/mrceyhun/CMSKubernetes/blob/f-cmsmon-crons-helm/helm/cmsmon-cron/values.yaml .

### Tasks

1. Create first version of helm charts for cmsmon cronjobs (kyrylo) 
2. Parametrize for `prod` and `test` (kyrylo)
3. Saving cron logs in S3 using fluentd (ceyhun)
4. Conditioning EOS volume mount according to cronjob (some of them does not need) (kyrylo)
5. Implement ingress for port management (ceyhun)
6. Evaluate using SOPS secret management using cmsweb sops service (ceyhun)

### Helpful documentations:

- https://kubernetes.docs.cern.ch/docs/automation/gitops/
- https://gitlab.cern.ch/kubernetes/automation/releases/gitops-getting-started
- https://cms-http-group.docs.cern.ch/helm/deploy-services/
- https://cms-http-group.docs.cern.ch/k8s_cluster/sops_age/

@kyrylogy Let's share the tasks in this issue. As we talked, just create the initial version of helm chart and later we can decide our direction. I though that we may give priority to the first task over CMSSpark tasks ;)

mrceyhun:
fyi @leggerf @brij01

mrceyhun:
More details will be provided in [CMSMONIT-525](https://its.cern.ch/jira/projects/CMSMONIT/issues/CMSMONIT-525)

Closing...

PR #1215, opened 2022-09-18
Create a helm chart for cmsmonitoring spark running cronjobs

kyrylogy:
#1214 @mrceyhun 
This helm chart simplifies deployment of multiple spark running cronjobs.
### Notes:
- Replaced `configMap.yaml`  with a script template in [_helpers.tpl](https://github.com/kyrylogy/CMSKubernetes/blob/3738f28285440bc074c7cdf9e7fbb2f980ce7a17/helm/cmsmon-cron/templates/_helpers.tpl#L65)
Command and arguments for each cronjob can be specified individually in corresponding fields.
- Added [test](https://github.com/kyrylogy/CMSKubernetes/blob/3738f28285440bc074c7cdf9e7fbb2f980ce7a17/helm/cmsmon-cron/values.yaml#L5) field to run cronjobs with `--test` flag
- EOS Access is granted to cronjobs with `eosEnabled` parameter
- **spark.driver** and **spark.driver.blockManager** pulled out to service field
- NodePorts are assigned according to the amount of cronjobs specified 
(However, I suggest looking at https://helm.sh/docs/chart_template_guide/functions_and_pipelines/#using-the-lookup-function to figure out which ports we can use and building on that)
### Install:
For now, directories for test output are not replaced by **test.Enabled** parameter so it should be specified directly in each cronjob.
`helm install crons cmsmon-crons/ --debug`
`helm list`
. . .
`helm uninstall crons`

### DIFF
--- a/helm/cmsmon-cron/.helmignore
+++ b/helm/cmsmon-cron/.helmignore
@@ -1,0 +1,23 @@
+# Patterns to ignore when building packages.
+# This supports shell glob matching, relative path matching, and
+# negation (prefixed with !). Only one pattern per line.
+.DS_Store
+# Common VCS dirs
+.git/
+.gitignore
+.bzr/
+.bzrignore
+.hg/
+.hgignore
+.svn/
+# Common backup files
+*.swp
+*.bak
+*.tmp
+*.orig
+*~
+# Various IDEs
+.project
+.idea/
+*.tmproj
+.vscode/
--- a/helm/cmsmon-cron/Chart.yaml
+++ b/helm/cmsmon-cron/Chart.yaml
@@ -1,0 +1,24 @@
+apiVersion: v2
+name: cmsmon-cron
+description: A Helm chart for k8s CMSSpark cronjobs
+
+# A chart can be either an 'application' or a 'library' chart.
+#
+# Application charts are a collection of templates that can be packaged into versioned archives
+# to be deployed.
+#
+# Library charts provide useful utilities or functions for the chart developer. They're included as
+# a dependency of application charts to inject those utilities and functions into the rendering
+# pipeline. Library charts do not define any templates and therefore cannot be deployed.
+type: application
+
+# This is the chart version. This version number should be incremented each time you make changes
+# to the chart and its templates, including the app version.
+# Versions are expected to follow Semantic Versioning (https://semver.org/)
+version: 0.1.0
+
+# This is the version number of the application being deployed. This version number should be
+# incremented each time you make changes to the application. Versions are not expected to
+# follow Semantic Versioning. They should reflect the version the application is using.
+# It is recommended to use it with quotes.
+appVersion: "1.16.0"
--- a/helm/cmsmon-cron/README.md
+++ b/helm/cmsmon-cron/README.md
@@ -1,0 +1,25 @@
+This Helm chart provides all-in-one template for CMSSpark cronjobs. 
+
+### Notes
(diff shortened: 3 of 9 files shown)


mrceyhun:
@kyrylogy very good! I will test this tomorrow.

- It's a perfect idea to get rid of configmaps and inject run commands in `_helpers.tpl`. May be we don't need crontab environment variables workaround. Command will run in the same environment with the pod, so no need to save envs to file and then source again. Need a test though.
- `--test` field is good idea but running a test in each 1 minute is not :) Some tests may take more than 30 minutes. So my suggestion is that change the test schedule to `1 hour` for now and put a `TODO` comment to make it more applicable by looking to community versions for tests. Personally, someone like me does not want to wait long to see the test result.
- `eosEnabled` perfect!
- About nodePorts. We can find all the used NodePorts with `lookup` function, for example jq command is like `kubectl get svc -A -o json | jq '.items[].spec.ports[].nodePort'`. However, it is not the best solution IMO. One worst case scenario: let's say a service is deleted but will be created within 1 minute, helm look function returned that port as free and cron job is bound to that port. So previous service will fail. Therefore, for now, let's assign ports statically, and start range from `32500`. Later we'll use Ingress to solve this issue.
- One last request, can you put a NOTES.txt to produce pretty logs and also README.md which gives brief overview.

Thanks, it is very good :rocket:

mrceyhun:
@kyrylogy I found the Spark problem. It is hard to understand and I really don't know the reason behind it but we need to give same port numbers to NodePort and container port. So, we cannot use static `5001` and `5101` ports. 

What we can do:

In K8s, each service port has its own env variable. Let's say our service name is `cmsmon-rucio-daily`. And we defined its service like
```
kind: Service
apiVersion: v1
metadata:
  name: cmsmon-rucio-daily
spec:
  selector:
    app: cmsmon-rucio-daily
  type: NodePort
  ports:
    - name: port-1 # spark.driver.port
      nodePort: 32501
      port: 32501
      protocol: TCP
      targetPort: 32501
    - name: port-1 # spark.driver.blockManager.port
      nodePort: 32502
      port: 32502
      protocol: TCP
      targetPort: 32502
 ```
 Inside the pod, you can find these env variables(`printenv`) : `CMSMON_RUCIO_DAILY_SERVICE_PORT_PORT_1` AND `CMSMON_RUCIO_DAILY_SERVICE_PORT_PORT_2`. This convention never changes.

Therefore, my suggestion is that, remove `service` part in values.yaml which sets 5001 and 5101 as ports, instead use these environment variables to sat `--p1 and --p2`. As i understand from the last part of `_helpers.tpl`, we can get service name via `cron.name`. Our port names will not change (port-1 and port-2). So last part of service port env variable will be same as `_SERVICE_PORT_PORT_1` or `_2`. 

And also we need to set same ports in cronjob.yaml container ports with the service ports. That's why, let's join service and cronjob template in one template `cronjob.yaml` to be able to get same ports from for loop.

I put the working example here https://gist.github.com/mrceyhun/d072f365ee2dbe5e763a286d72005135 Could you please arrange values.yaml accordingly, remove unnecessary service part etc. Helm is good, allows us hack lots of things but let's put proper documentation for all these tricks :)

mrceyhun:
One last request, in the README.md, can you put a TODO to migrate to google's spark k8s operator : https://github.com/GoogleCloudPlatform/spark-on-k8s-operator

kyrylogy:
@mrceyhun I think I added everything you asked for :) I splitted service in cronjob again because of adding a new yaml file which will deploy all the same but with Job type, i.e. it executes immediately when testing is needed. There could be a simpler way - creating a Helm post-hook that will run `kubectl create job --from=cronjob...` but there are some limitations on Helm and K8s side that block this option. Thanks for providing a way to get env variables!

mrceyhun:
@kyrylogy is this done, it looks okay for me.

PR #1218, opened 2022-09-22
Remove unused secret eos-krb

micsucmed:


muhammadimranfarooqi:
Hi @micsucmed 

I think these secrets are required to mount eos space?

muhammadimranfarooqi:
As per doc.

https://cms-http-group.docs.cern.ch/k8s_cluster/eos/#adjusting-services-manifest-file

micsucmed:
Hi @muhammadimranfarooqi,
I think you are right, but remember we used newdqmgui-secrets as the kerberos key? How should this be handled then?

micsucmed:
Should we still add an eos-krb secret to the namespace along side the newdqmgui-secret?

muhammadimranfarooqi:
Ok. Then there is no need to create a separate secrets. 

Cheers

micsucmed:
@muhammadimranfarooqi an error of eos-krb not being found is still appearing, maybe this secret should be used by default with eos mount. I think using both secrets is necessary. Could we rollback this commit?

muhammadimranfarooqi:
Ok. 
I have rolled back it https://github.com/dmwm/CMSKubernetes/pull/1220

PR #1235, opened 2022-10-20
Updating cmsweb and cmsweb-base images + Updating cmsweb and cmsweb-base tags in docker files.

arooshap:


muhammadimranfarooqi:
Looks good. Thanks

PR #1237, opened 2022-10-30
Increase polling cycle of liveness probes

goughes:
This PR increases the timeout from 1 second to 5 seconds for each service and increasing the polling cycle to 30 seconds.

Fixes https://github.com/dmwm/WMCore/issues/10223


amaltaro:
@goughes these changes look good to me. However, I wonder if we should revert these changes made for:
* t0wmadatasvc (owned by the T0 team)
* wmstats (Go-based prototype created by Valentin, still only available in testbed)

PR #1238, opened 2022-11-01
Pins certifi dependency to Python 2.7 compatible version

Panos512:


vkuznet (changes requested):
Panos, could you please remove single quotes and explain why it should be less equal?

Panos512:
Thanks Valentin.  I removed the quotes and added `==`.

The problem is that `requests` has a dependency on `certifi` and the latest certifi version is not compatible with Python 2.7. This breaks Python in our VMs. The change makes sure we use the last Python 2.7 compatible version of certifi.

vkuznet:
Panos, you realize that using old certifi leads to security flaw as they will be less secure. On top of the I want to understand why we need to care about python 2.7 as we should not use it at all. In other words, what is exactly required to have python 2.7 and why we can't upgrade it to recent python 3.X version?

PR #1252, opened 2022-11-18
Persistant kerberos authentication on container with K5start

micsucmed:


kubernetes/cmsweb/services/newdqmgui.yaml
```diff
@@ -61,7 +61,7 @@ spec:
         fsGroup: 2000
       containers:
       - image: registry.cern.ch/cmsweb/newdqmgui #imagetag
-#      - image: msuccarm/new_dqmgui:4.0 #imagetag
+#      - image: msuccarm/new_dqmgui:9.0@sha256:216a2cdb309ed2fb0d0d21ce6ecb5433432a5550cc3d44df1a4e5e0efa66c807 #imagetag
```
muhammadimranfarooqi, line 64: Hi @micsucmed 

We have `#imagetag`  that is replaced with the relevant tag during deployment using `deploy-srv.sh` script. So, I suggest to remove this part `:9.0@sha256:216a2cdb309ed2fb0d0d21ce6ecb5433432a5550cc3d44df1a4e5e0efa66c807` from this line. Thanks

micsucmed: Hi @muhammadimranfarooqi 
These line is for my own local debugging with my docker hub images on the testbed cluster that's why its commented out, the official use should be the line above which refers to the harbour registry. I hope is ok to keep this line as it is.

PR #1253, opened 2022-11-18
Fix ReqMgr2MS configuration file names in PyPi run.sh

amaltaro:
While trying to test the PyPi-based WMCore images discussed in this WMCore ticket:
https://github.com/dmwm/WMCore/issues/11354#issuecomment-1319168170

I noticed that all of the Microservices PODs fail to start (`CrashLoopBackOff`) and inspecting the logs, this is what is returned:
```
amaltaro@lxplus8s19:~/private/cmsweb-deploy/CMSKubernetes/kubernetes/cmsweb $ kubectl logs ms-transferor-7558cb86f6-hsb6t -n dmwm
...
/usr/local/bin/wmc-httpd: /etc/secrets/config.py: invalid configuration file
tail: cannot open '/data/srv/logs/reqmgr2ms/*.log' for reading: No such file or directory
```

With these changes we should be able to start those microservices successfully. I haven't tested it though because I think a new docker image is required(?)

@vkuznet can you please review and merge it? Thanks

### DIFF
--- a/docker/pypi/reqmgr2ms-monitor/run.sh
+++ b/docker/pypi/reqmgr2ms-monitor/run.sh
@@ -4,7 +4,7 @@
 srv=`echo $USER | sed -e "s,_,,g"`
 STATEDIR=/data/srv/state/$srv
 LOGDIR=/data/srv/logs/$srv
-CFGFILE=/etc/secrets/config.py
+CFGFILE=/etc/secrets/config-monitor.py
 
 mkdir -p /data/srv/logs/$srv
 mkdir -p /data/srv/state/$srv
--- a/docker/pypi/reqmgr2ms-output/run.sh
+++ b/docker/pypi/reqmgr2ms-output/run.sh
@@ -4,7 +4,7 @@
 srv=`echo $USER | sed -e "s,_,,g"`
 STATEDIR=/data/srv/state/$srv
 LOGDIR=/data/srv/logs/$srv
-CFGFILE=/etc/secrets/config.py
+CFGFILE=/etc/secrets/config-output.py
 
 mkdir -p /data/srv/logs/$srv
 mkdir -p /data/srv/state/$srv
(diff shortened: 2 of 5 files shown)


vkuznet:
Alan, I rebuilt and uploaded new images. But you can do it on your own (and in fact I encourage dev-team to learn how to do it). On a machine with docker I did the following:
```
# step 1: clone CMSKubernetes
# step 2: cd CMSKubernetes/docker
# step 3: build images
ls pypi/ | grep -v gfal | awk '{print "docker build -t registry.cern.ch/cmsweb/"$1":pypi pypi/"$1""}' | /bin/sh
# optional step check login to registry
docker login registry.cern.ch
# here you will provide your CERN user name and need a CLI secret which you can obtain from visiting
# registry.cern.ch, click on your profile and copy CLI secret
# step 4: upload images to registry
docker images | grep registry | grep -v gfal | sort | uniq | awk '{print "docker push "$1":pypi"}' | sort | uniq | /bin/bash
```

Please note, that this will create and upload images with tag `pypi`. In a future, we need to decide on naming convention to tags. Since we upload so far images with the same tag `pypi` on k8s we need to ensure that latest images is always picked up. For that k8s manifest files should have `imagePullPolicy: Always` (e.g. https://github.com/dmwm/CMSKubernetes/blob/master/kubernetes/cmsweb/services/dbs2go-global-r.yaml#L71). Unfortunately, for python based images we do not use it since their size is too large and it lead to longer start-up time as new image always pulled from the repo. Therefore, I rebuilt images with `pypi-20221119` tag and uploaded them to registry:
```
ls pypi/ | grep -v gfal | awk '{print "docker build -t registry.cern.ch/cmsweb/"$1":pypi-20221119 pypi/"$1""}' | /bin/sh
docker images | grep registry | grep pypi-20221119 | grep -v gfal | sort | uniq | awk '{print "docker push "$1":pypi-20221119"}' | sort | uniq | /bin/bash
```
So, you may use new tag `pypi-20221119` which contains these changes. 

If we want to re-use latest tag, then we'll need to adjust all manifest files to include `imagePullPolicy: Always` policy, but as I said it will lead to longer time for image to start-up and so far I choose not to do that for python based images. Therefore, our option should be always create new tag for pypi images and upload them to registry.

PR #1254, opened 2022-11-20
Improve cmsmon spark docker images and apply in services

mrceyhun:
- Because of the shared requirements, all Spark required docker images merged into `cmsmon-spark` and `cmsmon-spark2mng`. Base `cmsmon-hadoop-base` image will be same which includes all base requirements. They replace condor-cpu-eff, cmsmon-rucio-ds and cmsmon-rucio-spark2mng images. Further spark related images will be build up on them.
- Our acron jobs in cmsmonit user are mostly run monthly and their deployment not easy to run CronJob(debugging, logging). And they take so long time even in test. That's running all of them as cron job in a deployment makes sense. May be we can deploy them as cron job after they are matured in K8s environment.


mrceyhun:
I need to merge this PR to be able to use them in CMSSpark GH workflows. All reviews are welcomed even though it is merged.

PR #1258, opened 2022-11-21
Update WMCore microservices docker image name

amaltaro:
Changes related to the WMCore migration to pypi packaging, currently being discussed in:
https://github.com/dmwm/WMCore/issues/11354

This should not be merged until we can properly test these new docker image.

### DIFF
--- a/kubernetes/cmsweb/services/reqmgr2ms-monitor.yaml
+++ b/kubernetes/cmsweb/services/reqmgr2ms-monitor.yaml
@@ -86,7 +86,7 @@ spec:
         runAsGroup: 1000
         fsGroup: 2000
       containers:
-      - image: registry.cern.ch/cmsweb/reqmgr2ms #imagetag
+      - image: registry.cern.ch/cmsweb/reqmgr2ms-monitor #imagetag
         name: ms-monitor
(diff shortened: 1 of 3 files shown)


vkuznet:
Alan, I'm not sure it is backward compatible, since we use RPM based image `reqmgr2ms` for all services, while pypi images are unique to the service. So if we'll merge this then we'll break RPM based image until we create similar images like `reqmgr2ms-output` in RPM land as well. What we need to complement this PR is to create alias images `reqmgr2ms` RPM based image should have alias (or clone) `reqmgr2ms-output`, etc.

amaltaro:
I do not think we need to make it backwards compatible. Reason why I said: "This should not be merged until we can properly test these new docker image."

IMO, we test things out and if we are happy with them, we go ahead and merge it. We are not making these changes planning to go back anyways.

vkuznet:
@amaltaro , I think we can merge this PR now since we do have both RPM and pypi based images which you proposed to use in this PR, are you agree?

amaltaro:
@vkuznet yes Valentin, I think we can merge this now. Thanks

vkuznet (approved):

PR #1267, opened 2022-11-30
Add cron-py to run eos_path_size and rucio quotas jobs

mrceyhun:
Runs https://github.com/dmwm/CMSMonitoring/pull/183

- We need CVMFS access, so its PVC and SC are defined.


mrceyhun:
> @mrceyhun , could you please come up with better name instead of `cron-py` since it is misleading. I rather name it like `cron-size-quotas` or something similar.

Thanks for the suggestion @vkuznet , `cron-size-quotas` makes sense and I replaced `cron-py` with it.

mrceyhun:
Rucio Python and others working fine. Deployed to prod, merging..

PR #1269, opened 2022-12-01
Allow to use local services area via optional parameter

vkuznet:


amaltaro (commented):
Valentin, I have a general comment on this script which is, whenever I have to use it, I don't pass the `<env>` option and I simply rely on it to be self-discovered by the script and my environment. In that scenario, we might have a mix up of parameters 3 and 4 (aka. positional arguments).

So, I think to have a clear and solid script option, we should either:
a) remove the `env` option and always rely on the environment for the correct execution
b) or, make it keyword arguments, e.g. `-t <tag> -e <env> -p <services_path>` etc

What do you think? @muhammadimranfarooqi and @arooshap might have a different view as well.

muhammadimranfarooqi:
Thanks @vkuznet 

Changes look fine. @amaltaro I don't think this scripts works without `env` as all three parameters are compulsory to run the script

amaltaro:
@muhammadimranfarooqi from line 4, we need to provide >= 2 parameters. Thus, running it only with the service name and the tag works just fine.

arooshap:
@amaltaro Yes, this can cause a mixup of parameters 3 and 4. I think rather than removing the `env` variable entirely, we can make the script take keyword arguments because when you're doing multiple deployments at the same time, it helps to make sure that you are deploying in the correct environment.

kubernetes/cmsweb/scripts/deploy-srv.sh
```diff
@@ -2,7 +2,7 @@
 # helper script to deploy given service with given tag to k8s infrastructure
 
 if [ $# -lt 2 ]; then
-     echo "The required parameters for service and tag are missing. Please use deploy-srv.sh <service> <tag> <env> "
```
arooshap, line 5: @amaltaro  Yes, this can cause a mixup of parameters 3 and 4. I think rather than removing the env variable entirely, we can make the script take keyword arguments because when you're doing multiple deployments at the same time, it helps to make sure that you are deploying in the correct environment.

vkuznet:
I implemented script variables, please see updated code. For simplify here is a stand-alone script to demonstrate how it works (I name this script as t.sh):
```
#!/bin/bash

function Usage()
{
     echo "t.sh -s <service> -t <tag> -e <env> -d <optional services directory>"
     exit 1;
}
service=""
tag=""
env=""
sdir=""

while getopts "s:t:e:d:" opt; do
    case $opt in
    s)
        service=$OPTARG
        ;;
    t)
        tag=$OPTARG
        ;;
    e)
        env=$OPTARG
        ;;
    d)
        sdir=$OPTARG
        ;;
    *)
        Usage
        ;;
    esac
done
if [ "$service" == "" ]; then
    echo "missing service value"
    Usage
fi
if [ "$tag" == "" ]; then
    echo "missing tag value"
    Usage
fi
if [ "$env" == "" ]; then
    echo "missing env value"
    Usage
fi
echo "service=$service tag=$tag env=$env sdir=$sdir"
```
If you'll run it you'll see something like this:
```
# the case if we miss mandatory input parameter
./t.sh -e test -s dbs
missing tag value
t.sh -s <service> -t <tag> -e <env> -d <optional services directory>

# the case when all mandatory parameters are in place
./t.sh -e test -s dbs -t HG123
service=dbs tag=HG123 env=test sdir=
```

@muhammadimranfarooqi , @arooshap please review and **test** this changes by yourself in your own environment and report if everything is fine and you are satisfied with them (i.e. I mean try your wokrflows and test with multiple setups).

@amaltaro please provide your feedback as user if you are satisfied with proposed solution.

arooshap:
Hi @vkuznet,

I have tested this script separately on many instances, and it works as expected.

arooshap:
@vkuznet @amaltaro would you like to merge this PR or keep it open?

vkuznet:
@arooshap , I'm in favor of merging this PR but I will let Alan to confirm if it is ready or will require further adjustments.

PR #1270, opened 2022-12-01
Deploy cert-checker to monitoring cluster to check all certs keytabs

mrceyhun:
Cert-checker implementation to check all monitoring certificates and keytabs. Deployed in cmsmonitnew-cluster-vef5nqhek4nf-node-0:30901 . Metrics: `cert_valid_sec` , `keytab_valid_sec`.

Used repo: https://github.com/vkuznet/cert-checker with some modifications in https://github.com/vkuznet/cert-checker/pull/1.

### DIFF
--- a/kubernetes/monitoring/services/cert-checker.yaml
+++ b/kubernetes/monitoring/services/cert-checker.yaml
@@ -1,0 +1,103 @@
+# Ref: kubernetes/cmsweb/services/cert-checker.yaml
+kind: Service
+apiVersion: v1
+metadata:
+  name: cert-checker
+  namespace: http
+spec:
+  type: ClusterIP
+  ports:
+    - port: 8888
+      protocol: TCP
+      name: http
+  selector:
+    app: cert-checker
+---
+kind: ConfigMap
+apiVersion: v1
+metadata:
+  name: cert-checker
+  namespace: http
+  labels:
+    app: cert-checker
+data:
+  configs.json: |
+    [
+      {"cert": "/etc/secrets/robotcert.pem", "ckey": "/etc/secrets/robotkey.pem"},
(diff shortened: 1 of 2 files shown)


kubernetes/monitoring/services/cert-checker.yaml
```diff
@@ -0,0 +1,103 @@
...
+    metadata:
+      labels:
+        app: cert-checker
+      annotations:
+        prometheus.io/scrape: 'true'
+        prometheus.io/path: "/metrics"
+        prometheus.io/port: "8888"
+    spec:
+      containers:
+        - image: registry.cern.ch/cmsmonitoring/cert-checker:test
```
vkuznet, line 57: please do not use concrete image tag and rather replace `:test` with ` #imagetag` which we use in `deploy-srv.sh` script when we deploy services

vkuznet (changes requested):
in addition, I ask @muhammadimranfarooqi to review these changes too

mrceyhun:
@vkuznet I applied the changes you suggested in https://github.com/vkuznet/cert-checker/pull/1 and create a new image with test tag again. It is deployed now. After my PR merged into your repo, we can continue to use `cmsweb/cert-checker` with new build. Then I can remove `cmsmonitoring/cert-checker` one which is just for test.

You can see that empty string keytab and certs are removed from the configs.json. Waiting for your green light and new `cmsweb/cert-checker` build with my PR changes. Then it will be deployed to production.

CMSWEB probably will use extra parameters like `#imagetag`, `env` label in service definition as you said. One thing is that, each request checks all the certs and keytabs , therefore we should not set livenessProbe period as 3 seconds. I set 600s, but above 30s is fine IMO.

vkuznet:
@mrceyhun please feel free to proceed, I merged your PR.

vkuznet:
@muhammadimranfarooqi , please review and merge this PR.

PR #1272, opened 2022-12-02
Update build.sh script to correct an error.

arooshap:
There is a wrong operator used here, which does the opposite of what we need to do. I came across this while building an image for `workqueue`.


arooshap:
@vkuznet could you please merge this?

PR #1273, opened 2022-12-05
Using kube-state-metrics to configure alerts for dmwm namespace.

arooshap:
In order to set up alerts for the dmwm namespace, some alerting rules have been configured in this PR.

### DIFF
--- a/helm/dbs2go-global-r/Chart.yaml
+++ b/helm/dbs2go-global-r/Chart.yaml
@@ -15,7 +15,7 @@ type: application
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
-version: 0.1.2
+version: 0.1.3
 
 # This is the version number of the application being deployed. This version number should be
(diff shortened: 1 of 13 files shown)


arooshap:
@mrceyhun could you please have a look at these rules?

mrceyhun:
@arooshap I don't know if first rule will help much. But second rule seems okay, just I suggest to decrease `10m` to `1m` or `5m`. You can see your rule expressions from here too: https://monit-grafana.cern.ch/d/jnqK7LK4z/user-aroosha-prom-rules-test?orgId=11

I could not find a way to add node name to metrics but you should put it to alerts ($labels.host). For cmsweb services, `tag` should be `cmsweb`. `service` name is important to define which emails/Slack will get the notification. See examples: https://github.com/dmwm/CMSKubernetes/blob/master/kubernetes/cmsweb/monitoring/prometheus/rules/wmarchive.rules#L28-L32 . Depending on the receiver email, I will modify [alertmanager.yaml](https://gitlab.cern.ch/cmsmonitoring/cmsmon-configs/-/blob/master/alertmanager/alertmanager.yaml) (this is my task)

PR #1274, opened 2022-12-07
Create monitoring deploy-ha.sh

mrceyhun:
New `monitoring/deploy-ha.sh` script is created by borrowing logics from `deploy.sh`(outdated)
- It only covers HA cluster deployments.
- For secret management, it depends on `deploy-secrets.sh` which is production script to manage our secrets.
- Without modifying `deploy-secrets.sh`, deploy-ha.sh performs sed replacement for `sdir` and `cdir`, writes modified `deploy-secrets.sh` to different temporary file to use in secret creatins.
- Deploy logic is simple but not provide much options. My initial aim is to use it while creating a cluster.

Additionally:
- I always forgot to create ServiceAccounts of `proxy-account` and `kerberos-account` while creating proxy and kerberos CronJobs. So, i moved them to CronJob configs.

deploy-secrets changes:
- Because of starting to use single keytab folder and copying keytabs from these folders for specific secrets in `secrets/-/merge_requests/42`, secrets are modified accordingly.
- `hdfs-secrets` seemed a redundant one which has no usage: `k describe pods,cj,cm -A | grep "hdfs-secrets"`.
- Fixed `proxy-secrets` with better error messages and explanations.


vkuznet (approved):

vkuznet:
@mrceyhun , even though I approved this PR it has a conflict files, please resolve it.

mrceyhun:
Thanks @vkuznet , they are resolved now.

Cert-checker update:
- In the second commit, i added certchecker to HA deployment script
- Its name changed from `cert-checker.yaml` to `http-exporter-certcheck.yaml` to be consistent with other exporters and for easy deployments.
- Its prometheus job config is on the way.
- `registry.cern.ch/cmsmonitoring/cert-checker` will be updated with latest changes and now it has `20221207` tag in cmsmonitoring registry.

PR #1286, opened 2022-12-13
Prometheus update to include rules for pod restart and reason for the dmwm namespace.

arooshap:


arooshap:
@muhammadimranfarooqi could you please merge this?

PR #1288, opened 2022-12-14
Consolidate run.sh script for the pypi docker images

amaltaro:
Related to https://github.com/dmwm/WMCore/issues/10178

I found that the `reqmgr2-tasks` service, PyPi-based, had a few CherryPy threads down with the following error [1], so I decided to refactor this run.sh script. Some highlights are:
* consolidate directory names and use the same shell variable everywhere
* robot and proxy certificates are placed under the same directory (AUTHDIR)
  * Note: I am not sure whether we should be exporting the X509_USER_PROXY variable though, I would be inclined not to.
* why the hmac file has to be placed under the /auth directory?
* removed tail -f logs that was executed at the very end (?)

@vkuznet can you please clarify some of these questions? It's not clear to me why and whether we actually need them.

[1]
```
[13/Dec/2022:21:37:43]  INFO: instantiating extension statusChangeTasks
advanceStatus
[13/Dec/2022:21:37:43]  INFO: instantiating extension parentLock
fetchIncludeParentsRequests
ERROR:root:DBSReaderError
Message: Error in DBSReader with DbsApi
ClientAuthException:'key or cert file does not exist: /data/srv/state/reqmgr2/proxy/proxy.key, /data/srv/state/reqmgr2/proxy/proxy.cert'
```

### DIFF
--- a/docker/pypi/dmwm-base/run.sh
+++ b/docker/pypi/dmwm-base/run.sh
@@ -65,4 +65,6 @@ done
 export PYTHONPATH=$PYTHONPATH:/etc/secrets:$AUTHDIR/$fname
 
 # start the service
-nohup wmc-httpd -r -d $STATEDIR -l "|rotatelogs $LOGDIR/$srv-%Y%m%d-`hostname -s`.log 86400" $CFGFILE &
+wmc-httpd -r -d $STATEDIR -l "|rotatelogs $LOGDIR/$srv-%Y%m%d-`hostname -s`.log 86400" $CFGFILE
+# hack to keep the container running
+tail -f /etc/hosts
--- a/docker/pypi/dmwm-base/run.sh
+++ b/docker/pypi/dmwm-base/run.sh
@@ -13,14 +13,14 @@ mkdir -p $LOGDIR
 mkdir -p $STATEDIR
 mkdir -p $AUTHDIR
 mkdir -p $CONFIGDIR
-mkdir -p $AUTHDIR/proxy
 mkdir -p $AUTHDIR/../wmcore-auth
 
 # environment variables required to run some of the WMCore services
 export REQMGR_CACHE_DIR=$STATEDIR
 export WMCORE_CACHE_DIR=$STATEDIR
 
-# overwrite host PEM files in /data/srv area
+# overwrite host PEM files in /data/srv area by the robot certificate
+# Note that the proxy file is not required and used
 if [ -f /etc/robots/robotkey.pem ]; then
     sudo cp /etc/robots/robotkey.pem $AUTHDIR/dmwm-service-key.pem
     sudo cp /etc/robots/robotcert.pem $AUTHDIR/dmwm-service-cert.pem
@@ -34,29 +34,10 @@ if [ -e $AUTHDIR/dmwm-service-cert.pem ] && [ -e $AUTHDIR/dmwm-service-key.pem ]
     export X509_USER_KEY=$AUTHDIR/dmwm-service-key.pem
 fi
 
-# overwrite proxy if it is present in /etc/proxy
-if [ -f /etc/proxy/proxy ]; then
-    sudo cp /etc/proxy/proxy $AUTHDIR/proxy/proxy
-    export X509_USER_PROXY=$AUTHDIR/proxy/proxy
-fi
-
 # overwrite header-auth key file with one from secrets
-
 if [ -f /etc/hmac/hmac ]; then
-    if [ -f $AUTHDIR/../header-auth-key ]; then
-        sudo rm $AUTHDIR/../header-auth-key
-    fi
     sudo cp /etc/hmac/hmac $AUTHDIR/../wmcore-auth/header-auth-key
-    sudo chown $USER.$USER $AUTHDIR/../wmcore-auth/header-auth-key
-
-    if [ -f $AUTHDIR/header-auth-key ]; then
-        sudo rm $AUTHDIR/header-auth-key
-    fi
-    sudo cp /etc/hmac/hmac $AUTHDIR/header-auth-key
-    sudo chown $USER.$USER $AUTHDIR/header-auth-key
-    # why do we need this hmac file at /auth directory as well?
-    sudo mkdir -p /auth/wmcore-auth
-    sudo ln -s /data/srv/current/auth/$srv/header-auth-key /auth/wmcore-auth/header-auth-key
+    sudo chown 0600 $USER.$USER $AUTHDIR/../wmcore-auth/header-auth-key
 fi
 
 # use service configuration files from /etc/secrets if they are present
@@ -84,4 +65,4 @@ done
 export PYTHONPATH=$PYTHONPATH:/etc/secrets:$AUTHDIR/$fname
 
 # start the service
-wmc-httpd -r -d $STATEDIR -l "|rotatelogs $LOGDIR/$srv-%Y%m%d-`hostname -s`.log 86400" $CFGFILE
+nohup wmc-httpd -r -d $STATEDIR -l "|rotatelogs $LOGDIR/$srv-%Y%m%d-`hostname -s`.log 86400" $CFGFILE &
--- a/docker/pypi/dmwm-base/run.sh
+++ b/docker/pypi/dmwm-base/run.sh
@@ -4,28 +4,31 @@
 srv=`echo $USER | sed -e "s,_,,g"`
 STATEDIR=/data/srv/state/$srv
 LOGDIR=/data/srv/logs/$srv
+AUTHDIR=/data/srv/current/auth/$srv
+CONFIGDIR=/data/srv/current/config/$srv
 CONFIGFILE=${CONFIGFILE:-config.py}
 CFGFILE=/etc/secrets/$CONFIGFILE
 
-mkdir -p /data/srv/logs/$srv
-mkdir -p /data/srv/state/$srv
-mkdir -p /data/srv/current/auth/$srv
-mkdir -p /data/srv/current/config/$srv
-mkdir -p /data/srv/current/auth/proxy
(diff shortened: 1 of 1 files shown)


vkuznet:
@amaltaro , before you make any changes, let's clarify few things:
- can you clearly show what WMCore services needs in terms of dir structure and file placement. I'm mostly doing reverse engineering from RPM based procedure and "hidden" config/hard-coded features of WMCore about dir structure and file placement
- I rather see everything driven from env variables since it allows to place things in different places, but in WMCore everything is hard-wired, e.g. this line
```
sudo cp /etc/robots/robotkey.pem $AUTHDIR/dmwm-service-key.pem
```
will not be required if WMCore will use env variables to locate such file. But instead, it relies on specific path which makes script organization and deployment procedure very peculiar. If we'll use env variable such file(s) can be placed anywhere, e.g. in k8s we keep them in /etc/secrets, and everything can be transparent from running services
- I rather prefer to see usage of X509_USER_XXX env variable (as adapted in CMS) instead of hard-wired location/files
- please do not remove tail since run.sh should run continuously, i.e. in k8s we should start a daemon, or if you prefer not to use tail please specify how wmc-httpd should be started in continuous mode, i.e. should not exit after the start. Without that the k8s pod will be treated as cron and it will not run your service
- regarding hmac file, I do not know how WMCore services read it but I'm pretty sure it is required because WMCore services perform authorization step. So, here again instead hard-wired location we should rely on env variable and allow to keep this file anywhere in a system

The complexity of run.sh script comes from the fact that WMCore services relies on pre-defined dir structure, and it is not obvious where certain files should be placed. If we'll rely on env variable it will not be an issue and we can use one location in k8s setup and another in RPM based or user-defined deployments.

So far your changes are not obvious will work because the only way to test it is to rebuild and redeploy image to see how things will work at run-time with specific configs. I think the main issue is WMCore configuration files which define many of this hard-wired paths.

Bottom line, if you want to adapt these changes please rebuild base image and at least one service with it, then redeploy and test it. So far I can't make be sure that if we'll merge your changes things will work.

amaltaro:
@vkuznet I don't exhaustively know all the directory structure used for every WMCore service. The changes I made here bring in the most common directory structure that we have been using over the past many years, plus being conservative on some changes that you had already made here.

I am not sure whether this is used across the WMCore repository, but this module is a good reference on how the certificate file(s) are loaded in the system:
https://github.com/dmwm/WMCore/blob/master/src/python/Utils/CertTools.py

as you can see, it does rely on environment variables, but of course those don't come out of blue.

> I rather prefer to see usage of X509_USER_XXX env variable (as adapted in CMS) instead of hard-wired location/files

these are very vague comments! Keeping the secrets in `/etc/secrets` is as "hard-wired" as it is keeping it under `/data/srv/current/auth`...
Anyhow, I see no reason why we should make a copy of the robot and proxy files. IMO, we can use them from their standard location: /etc/proxy and /etc/robots/, defined in the env vars.

> please do not remove tail since run.sh should run continuously, i.e. in k8s we should start a daemon, or if you prefer not to use tail please specify how wmc-httpd should be started in continuous mode, i.e. should not exit after the start. Without that the k8s pod will be treated as cron and it will not run your service

I don't know how it works exactly, but can't we use the same mechanism used in the RPM-based images:
https://github.com/dmwm/CMSKubernetes/blob/master/docker/reqmgr2/run.sh#L74
?

> regarding hmac file, I do not know how WMCore services read it but I'm pretty sure it is required because WMCore services perform authorization step. So, here again instead hard-wired location we should rely on env variable and allow to keep this file anywhere in a system

Honestly, I don't even know if we still need it. We were using hmac to communicate with CouchDB, AFAICT. Which is not in use in the latest CouchDB version. I still think it's worth it to keep it around in case we decide to re-enable it in the near future.

> The complexity of run.sh script comes from the fact that WMCore services relies on pre-defined dir structure, and it is not obvious where certain files should be placed. If we'll rely on env variable it will not be an issue and we can use one location in k8s setup and another in RPM based or user-defined deployments.

We don't need to keep them consistent, but we need to make sure that the functionality is consistent between RPM and PyPi based deployment. Once we have a fully functional PyPi setup, there is nothing else to do with RPM but to stop using it.

And just to be explicit, this PR is providing lots of improvements and defining a small set of variables required for the data structure organization, instead of repeating the same path all over.

I understand that we do not need X509_USER_PROXY env var, and the hmac location does not matter. So, I am going to make another commit here, build the images and test them.

amaltaro:
Actually we can't use the default secrets mounted in /etc because their ownership, and permissions. So indeed we need to "massage" those a bit before an ordinary unix user can actually use them in an application.

vkuznet:
Alan, the placement of files in /etc/secrers or /data/srv areas is not a problem. The problem is WMCore service configuration files which requires specific location, e.g.
https://gitlab.cern.ch/cmsweb-k8s/services_config/-/blob/test/reqmgr2ms-output/config-output.py#L46
or
https://gitlab.cern.ch/cmsweb-k8s/services_config/-/blob/test/reqmgr2ms-output/config-output.py#L28
Because of this we must keep consistent RPM and PyPI dir structure, but it is not required if configs will not use hard-wired location and will only rely on env variables.


As  I wrote WMCore code/configs hard-wire reading from certain location, and this is the problem. It implies that we can't use /etc/secrets since configuration files insists of reading it from certain locations. If configuration or code will ready from env variables then we can easily put our secrets or any other file in any location and only properly define env variable.

Yes, you just discovered why we need to copy proxy files because of file permission which we can't change when mount secrets in k8s.

Hmac file is used by this code https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/REST/Auth.py

And, we can use https://github.com/dmwm/CMSKubernetes/blob/master/docker/reqmgr2/run.sh#L74 IF we install cron into the image. So far I don't see any reason why we should install cron tool into image because no one is using it.

vkuznet:
Alan,
let me explain again. In k8s the pod is in Running state when the process which is started is currently running. But if the process we use in k8s manifest files is finished the pod will be in Completed state. Here is an example:
```
cron-token-27849465--1-ckssp       0/1     Completed   0              100s
ms-monitor-5d484cf658-fjjcf        1/1     Running     0              125m
```
The cron, is like shell script (run.sh), when it runs it runs once and is finished with some UNIX exit code. The k8s checks this and mark pod as Completed. While, the ms-monitor in this way is continuously running process since last command of run.sh script is never ended (it is tail -f). Therefore, the pod is in Running state.

What you just changed, you modified run.sh script to run as last command `nohup ... &`. This command will be execute and start the daemon, but run.sh will be finished and exited with 0 UNIX code. So, the k8s will mark this pod as Completed, even if it spawn another unix process.In our k8s setup we have:
- each services is started via this command: https://github.com/dmwm/CMSKubernetes/blob/master/kubernetes/cmsweb/services/reqmgr2ms-output.yaml#L125
- which by itself is defined here: https://github.com/dmwm/CMSKubernetes/blob/master/kubernetes/cmsweb/services/reqmgr2ms-output.yaml#L52
- which by itself calls run.sh script
- and if run.sh script will exit the pod will be turned into completed state, but if run.sh will not exit the pod will be marked as Running.

Ideally, the service should run as daemon but not in a background, e.g. `tail -f`, or `crond -n`. In other words, we need wmc-httpd to start and do not exit. But so far I didn't find a way to start wmc-httpd this way and came up with solution to start either `tail -f` or `crond -n` instead as last command of the run.sh script.

amaltaro:
I didn't check all the WMCore services, but the initial issues reported here are gone with these changes.
@vkuznet could you please review this again? Thanks

vkuznet:
@amaltaro , even though you claimed that your deployed images run fine they do not use your changes here. To do that you need to perform few steps:
- rebuild and upload new dmwm-base image with new tag
- change all Dockerfiles in pypi to use new tag of dmwm-base image, see https://github.com/dmwm/CMSKubernetes/blob/master/docker/pypi/reqmgr2ms-monitor/Dockerfile#L1
- rebuild and upload new pypi images with new dmwm-base to registry
- redeploy k8s services with new image tag

So far your images with `pypi-20221214-alan` tag use `registry.cern.ch/cmsweb/dmwm-base:pypi-20221213` dmwm-base image, and `run.sh` which was used before this change. You can easily see that if you'll login into one of your pods, e.g.

```
kbash dmwm ms-monitor-56f5489fb-mqzc2
_reqmgr2ms@ms-monitor-56f5489fb-mqzc2:/data$ head -20 run.sh
#!/bin/bash
# script to start ReqMgr2

srv=`echo $USER | sed -e "s,_,,g"`
STATEDIR=/data/srv/state/$srv
LOGDIR=/data/srv/logs/$srv
CONFIGFILE=${CONFIGFILE:-config-monitor.py}
CFGFILE=/etc/secrets/$CONFIGFILE

mkdir -p /data/srv/logs/$srv
mkdir -p /data/srv/state/$srv
mkdir -p /data/srv/current/auth/$srv
mkdir -p /data/srv/current/config/$srv
mkdir -p /data/srv/current/auth/proxy
mkdir -p /data/srv/current/auth/wmcore-auth
....
```

So, the changes we made here did not applied to your images.

vkuznet:
The new base image with these changes is registry.cern.ch/cmsweb/dmwm-base:pypi-20221215, now I'm rebuilding all pypi images with it.

amaltaro:
You are right, I rebuilt the dmwm-base image, but forgot to update its label in the WMCore dockerfiles, sorry.

I have tests running on cmsweb-test9, which includes a WMAgent processing those workflows. But provided that we do not redeploy CouchDB, we should be fine upgrading the WMCore services.

vkuznet:
The new pypi images with these changes now exists in registry with tag pypi-20221215-base, feel free to redeploy them in test9 when it convenient for you, and then verify if the issue is really gone with these changes. So far I committed changes to use new base image tag.

PR #1289, opened 2022-12-14
Helm chart for speedbump.

arooshap:
Hi @vkuznet,

I have created a helm chart for the speedbump image. It allows the use of speedbump as a container. We can edit it at runtime, for example, `helm install speedbump speedbump -n http --set args.sine-amplitude=200ms --set args.sine-period=2m --set args.host="localhost" --set args.port=8888`. 

When it comes to customizing the flags for the command, this array ensures that it is well-formatted:          

```
args:
          {{- range $key, $value := .Values.args }}
            {{- if eq ($key| toString) "localhost" }}
            - {{ $key }}:{{ $value }}            
            {{- else if eq ($key| toString) "version"  }} 
            - --{{ $key }}
            {{- else if or (eq ($value| toString) "0") ( not $value) }}
            {{- else}}
            - --{{ $key }}={{ $value }}
            {{- end }}
          {{- end }}
```

Can you please review this?


vkuznet (approved):
@arooshap the changes look fine to me

PR #1292, opened 2022-12-17
Fix fts log-clustering, implement cms-dm-monit deploy scripts, improve sqoop and hdfs exporter

mrceyhun:
- **cms-dm-monit**(for rucio datasets monitoring) cluster create and deploy scripts are created: `create_cluster_dmmon.sh` and `deploy-dmmon.sh`. [Documentation](https://cmsmonit-docs.web.cern.ch/k8s/cluster_upgrades/#dm-mon) provides more details about current ingress bugs and solutions which took my hours last week.
- **hdfs-exporter-eos** is improved and image tag is updated
- **Sqoop** deployment is modified to more implicit way using CM. It is convenient, and no hardcoding like pushateway url.
- **log-clustering** is fixed
- Sqoop and log-clustering moved to CRON cluster

@vkuznet This repository https://github.com/vkuznet/log-clustering size is so big because of its git history, more than 300MB which is not good for docker image. I truncated old unnecessary big files/directories from git history in my fork https://github.com/mrceyhun/log-clustering but I cannot open PR to your repo, it can be because of i modified old  commits. May be you or owner of the repo (nikodemos) can wipe these unnecessary files/folders from your git history than i can open PR.

I found big unnecessary files using[1] and reduced the size to 20MB. Final commands that use:
```
git filter-branch -f --index-filter "git rm -rf --cached --ignore-unmatch word2vec.model error1109.txt word2vec.model.wv.vectors.npy word2vec.model.trainables.syn1neg.npy workflow/log-clustering/lib data ascii_patterns.txt test.ipynb errorSHORT* workflow/log-clustering/share clusterlogs/__pycache__ " -- --all

rm -rf .git/refs/original/
git reflog expire --expire=now --all
git gc --prune=now
git gc --aggressive --prune=now

git push --all --force
```

[1] https://stackoverflow.com/a/10622293/6123088

fyi @leggerf @brij01


vkuznet:
@mrceyhun , I made clean-up of the repo following your steps. Should we merge the PR?

Meanwhile, it is unclear why do you need git for image if it is too big. You can simply fetch tarball from github and build your image, isn't it?

mrceyhun:
Thanks @vkuznet . My changes are ready in the PR now : https://github.com/vkuznet/log-clustering/pull/27 . Dockerfile will continue to use your repo, so this PR is also ready to merge. 

There is one issue that I don't understand. log-clustering producer has been running since a few days: https://monit-kibana.cern.ch/kibana/goto/3622a7c1e642f2ba18d78738320fddef
There is an incremental pattern which I suspect from hourly running and this [code block](https://github.com/vkuznet/log-clustering/blob/master/workflow/workflow.py#L102-L105). When we run cron job hourly, it will query from the beginning of the day(YYYY-MM-DD represents 00:00 hour) and it will continue like that. Is this expected or should we make arrangement to fix it: run daily, query daily (like now) or run hourly, query previous hour?

vkuznet:
Regarding pattern observation. Everything depends on use-case where we apply log-clustering. I think the decision about query should be driven by use-case(s). I suggest to evaluate who is using this info and adjust accordingly.

mrceyhun:
I am not sure if anyone is using fts log clustering . @leggerf @brij01 can we make firstly re-advertisement of this service? May be it is useful for some people but they are not aware of it. If nobody is interested, we can decommission it. It would be useful if there is already presentation about it, probably from years 2019/2020.

mrceyhun:
New Kibana link after OpenSearch switch: https://monit-opensearch.cern.ch/dashboards/goto/e4b39375b2f2fa7afa0f3491a3192360?security_tenant=global

PR #1295, opened 2023-01-05
Fix permissions for /data/srv in the PyPi based images

amaltaro:
After deploying WMCore images `2.1.6rc4` into testbed, many of the services are not functional. Scanning the logs, I see this for reqmgr2:
```
mkdir: cannot create directory ‘/data/srv/state’: Permission denied
mkdir: cannot create directory ‘/data/srv/current’: Permission denied
mkdir: cannot create directory ‘/data/srv/current’: Permission denied
mkdir: cannot create directory ‘/data/srv/current’: Permission denied
```

It's not clear to me why it works on test9 but not in testbed, but this might resolve the issue.

Please do NOT merge it because the other WMCore services will need the same changes, but I still have to test it.

### DIFF
--- a/docker/pypi/dmwm-base/run.sh
+++ b/docker/pypi/dmwm-base/run.sh
@@ -9,6 +9,9 @@ CONFIGDIR=/data/srv/current/config/$srv
 CONFIGFILE=${CONFIGFILE:-config.py}
 CFGFILE=/etc/secrets/$CONFIGFILE
 
+### permission update to workaround issues with mounting logs volume
+sudo chown -R $USER.$USER /data
+
 mkdir -p $LOGDIR
 mkdir -p $STATEDIR
 mkdir -p $AUTHDIR
--- a/docker/pypi/global-workqueue/Dockerfile
+++ b/docker/pypi/global-workqueue/Dockerfile
@@ -1,4 +1,4 @@
-FROM registry.cern.ch/cmsweb/dmwm-base:pypi-20221219
+FROM registry.cern.ch/cmsweb/dmwm-base:pypi-20230106
 MAINTAINER Valentin Kuznetsov vkuznet@gmail.com
 ENV TAG=2.1.4
 RUN pip install global-workqueue==$TAG
--- a/docker/pypi/reqmgr2/Dockerfile
+++ b/docker/pypi/reqmgr2/Dockerfile
@@ -1,4 +1,4 @@
-FROM registry.cern.ch/cmsweb/dmwm-base:pypi-20221219
+FROM registry.cern.ch/cmsweb/dmwm-base:pypi-20230106
 MAINTAINER Valentin Kuznetsov vkuznet@gmail.com
 ENV TAG=2.1.4
 RUN pip install reqmgr2==$TAG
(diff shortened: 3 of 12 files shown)


vkuznet:
I think the actual issue is that we mount external disk for /data/srv/logs area, e.g. https://github.com/dmwm/CMSKubernetes/blob/master/kubernetes/cmsweb/services/reqmgr2.yaml#L150 on testbed/production clusters. As such I doubt any changes needed to the images. You can see it by pulling docker image and log into it, e.g.
```
docker pull registry.cern.ch/cmsweb/reqmgr2:pypi-20221219-base
sudo docker run -ti registry.cern.ch/cmsweb/reqmgr2:pypi-20221219-base /bin/bash
_reqmgr2@c9737a90bd02:/data$ ls -al /data/
total 8
drwxr-xr-x. 1 _reqmgr2 _reqmgr2   34 Dec 19 19:30 .
drwxr-xr-x. 1 root     root        6 Jan  5 14:40 ..
-rwxr-xr-x. 1 _reqmgr2 _reqmgr2 1956 Dec 13 17:01 manage
-rwxr-xr-x. 1 _reqmgr2 _reqmgr2 2567 Dec 19 19:29 run.sh
```
The /data area is owned by `_reqmgr2` and it works in dev clusters, but since we mount external disk to logs it overwrites permissions (I think). The proper fix should be done in k8s manifest files where we **ONLY** need to mount `/data/srv/logs` area via emptyDir directive. You may ask @muhammadimranfarooqi  to investigate and fix corresponding manifest entries (for all WMCore images). Or, we should change location of logs and use different mount point (the changes will be required in manifest and WMCore configuration files then).

vkuznet:
You may try it and it may work, but I doubt it is correct solution since we are adapting RPM based deployment to k8s and pypi installation which does not require it.

amaltaro:
I read a bit about the emptyDir directive and it looks like it removes this directory and the content whenever the POD is removed/deleted. Unless I am missing something, I fail to see why it would be the way to address this issue. Can you please elaborate on this?

amaltaro:
Before proceeding with the `chmod` suggestion I made above, I will wait to hear if Imran and/or Aroosha have any advice here. For now, I have reverted WMCore testbed services to HG2212d.

amaltaro:
@arooshap Aroosha, if I understood you right, you are suggesting to first try making changes to the run.sh script and execute the `chmod` command on the `/data` directory, right? If so, I will update this PR and build a couple of images to try this out.

amaltaro:
After changing the run.sh script, rebuilding the dmwm-base image and reqmgr2 (tagged as `pypi-20230106-alan2`), I managed to successfully deploy and run this service in testbed.

Here is a snippet of a curl call to reqmgr2, where we can confirm that it's now running the latest RC release:
```
curl --cert $X509_USER_CERT --key $X509_USER_KEY "https://cmsweb-testbed.cern.ch/reqmgr2/data/info"
{"result": [
 {
  "wmcore_reqmgr_version": "2.1.6rc4",
  "reqmgr_db_info": {
...
```

I am going to update this PR (rolling back my changes to the Dockerfiles) and it then it will be ready to merge.

amaltaro:
I have built and updated a new `dmwm-base:pypi-20230106` image and also updated all the WMCore Pypi-based dockerfiles to use this new tag.
@arooshap or @vkuznet please review, and if everything is fine, merge it such that I can deploy fresh new images in testbed.

PR #1301, opened 2023-01-25
Add monitoring script for all pypi images

vkuznet:


vkuznet:
@amaltaro , the new image is available with this tag `registry.cern.ch/cmsweb/dmwm-base:pypi-20230125`. Therefore, if we'll agree to merge this PR we'll need to create another one to use new `dmwm-base` image tag in all pypi images. Said that, there are different monitor exporters we provide:
- `process_exporter` provides basic UNIX metrics, like CPU, RAM, number of open/closed connections, file descriptors
- `cpy_exporter` is a new (experimenal) exporter which can scrape metrics from CherryPy server. I did not enable it (as it will require changes on k8s manifest and/or monitoring sides), but I added this exporter to system area that we can experiment with it and decide later if it will be useful for WM monitoring needs. This exporter provides comprehensive CherryPy metrics, including individual thread-based metrics, and therefore can be very valuable for debugging individual threads issues.

docker/pypi/dmwm-base/monitor.sh
```diff
@@ -0,0 +1,19 @@
...
+
+# start process exporter
+configs="config config-monitor config-output config-transferor config-ruleCleaner config-unmerged"
+for p in $configs; do
+    pat="wmc-httpd.*$p"
+    pid=`ps axjfwww | grep "$pat" | grep -v grep | grep -v process_monitor | grep -v " 1 " | awk '{print $1}'`
+    if [ -n "$pid" ]; then
+        app=`grep ^main.application /etc/secrets/${p}.py | sed -e 's,#.*,,g' | awk '{split($0,a,"="); print a[2]}' | sed -e "s, ,,g" -e 's,",,g'`
+        prefix=process_exporter_${app}
+        port=`grep main.port /etc/secrets/config.py | sed -e 's,#.*,,g' | awk '{split($0,a,"="); print a[2]}' | sed -e "s, ,,g"`
```
amaltaro, line 11: Maybe you wanted to write `/etc/secrets/${p}.py`?

amaltaro (changes requested):
I think there is one required change to be performed. Other than that, it all looks good and I agree that we should keep `cpy_exporter` disabled for the moment. Once we are in a stable ground with all of this, we can then explore the CherryPy based monitoring.

docker/pypi/dmwm-base/monitor.sh
vkuznet: good catch, I pushed the changes.

vkuznet: and, I rebuilt and uploaded again dmwm-base image to registry with the same tag

amaltaro (approved):
Looks good to me. Can you please also update all of the WMCore dockerfiles with the new dmwm-base image tag?

vkuznet:
ok, I'll merge this PR, and will create another one for pypi images using new dmwm-base image tag.

PR #1303, opened 2023-01-26
Changes to mongodb Chart Name.

arooshap:


arooshap:
@muhammadimranfarooqi The issue was with the Chart Name. I am guessing that the build script packages the charts with `--generate-name` flag, hence the discrepancy that we noticed earlier.

PR #1317, opened 2023-02-15
Update frontendx509-ds.yaml

arooshap:


arooshap:
Hi, I am merging this as this fixes the liveness probe in the frontendx509 image.

PR #1318, opened 2023-02-16
Script for renewing proxies on testbed and production VMs.

arooshap:
Sometimes, the proxies in the VMs expire, and they urgently need to be renewed. This script helps in urgently renewing the proxies in all the testbed and production VMs.

### DIFF
--- a/kubernetes/cmsweb/daemonset/frontendx509-ds.yaml
+++ b/kubernetes/cmsweb/daemonset/frontendx509-ds.yaml
@@ -77,7 +77,7 @@ spec:
               apiVersion: v1
               fieldPath: metadata.name
         ports:
-        - containerPort: 80
+        - containerPort: 8080
           name: http
         - containerPort: 8443
           name: services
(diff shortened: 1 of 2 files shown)


muhammadimranfarooqi:
@arooshap 

This `CMSKubernetes` repository is for kubernetes cluster. We already have a script for proxy renewal in deployment repository here https://github.com/dmwm/deployment/blob/master/admin/ProxyRenew. This is deployed as cronjob in all VMs.

arooshap:
@muhammadimranfarooqi Thanks for the review, and yes, I have an idea of the script that you've mentioned and that's why I also refer to it in this script. The sole purpose of this script is to deploy the proxy in all the production and testbed VMs all at once. I had this script with me for a long time, and I presumed that it will also facilitate someone else in the future, so I decided to put it here.

muhammadimranfarooqi:
@arooshap 

Then I suggest putting it in the deployment repository in `admin` directory. Add some more comments and documentation in the script about its usage. Cheers

PR #1319, opened 2023-02-27
SOPs Dockerfile for CI/CD pipeline.

arooshap:


vkuznet (commented):
@arooshap , sops is Go based static executable, therefore you do not need any shared libraries. Moreover, it is security tool, and you may consider to avoid using ANY linux distribution and package it with distroless image `FROM gcr.io/distroless/static` which does not have any Linux OS and therefore free from vulnerabilities. If you need certs you may copy them from alpine to distroless since they are files. Bottom line, in those cases when you use Go executable and its executable is statically linked, e.g.
```
ldd sops-v3.7.3.linux
ldd: warning: you do not have execution permission for `./sops-v3.7.3.linux'
        not a dynamic executable
```
I strongly advise to use distroless image to avoid any security vulnerabilities and prohibit any shell access to the image.

vkuznet (approved):

PR #1325, opened 2023-03-08
Update dependencies in dockers and refactor, monitoring

mrceyhun:
Bump golang and alpine image versions

Images will be build after https://github.com/dmwm/CMSMonitoring/pull/203


vkuznet (commented):
@mrceyhun , if possible I suggest to re-evaluate usage of alpine image as final one. It is still full Linux OS which may contain vulnerabilities. The better alternative is to use distroless image, see # https://blog.baeke.info/2021/03/28/distroless-or-scratch-for-go-apps/, and it goes like this `FROM gcr.io/distroless/static AS final`. It can run go static executable but does not provide LInux OS shell, etc. In other words it does not have  vulnerabilities by definition. So, if you do not really need a shell to your pod, and your go executable is statically linked then you may switch from alpine to distroless. For example see https://github.com/dmwm/CMSKubernetes/blob/master/docker/cmsamqproxy/Dockerfile

mrceyhun:
Thanks @vkuznet , I'll go over them. Let's keep this PR open till then.

mrceyhun:
I dropped the Loki change and left its `gcr.io/distroless/static` change to CMSWEB team, because I need to merge this PR.

arooshap:
@mrceyhun thanks for the tag. I will take care of `Loki` change.

PR #1328, opened 2023-03-11
change order of wmc-http and monitor script

vkuznet:


amaltaro:
Haa, this explains why I could run this process in a container that was already running, but none started from the container creation.

Change looks good to me. We still need to build a new dmwm-base image and update all the other PyPi dockerfiles.

PR #1331, opened 2023-03-13
Replace dash with underscore for prefix name

vkuznet:


amaltaro:
@vkuznet did you manage to test it in one of the microservices pod?

vkuznet:
Alan, I did test it on ms-monitor pod, and after testing I added additional protection to start process_exporter. I would appreciate if you'll review changes again (I only added additional if/fi to check for file existence).

amaltaro:
@vkuznet I might have an implementation that will look more clear, which would be something like:
```
for p in $configs; do
    if [ -f /etc/secrets/${p}.py ]
        pat="wmc-httpd.*$p"
        pid=`ps axjfwww | grep "$pat" | grep -v grep | grep -v process_monitor | grep -v " 1 " | awk '{print $1}'`
        if [ -n "$pid" ]; then
```
what do you think?

docker/pypi/dmwm-base/monitor.sh
```diff
@@ -3,17 +3,21 @@
...
-        app=`grep ^main.application /etc/secrets/${p}.py | sed -e 's,#.*,,g' | awk '{split($0,a,"="); print a[2]}' | sed -e "s, ,,g" -e 's,",,g'`
-        prefix=process_exporter_${app}
-        port=`grep main.port /etc/secrets/${p}.py | sed -e 's,#.*,,g' | awk '{split($0,a,"="); print a[2]}' | sed -e "s, ,,g"`
-        address=":1${port}"
-        echo "Start ${prefix} on ${address}"
-        nohup process_exporter -pid $pid -prefix $prefix -address "$address" 2>&1 1>& ${prefix}.log < /dev/null &
-        #cpyAddr=`echo ${address} | sed "s,8,9,g"`
-        #echo "Start cpy_exporter on ${cpyAddr}"
-        #nohup cpy_exporter -address "$address" 2>&1 1>& cpy_${prefix}.log < /dev/null &
+    if [ -f /etc/secrets/${p}.py ]; then
```
amaltaro, line 6: Should we add an "echo" right after this if statement to record that it found configuration file XYZ?

docker/pypi/dmwm-base/monitor.sh
```diff
@@ -3,17 +3,21 @@
...
+        pat="wmc-httpd.*$p"
+        pid=`ps axjfwww | grep "$pat" | grep -v grep | grep -v process_monitor | grep -v " 1 " | awk '{print $1}'`
+        if [ -n "$pid" ]; then
```
amaltaro, line 9: I would add an echo here as well saying that a PID was found.

amaltaro (approved):
@vkuznet Valentin, these changes are good to me. However, I think there would be value in adding a couple more log records in there (as suggested inline).

Once you test these changes, I think it's good to be merged.

vkuznet:
sure, I added both echo's , please have a look last time and let me know if I should merge and proceed with rebuild of based image.

amaltaro (approved):
Thanks for the prompt fixes. It looks good to me.

PR #1336, opened 2023-03-22
Changes for mongodb to use different hostnames.

arooshap:


arooshap:
Hi @todor-ivanov,

I have updated the configuration for mongodb in both the helm chart and docker image to make sure that we are configuring mongodb on three different hostnames.

This parameters for hostNames can be passed on runtime, thanks to Helm. Since they are passed as environment variables, I think this is the most straightforward way. But as you mentioned, we can also pass them through a custom configuration through a configmap or secret.

The pod to node mapping is depicted in these lines: [pod-0](https://github.com/dmwm/CMSKubernetes/blob/f0f51dc5dbf9fec881bea47960437bdc8980259c/helm/mongodb/templates/deployment.yaml#L51), [pod-1](https://github.com/dmwm/CMSKubernetes/blob/f0f51dc5dbf9fec881bea47960437bdc8980259c/helm/mongodb/templates/deployment.yaml#L141), [pod-2](https://github.com/dmwm/CMSKubernetes/blob/f0f51dc5dbf9fec881bea47960437bdc8980259c/helm/mongodb/templates/deployment.yaml#L221).

As such, we do not rely on any external/manual operations to get the mongodb service to run. As you can see [here](https://github.com/dmwm/CMSKubernetes/blob/f0f51dc5dbf9fec881bea47960437bdc8980259c/docker/mongodb/source/dockerfile#L10), using the `CMD` functionality, mongodb along with the replicasets is configured as soon as the pods start.

Let me know what further changes you require in terms of configuring the `mongodb` service.

docker/mongodb/source/startup-script-mongo/initialize-mongo-rs.sh
```diff
@@ -2,7 +2,7 @@
 
 echo "Executing initialize-mongo-rs.sh"
 
-mongo --eval "mongodb = ['$NODE_HOSTNAME:32001', '$NODE_HOSTNAME:32002', '$NODE_HOSTNAME:32003'], rsname = '$RS_NAME'" --shell << EOL
+mongo --eval "mongodb = ['$NODE_HOSTNAME_ONE:32001', '$NODE_HOSTNAME_TWO:32002', '$NODE_HOSTNAME_THREE:32003'], rsname = '$RS_NAME'" --shell << EOL
```
amaltaro, line 5: Aroosha, instead of reading these values from the command line, could we define a configuration file and simply load it when initializing MongoDB?
If I understand this correctly, this is exactly what we do between lines 6 and 15. But having a configuration file is likely more robust and less error prone to typos and mistakes.

amaltaro: I don't know how flexible you need these host names to be. But the way I was thinking is along the lines of what we have for services_config, so you would have separate configuration files (e.g. preprod, prod, dev, etc) and load the correct configuration during mongo initialization.

The configuration file would have the node names and their ports. Similar to what we have when pointing to CMSWEB services. I could be missing specific details here as well, in that case, just disconsider my comment!

arooshap: @amaltaro that's also a way to do it, but specific ports and node names can be also configured through values.yaml file. So, if we have mongodb in production, we can configure it to have different hostnames/ports/configuration compared to the preproduction through a single helm chart. I personally think that having a separate configuration file when we can configure everything through a single Helm chart adds an additional layer to the setup! (that is one of the motivations for moving to Helm charts rather than relying on simple manifest files).

But, of course, the decision lies with you and @todor-ivanov. Let me know your thoughts!

vkuznet:
I suggest sticking with the standard helm chart approach, i.e. use values yaml file which is the equivalent of configuration one for helm chart itself. Alan, Todor please spend some time to learn about it and you will understand the structure. In simple words the helm manifest file is a template while the values.yaml file is configuration file which you pass to helm. The helm uses a template and fills out its placeholders using the values.yaml file.

Sent from Proton Mail mobile

-------- Original Message --------

todor-ivanov:
Hi @vkuznet we do understand the concept of helm chart vs. templates vs. values ... that was not the request @amaltaro was making. We can still use those  to regenerate the config files obtaining the proper variables (as hostnames/pvNames etc.) from the deployment at hand. What we were talking about is... Instead of re-configuring the replicaset every time the primary node is restarted through some bash redirects as done here:  
https://github.com/dmwm/CMSKubernetes/blob/4ec79417db9ba749409eabcecff91d4d710761d8/docker/mongodb/source/startup-script-mongo/initialize-mongo-rs.sh#L6-L14

We can use the database internally provided mechanisms for preserving those configuration inside the pod at  the proper place/file:  i.e. `/etc/mongo.conf` as explained here: https://www.mongodb.com/docs/manual/reference/configuration-options/  and ehere https://www.mongodb.com/docs/manual/reference/configuration-options/#mongodb-setting-replication.replSetName 

The so done redirects in this bash script are actually generating one config file on the fly and passing it to the database with `mongo shell` which results in exactly the same thing that would have hapened if preserved inside a file and left visible for later debugging when logged at the pod. 

In addition, this way we would avoid reconfiguration at very  primary node restarts, and can let the configuration which is preserved inside the database itself take over when needed. Currently on every restart this reconfiguration at runtime is called from here: https://github.com/dmwm/CMSKubernetes/blob/4ec79417db9ba749409eabcecff91d4d710761d8/docker/mongodb/source/startup-script-mongo/startup-mongo-0.sh#L5 

Well this reconfigure script does work, even though it is counter intuitive, because this line here 
https://github.com/dmwm/CMSKubernetes/blob/4ec79417db9ba749409eabcecff91d4d710761d8/docker/mongodb/source/startup-script-mongo/reconfig-mongo-rs.sh#L20

would always evaluate to `True` for every replicamember in the replicaset. But anywa we are safe since it is always called only from the node we point as primary as primary ... all good. I tested it and it works. The replicaset recovers/recreates itslef upon various modes of failure even for the pod serving as primary. It is just really difficult to follow what is going on during these server restarts.

todor-ivanov:
@arooshap  one final question left here:
> so you would have separate configuration files (e.g. preprod, prod, dev, etc) and load the correct configuration during mongo initialization

How do we currently differentiate the setup helm values (mostly DNS aliases and server names) for the different  deployment scenarios `mongo-prod` vs `mongo-preprod`?

arooshap:
@todor-ivanov currently, we set the hostnames at runtime(with the helm install command). The rest of the setup is similar for both clusters. We can also customize helm to have urls based on the environment(prod, preprod, e.t.c.), thereby, eliminating the need for specifying URLs at runtime because that can be error-prone and tedious.

todor-ivanov:
Hi @arooshap 

Thanks for the quick reply!

> We can also customize helm to have urls based on the environment(prod, preprod, e.t.c.), thereby, eliminating the need for specifying URLs at runtime because that can be error-prone and tedious

And this was  exactly @amaltaro 's second point. So if we  can do that in a properly visible way so that  it becomes immediately obvious what are the host aliases for the environment one currently deploys and especially if we can preserve those somewhere in git/gitlab that would be great. 

Regarding saving the configuration files inside the pod ... Well we can always make things better and better...  but I am willing to merge this PR now  (since it is already tested and works) and move on.  This refinement we can make it later.

todor-ivanov:
@arooshap you may merge this PR and we move on. I believe @amaltaro agrees as well.

PR #1348, opened 2023-04-07
Fix badly set sops executable.

todor-ivanov:
In the current code the `sops` executable is badly set and always assumed to be present at the user's $HOME.  because the results from the test for the presence of the `sops` command here: https://github.com/dmwm/CMSKubernetes/blob/ee4dd47e0b1a39cc6cabba4acaf0c952e15df28b/kubernetes/cmsweb/scripts/deploy-secrets.sh#L21-L28 are simply ignored for the rest of the script. Currently the test at the lxplus cluster gives positive result:
```
[<user>@lxplus804 cmsweb]$ echo "$(command -v sops)"
/usr/bin/sops
```
And the above manual deployment of the `sops` executable at  the user's home destination simply never happens.
 
This misbehavior may have been masked because in the past the machines at the `lxplus 8` cluster have been missing the the executable at `/usr/bin/sops` and the above test has been always returning negative results, hence  the people who were using the script were always ending up having the package installed at their  $HOME. This is not the case in lxplus machines any more... so this script cannot work in the current form for every body.   

With the current PR the executable to be used down the script is tied to the  result of  the test for the presence of the command. And the artificial installation at the user's $HOME destination happens only if it is missing at the machine from which the current deployment is happening.

### DIFF
--- a/kubernetes/cmsweb/scripts/deploy-secrets.sh
+++ b/kubernetes/cmsweb/scripts/deploy-secrets.sh
@@ -35,7 +35,7 @@ SOPS=$(command -v sops) || installSops ||  { err=$?; echo "`sops` command is not
 [[ -e $SOPS ]] || installSops || { err=$?; echo "$SOPS is not executable"; exit $err; }
 
 # check if sops is the expected version:
-[[ $($SOPS --version) =~ sops[[:blank:]]+3\.7\.[2,3]+.* ]] || installSops || { err=$?; echo "$SOPS is not the expected version"; exit $err; }
+[[ $($SOPS --version) =~ sops[[:blank:]]+3\.[7-9]+\.[0-9]+.* ]] || installSops || { err=$?; echo "$SOPS is not the expected version"; exit $err; }
 
 # cmsweb configuration area
 echo "+++ cluster name: $cluster_name"
--- a/kubernetes/cmsweb/scripts/deploy-secrets.sh
+++ b/kubernetes/cmsweb/scripts/deploy-secrets.sh
@@ -18,14 +18,15 @@ fi
 mkdir -p $tmpDir
 cd $tmpDir
 
-if [ -z "$(command -v sops)" ]; then
+SOPS=$(command -v sops)  || {
     # download soap in tmp area
     wget -O sops https://github.com/mozilla/sops/releases/download/v3.7.2/sops-v3.7.2.linux.amd64
     chmod u+x sops
     mkdir -p $HOME/bin
     echo "Download and install sops under $HOME/bin"
     cp ./sops $HOME/bin
-fi
+    SOPS="$HOME/bin/sops"
+}
 
 # cmsweb configuration area
 echo "+++ cluster name: $cluster_name"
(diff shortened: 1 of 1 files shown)


todor-ivanov:
FYI: @arooshap  @vkuznet @amaltaro  

p.s. I have no idea why I cannot request for review of this PR but anyway .. I ping you here in the comment.

kubernetes/cmsweb/scripts/deploy-secrets.sh
```diff
@@ -18,14 +18,15 @@ fi
...
+SOPS=$(command -v sops)  || {
     # download soap in tmp area
     wget -O sops https://github.com/mozilla/sops/releases/download/v3.7.2/sops-v3.7.2.linux.amd64
     chmod u+x sops
     mkdir -p $HOME/bin
     echo "Download and install sops under $HOME/bin"
     cp ./sops $HOME/bin
-fi
+    SOPS="$HOME/bin/sops"
+}
```
vkuznet, line 29: I suggest to add additional test here something like this:
```
if [ ! -f $SOPS ]; then
   echo "$SOPS does not exist"
   exit 1
fi
```
The point is that we should verify existence of the file. Moreover, I suggest to run  to run `$SOPS --version` and apply some regexp pattern matching to verify that it is indeed sops executable with version which we expect.

vkuznet (changes requested):

kubernetes/cmsweb/scripts/deploy-secrets.sh
todor-ivanov: which is the expected  version ?

todor-ivanov:
hi @vkuznet 
> sops executable with version which we expect

And there we have it with my latest commit. Please take another look

vkuznet (approved):

kubernetes/cmsweb/scripts/deploy-secrets.sh
```diff
@@ -29,13 +29,13 @@ installSops() {
...
-SOPS=$(command -v sops) || installSops ||  { err=$?; echo "`sops` command not setup"; exit $err; }
+SOPS=$(command -v sops) || installSops ||  { err=$?; echo "`sops` command is not properly setup"; exit $err; }
 
 # check if sops is executable:
-[[ -e $SOPS ]] || installSops || { err=$?; echo "`sops` command not executable"; exit $err; }
+[[ -e $SOPS ]] || installSops || { err=$?; echo "$SOPS is not executable"; exit $err; }
 
 # check if sops is the expected version:
-[[ $($SOPS --version) =~ sops[[:blank:]]+3\.7\.[2,3]+.* ]] || installSops || { err=$?; echo "`sops` command not expected version"; exit $err; }
+[[ $($SOPS --version) =~ sops[[:blank:]]+3\.7\.[2,3]+.* ]] || installSops || { err=$?; echo "$SOPS is not the expected version"; exit $err; }
```
vkuznet, line 38: since you're checking and printing in error actual version, I suggest to check against `3.[7-9]` pattern instead of `3.7,x` one.

vkuznet (changes requested):

kubernetes/cmsweb/scripts/deploy-secrets.sh
todor-ivanov: ok

todor-ivanov:
Hi @vkuznet ,  please take another look

vkuznet (approved):

muhammadimranfarooqi:
Thanks @todor-ivanov  for this PR.

PR #1349, opened 2023-04-11
Add crabserver process exporter to cmsweb prometheus

mrceyhun:
CRAB team deployed new process exporter to monitor crabserver processes. To be able to remote write metrics to CMS Monitoring, we need to scrape them in CMSWEB prometheus. Its job config is added.

@arooshap could you please deploy this to testbed?

@vkuznet we can scrape them directly from Monitoring prometheus like we do for [cherrypy](https://gitlab.cern.ch/cmsmonitoring/cmsmon-configs/-/blob/master/prometheus/cherrypy.json) . I don't want to go against the convention, but I am open to suggestions. What do you think?

### DIFF
--- a/kubernetes/cmsweb/monitoring/prometheus-services.yaml
+++ b/kubernetes/cmsweb/monitoring/prometheus-services.yaml
@@ -51,6 +51,10 @@ data:
       - job_name: 'crabserver-service'
         static_configs:
             - targets: ['crabserver.crab.svc.cluster.local:18270']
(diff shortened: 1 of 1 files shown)


vkuznet (approved):

arooshap:
Hi @mrceyhun @mapellidario, I deployed this on testbed and it seems to scrape the crabserver metrics just fine. 

Please validate the results, and then I will deploy it to the production as well.

PR #1350, opened 2023-04-12
Set CouchDB image to use 3.2.2 version; fix monitoring.ini in the CouchDB image

amaltaro:
As the title says:
* ensure we are using the same CouchDB version as used in preprod/prod (3.2.2)
* stop using monitoring.ini for vanilla couchdb
* fix credentials used for couchdb_prometheus exporter

UPDATE: exporter setup similar to what we have been using in preprod/prod:
https://github.com/dmwm/deployment/blob/master/exporters/manage#L167

### DIFF
--- a/docker/couchdb/monitor.sh
+++ b/docker/couchdb/monitor.sh
@@ -1,7 +1,9 @@
 #!/bin/bash
 
 # start couchdb exporter
-nohup couchdb-prometheus-exporter -telemetry.address=":9984" 2>&1 1>& couchdb_exporter.log < /dev/null &
+COUCH_CONFIG=/data/srv/current/auth/couchdb/couchdb_config.ini
+nohup couchdb-prometheus-exporter -telemetry.address=":9984" -logtostderr=true \
+      --config=$COUCH_CONFIG -databases.views=false 2>&1 1>& couchdb_exporter.log < /dev/null &
 
 # run filebeat
 if [ -f /etc/secrets/filebeat.yaml ] && [ -f /usr/bin/filebeat ]; then
--- a/docker/couchdb/monitor.sh
+++ b/docker/couchdb/monitor.sh
@@ -2,8 +2,15 @@
 
 # start couchdb exporter
 COUCH_CONFIG=/data/srv/current/auth/couchdb/couchdb_config.ini
-nohup couchdb-prometheus-exporter -telemetry.address=":9984" -logtostderr=true \
+# test if file is not zero size
+if [ -s "${COUCH_CONFIG}" ]; then
+    sudo cp /etc/secrets/$fname /data/srv/current/auth/$srv/$fname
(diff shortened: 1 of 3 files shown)


amaltaro:
I just built CouchDB `3.2.2-v5` and I can confirm that logs are clean of authentication error and we also have a functional prometheus exporter. New image is available at the usual location:
https://registry.cern.ch/harbor/projects/1771/repositories/couchdb/artifacts-tab

@vkuznet can you please review it? Thanks

vkuznet (changes requested):

docker/couchdb/monitor.sh
```diff
@@ -1,7 +1,9 @@
 #!/bin/bash
 
 # start couchdb exporter
-nohup couchdb-prometheus-exporter -telemetry.address=":9984" 2>&1 1>& couchdb_exporter.log < /dev/null &
+COUCH_CONFIG=/data/srv/current/auth/couchdb/couchdb_config.ini
+nohup couchdb-prometheus-exporter -telemetry.address=":9984" -logtostderr=true \
```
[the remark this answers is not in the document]
vkuznet: Well, in my view you should print appropriate message but there is no need to start exporter which will fail every cycle it will query couchdb. Doing this way we'll have appropriate message in a log and no exporter.

amaltaro:
@vkuznet the 3rd commit should address your comment. Please have another look.

docker/couchdb/monitor.sh
```diff
@@ -1,7 +1,16 @@
 #!/bin/bash
 
 # start couchdb exporter
-nohup couchdb-prometheus-exporter -telemetry.address=":9984" 2>&1 1>& couchdb_exporter.log < /dev/null &
+COUCH_CONFIG=/data/srv/current/auth/couchdb/couchdb_config.ini
+# test if file is not zero size
+if [ -s "${COUCH_CONFIG}" ]; then
```
vkuznet, line 6: I do not know if you'll find it useful, but I suggest to further separate this if block into two if blocks. First, should check if $COUCH_CONFIG file exist and if not make a copy from secrets. The second should check if file is empty or not. This separation should avoid copying secrets over and over if we'll re-turn monitor.sh again and again (when you restart within a pod). But this suggestion is not necessary per-se and I'll approve the PR.

vkuznet (approved):

docker/couchdb/monitor.sh
amaltaro: This would add code duplication, given that the file is already copied in the Dockerfile.

amaltaro:
Tag `3.2.2-v5` has been built with all the changes in this PR. If you think these are good to go, then I'd suggest getting this merged. Please let me know if you want me to squash the commits @vkuznet

vkuznet:
As I said, I'm ok with the merge, and I'm not big fan of squashing, therefore using my admin rights on this repo, I'm merging.

PR #1362, opened 2023-04-21
Fix WMAgent rpm dcoker image

todor-ivanov:
Do NOT merge Yet. 

The current PR is intended only to fix the old RPM based Docker image. so that we can test it and reuse bits of it in the new Pypi based setup.

### DIFF
--- a/docker/wmagent/Dockerfile
+++ b/docker/wmagent/Dockerfile
@@ -1,4 +1,4 @@
-FROM cmssw/cmsweb:20221130-stable
+FROM registry.cern.ch/cmsweb/cmsweb:20221130-stable
 
 RUN echo "Adding the AFS cmst1:zh user"
 RUN groupadd -g 1399 zh
@@ -8,6 +8,7 @@ RUN adduser -u 31961 -g 1399 cmst1
 RUN install -o cmst1 -g 1399 -d /data/srv /data/admin /data/certs
 RUN chown root:1399 /data
 
+ARG WMA_TAG=None
 ENV WDIR=/data
 
 # Add install script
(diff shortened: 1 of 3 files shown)


docker/wmagent/install.sh
```diff
@@ -51,8 +56,8 @@ mkdir -p $DEPLOY_DIR || true
 cd $BASE_DIR
 rm -rf deployment deployment.zip deployment-${DEPLOY_TAG};
 
-set -e 
-wget -nv -O deployment.zip --no-check-certificate https://github.com/dmwm/deployment/archive/$DEPLOY_TAG.zip
```
amaltaro, line 55: Todor, if I pass `master` as the DEPLOY_TAG, this line seems to work just fine. And so it does for an old HG tag. Do we really need to change this github url?

docker/wmagent/Dockerfile
```diff
@@ -8,6 +8,7 @@ RUN adduser -u 31961 -g 1399 cmst1
 RUN install -o cmst1 -g 1399 -d /data/srv /data/admin /data/certs
 RUN chown root:1399 /data
 
+ARG WMA_TAG=None
```
amaltaro, line 11: Where does `WMA_TAG` come from? Shouldn't this be using the ENV keyword instead of ARG?

docker/wmagent/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM cmssw/cmsweb:20221130-stable
+FROM registry.cern.ch/cmsweb/cmsweb:20221130-stable
```
amaltaro, line 1: I think we have a more recent cmsweb image. I'd suggest to update it to the latest.

amaltaro (changes requested):
@todor-ivanov Todor, I assume you wanted to update this deployment model to the recent changes we had in WMCore, so I left a few comments along the code for your consideration.

todor-ivanov:
Hi @amaltaro thanks for looking into this PR as well. 

This one was actually just for me, to keep track of all the changes I had to do in order to make this old deployment method work and then use it as initial investigation on how to continue with the Python based docker image. If we want we may finalize this of course. Up to us.

amaltaro:
@todor-ivanov I guess this is no longer relevant. If so, should we close this out?

todor-ivanov:
I believe so. Closing it now.

PR #1365, opened 2023-04-24
Add proxy-secrets to cmsweb prometheus in monitoring namespace

mrceyhun:
- Because crabserver cherrypy process exporter requires proxy certificates to query `:port/crabserver/metrics` which is behind service authentication, cmsweb Prometheus needs proxy-secrets in `monitoring` namespace.
- Requires `cron-proxy` deployment and `proxy-secrets` secret creation in `monitoring` namespace from CMSWEB operators.

Together with https://github.com/dmwm/CMSKubernetes/pull/1356 , this is ready to deploy in testbed (not urgent).

fyi @mapellidario @novicecpp

### DIFF
--- a/kubernetes/cmsweb/monitoring/prometheus.yaml
+++ b/kubernetes/cmsweb/monitoring/prometheus.yaml
@@ -27,12 +27,18 @@ spec:
               mountPath: /etc/prometheus
             - name: prometheus-storage-volume
               mountPath: /prometheus/
+            - name: proxy-secrets
+              mountPath: /etc/proxy
+              readOnly: true
       volumes:
         - name: prometheus-secrets
           secret:
             secretName: prometheus-secrets
         - name: prometheus-storage-volume
           emptyDir: {}
+        - name: proxy-secrets
+          secret:
+            secretName: proxy-secrets
 ---
 apiVersion: v1
 kind: Service


arooshap:
@mrceyhun I have deployed this PR and the other one with the relevant secrets. Kindly validate it, thanks!

mrceyhun:
@mapellidario @novicecpp as you know this config is also not scraping the metrics because of CherryPy's its own authentication and it is impossible to define required CherryPy authentication headers in Prometheus configuration. To overcome this obstacle, you can use this basic script https://github.com/dmwm/cmsweb-exporters/pull/11 or `cmweb_ping` as you wish. 

I just wanted to report here for future reference.

mapellidario:
Thanks Ceyhun. What do you think, do you still want to merge this PR or should we just drop this option, since you found an alternative one?

mrceyhun:
@mapellidario It seems that I wrote to wrong PR. Because we will not need proxy secrets, I'll drop this PR. But we'll use this PR https://github.com/dmwm/CMSKubernetes/pull/1356/files with a bit refactored, which introduces container port service discovery which is required even in the alternative version. Crabserver pod will expose metrics in 2 container so we'll need container port discovery as suggested by Prometheus community.

PR #1369, opened 2023-04-28
Tell Helm Not To Uninstall a Resource

arooshap:
The annotation "helm.sh/resource-policy": keep instructs Helm to skip deleting this resource when a helm operation (such as helm uninstall, helm upgrade or helm rollback) would result in its deletion. 

FYI @todor-ivanov.


todor-ivanov:
thanks @arooshap
Are we currently using helm for deploying the MongoDB clusters in production?

arooshap:
@todor-ivanov yes, we use this helm chart.

PR #1371, opened 2023-05-08
Remove decommissioned NATS cluster manifests and dockerfiles

mrceyhun:
- Remove NATS services and Dockerfiles from CMS monitoring infrastructure


vkuznet:
@mrceyhun , I would suggest only to remove secrets and deployment but keep docker and manifest files since we can come back to them in a future if it will be required. We can keep them around but do not build images and do not deploy services.

PR #1378, opened 2023-05-24
Skip application_dir when parsing WMCore configuration files for process_exporter

amaltaro:
Take 2 fixing WMCore issue https://github.com/dmwm/WMCore/issues/11454

Real fix is just the extra exclusive grep for application_dir, otherwise we get 2 strings assigned to the `app` and it messes up with the command to be executed. See config:
https://gitlab.cern.ch/cmsweb-k8s/services_config/-/blob/prod/reqmon/config.py#L24-L25

Other changes are only cosmetic and making it slightly more verbose during container startup.

### DIFF
--- a/docker/pypi/dmwm-base/monitor.sh
+++ b/docker/pypi/dmwm-base/monitor.sh
@@ -1,20 +1,21 @@
 #!/bin/bash
 
+echo -e "\nTrying to start process_exporter..."
 # start process exporter
 configs="config config-monitor config-output config-transferor config-ruleCleaner config-unmerged"
 for p in $configs; do
     if [ -f /etc/secrets/${p}.py ]; then
-        echo "use /etc/secrets/${p}.py"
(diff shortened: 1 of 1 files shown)


vkuznet (approved):
Fine with me, but may be worth to revisit entirely how we define the `app` because so far it depends on many factors how we structure the config file. May be better approach would be to create dedicated settings in config for that purpose (this is just a suggestion though).

amaltaro:
Please don't merge it yet. Further changes are required.

amaltaro:
Valentin, on top of what you reviewed, I have made a 2nd commit to fix the metrics prefix.
Please review and merge it if everything looks Okay. I will then proceed with the dmwm-base docker image creation and update all the wmcore dockerfiles to the new image.

vkuznet (approved):

amaltaro:
The `cmsweb/dmwm-base` image has been created and uploaded with CERN registry. Tag name is: `pypi-20230525`

PR #1380, opened 2023-05-25
Fix prometheus monitoring for reqmgr2/reqmon/globalworkqueue

amaltaro:
Fixes https://github.com/dmwm/WMCore/issues/11598

First commit fixes prometheus monitoring configuration in the kubernetes manifests:
* fixes port used in `reqmgr2`
* enabled scraping for `reqmon` and `globalworkqueue`

Second commit fixes prometheus monitoring configuration in the helm charts:
* fixes port used in `reqmgr2`
* enable scraping for `globalworkqueue` (reqmon is already setup)

@arooshap as we discussed over Zoom, could you please review this. If everything is Okay, please go ahead and merge it as well.
FYI @vkuznet @todor-ivanov

### DIFF
--- a/helm/reqmgr2/templates/deployment.yaml
+++ b/helm/reqmgr2/templates/deployment.yaml
@@ -47,7 +47,7 @@ spec:
         - containerPort: 8246
           protocol: TCP
           name: reqmgr2
-        - containerPort: 18240
+        - containerPort: 18246
           protocol: TCP
           name: reqmgr2-mon
         command:
--- a/helm/reqmgr2/templates/service.yaml
+++ b/helm/reqmgr2/templates/service.yaml
@@ -10,7 +10,7 @@ spec:
     - port: 8246
       targetPort: 8246
       name: reqmgr2
-    - port: 18240
-      targetPort: 18240
+    - port: 18246
+      targetPort: 18246
       name: reqmgr2-mon
 ---
--- a/helm/reqmgr2/values.yaml
+++ b/helm/reqmgr2/values.yaml
@@ -28,7 +28,7 @@ serviceAccount:
 
 podAnnotations:
   prometheus.io/scrape: 'true'
-  prometheus.io/port: "18240"
+  prometheus.io/port: "18246"
 
 securityContext: 
   privileged: true
--- a/helm/workqueue/templates/deployment.yaml
+++ b/helm/workqueue/templates/deployment.yaml
@@ -33,6 +33,10 @@ spec:
         ports:
         - containerPort: 8240
           protocol: TCP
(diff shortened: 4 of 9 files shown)


amaltaro:
@vkuznet thank you for the prompt feedback. Your first bullets is what is fixed here for reqmgr2, which was not following that convention, hence we had issues after the migration to PyPi model and a different way to use the monitoring port.

>  Therefore, in this PR the monitoring ports are exposed as service to outside of container, but it is not strictly required. 

Regarding this, don't we scrape the `metrics` endpoint from outside of the container? If so, then I do not see how we would be able to fetch those metrics with the current setup if the port was only available within the container/localhost.

From your review, it is also not clear to me if you are requesting changes, or just pointing out that there are different ways to scrape prometheus metrics. Could you please clarify it?

arooshap (approved):
Hi @amaltaro, the changes look good to me. Please let me know if you would like Valentin to provide a confirmation before merging.

vkuznet:
The prometheus scrape happens internally within k8s network where (I think) it reads deployment of service section of the yaml file. That's how it knows about pod's ports. While the `Service` section of the yaml manifest file is used to expose specific pod service to outside of deployment. Said that, it is worth to double check this again, @arooshap could you please do that and document this appropriately. We do have mixture of deployments where we expose or not monitoring metrics, this can be easily seen via `kubectl get services -n <namespace>` command. @arooshap it will be useful if we'll align all service though to use one common approach, e.g. ensure that all pods has exposed monitoring port.

Said that, I'm fine with these changes too.

arooshap:
@amaltaro yes, I believe we can merge this PR as you made all the changes that we discussed on Zoom.

amaltaro:
Thanks Aroosha! Go ahead with the merging at your convenience.

arooshap:
Hi @vkuznet, I just double-checked and we certainly don't require the pod to be exposed for the Prometheus to be able to scrape logs. It is required only when Prometheus is running outside the cluster(which is not the case with our deployments). 
I tested this with the reqmgr2 service by first editing the service to only have `8246` port, and restarting the deployment. Just to be sure, I also checked with Ceyhun and he confirmed that for the internal Prometheus deployment, we don't need to expose the port. 

The next week, I will be coordinating with the MONIT team to have proper and detailed documentation for the CMSWEB monitoring. I will make sure to also include this part in the documentation.

vkuznet:
@arooshap , thanks for checking this out. Based on what you provide and security implication I suggest to remove monitoring port from `Service` sections of all manifest files since we do not use, and therefore nobody should use it from outside the cluster.

PR #1385, opened 2023-05-25
Increase montitoring HA clusters node memory

mrceyhun:
After recent incidence in HA2 VictoriaMetrics and increased number of CMSWEB cluster that remote-writes to our VictoriaMetrics, we need to increment HA node flavor. 

- Improve create_cluster_ha.sh script
- Change node flavor from `m2.xlarge` to `m2.2xlarge `
- Decrease node count from 3 to 2 which is a gain.


mrceyhun:
And update image tags

PR #1386, opened 2023-05-26
Synchronize the deployment procedure to have the prometheus port across all the services.

arooshap:
Hi @vkuznet,

As per your suggestion, I had a look at all the manifest files for all the services that we have user the cmsweb/services section, and I made the changes wherever we declared the container port in the Deployment section for the monitoring, or where we had an annotation to scrape the metrics.

I also checked to make sure that the deployment that are scraping the metrics(and have this port already) are using the correct ports to ensure we don't run have a problem like yesterday. Apart from that, there are some exceptional services that I would like you to comment on if you have the time:

- **dqm-square**: The port is exposed, but the annotation for scraping the pods does not exist. [1]
- **auth-proxy-server, x509-proxy-server, scitokens-proxy-server**: We do have separate pods for monitoring, but should we expose them? Or can we stick to the current procedure?
- **frontend**: we can expose the 18443 port, but do we really need that? I will make the changes if you say.
- **WMStats**: We are using the same port, 8360, for Prometheus. 

Please let me know if you would like me to make any other changes as well, at your convenience. 

Thank you for your time.

[1]. https://github.com/dmwm/CMSKubernetes/blob/beec6e2848412b7fb146f5fec129712ad7820001/kubernetes/cmsweb/services/dqmsquare.yaml#L11

### DIFF
--- a/kubernetes/cmsweb/services/dbs2go-global-migration.yaml
+++ b/kubernetes/cmsweb/services/dbs2go-global-migration.yaml
@@ -1,3 +1,15 @@
+kind: Service
+apiVersion: v1
+metadata:
+  name: dbs2go-global-migration
+  namespace: dbs
+spec:
+  selector:
+    app: dbs2go-global-migration
+  ports:
+    - port: 19251
+      targetPort: 19251
+      name: dbs-mon
 ---
 apiVersion: v1
 kind: ConfigMap
--- a/kubernetes/cmsweb/services/dbs2go-phys03-migration.yaml
+++ b/kubernetes/cmsweb/services/dbs2go-phys03-migration.yaml
@@ -1,4 +1,17 @@
 ---
+kind: Service
+apiVersion: v1
+metadata:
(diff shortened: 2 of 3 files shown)


vkuznet:
@arooshap , my suggestion to remove all monitoring ports for all services from `Service` section of manifest files since we do not use it outside the port (and k8s Prometheus can find it internally). For the rest of the services I suggest the following:
- if no monitoring is provide, no port declaration or annotation should be made
- we do need monitoring for `auth-proxy-xxx` or `frontend`, and in fact we do have `frontend` dashboard, but since you checked that we do not need to expose ports because local Prometheus can scrape metrics then monitoring ports in `Service` section of manifest files should be removed
- all ports for monitoring should start with 10000, it is our convention, therefore WMStats monitoring port should be adjusted.

vkuznet (changes requested):
Please see my previously posted comment and make necessary changes.

PR #1388, opened 2023-05-31
Removing custom configs for individual services because we no longer …

arooshap:
…need them, thanks to Kubernetes Discovery.


vkuznet (approved):
@arooshap , the changes look good to me and I understand that from now on we will only rely on k8s port discovery which you tested.

PR #1393, opened 2023-06-30
Changes to install WMAgent from PyPi; provided install and run.sh scripts

amaltaro:
Fixes https://github.com/dmwm/WMCore/issues/8797
Fixes https://github.com/dmwm/WMCore/issues/11564
Superseeds https://github.com/dmwm/CMSKubernetes/pull/1364

DISCLAIMER: this contains exactly the same changes that @todor-ivanov provided in #1364 (picked with a curl + patch command), plus a 2nd commit that I made to fix how TAG vs WMA_TAG is used in the Dockerfile. This was done just so we can get this merged before the quarter is over :-) Now I am pasting the original description provided by Todor, but please check the original PR for the full discussion and potential open questions.

With the current PR we modify the Dockerfile  such that:
* It calls the install.sh script instead of directly executing `pip install wmagent`
* It accept the WMA_TAG as  a parameter at  build time with `--build-arg WMA_TAG=<WMA_TAG>`
* It creates the expected GUID and UID which should match with the respective ones  at the VM 

The following new scripts are provided as well:
* We also add the basic `install.sh` script, which is configuring the whole deployment structure and the future places for the credentials mount points.

* We also add the `run.sh` script to configure and start the container a the host

* We also provide two wrapper scripts: `wmagent-docer-build.sh` and `wmagent-docker-run.sh`, which are taking care of all docker configs and parameters for both build and runtime.

**NOTE:** 
The minimal WMAgent tag to be used for this PR should be `2.2.1rc3`

**NOTE:**
The files: 
* `etc/local.ini` 
* `etc/my.cnf`
* `etc/rucio.conf`
* `bin/manage`  
Are a one to one copy from the old [deployment repository](https://github.com/dmwm/deployment/tree/master/wmagentpy3) (master branch), except a minor change in the few environment variables read by the `bin/manage` script, which I will explicitly point here: 
```
[user@unit01 data]$ diff -u  /data/CMSKubernetes/docker/pypi/wmagent/bin/manage  /data/deployment/wmagentpy3/manage 
--- /data/CMSKubernetes/docker/pypi/wmagent/bin/manage	2023-05-12 17:53:15.536200825 +0200
+++ /data/deployment/wmagentpy3/manage	2023-05-03 13:10:45.040435640 +0200
@@ -4,7 +4,10 @@
 #
 # Global variables etc
 #
-ROOT_DIR=$WMA_CURRENT_DIR
+THIS_SCRIPT=$(readlink -f $0)
[...]

PR #1394, opened 2023-07-03
Create a wmagent-base Dockerfile with all the relevant scripts; use docker build args

amaltaro:
Fixes https://github.com/dmwm/WMCore/issues/11647
Depends on https://github.com/dmwm/WMCore/pull/11638

This **was** meant to resolve an issue with the WMCore CD pipeline, problem inserted with: https://github.com/dmwm/CMSKubernetes/pull/1393, which has been fixed with https://github.com/dmwm/WMCore/pull/11639

Nonetheless, given that these changes had already been made, I refactored and rebased this branch to provide:
* `wmagent-base`: installing the OS-based dependencies for WMAgent
* `wmagent`: running the actual wmagent deployment, including all the needed scripts.

The benefits of this separation are:
* don't install all the OS libraries every time a new WMAgent/WMCore tag is made
* make WMAgent images more reproducible, by removing potential unwanted upgrades of the OS libraries
* quicker build time for wmagent


docker/pypi/wmagent-base/install.sh
```diff
@@ -114,7 +114,7 @@ echo "-----------------------------------------------------------------------"
 stepMsg="Replace the current /data/manage script coming from 'dmwm-base' image with a symlink link"
 echo "-----------------------------------------------------------------------"
 echo "Start $stepMsg"
-[[ -f /data/manage ]] && rm -f /data/manage && ln -s $WMA_MANAGE_DIR/manage /data/manage
+[[ -f /data/manage ]] && ln -s $WMA_MANAGE_DIR/manage /data/manage
```
todor-ivanov, line 117: Hi @amaltaro I know  you have not yet  asked for a review on this, but just noticed this change. This line would actually always fail. Let me put it with an example:
```
$ touch test.file
$ touch test.link
$ [[ -f test.link ]] && ln -s test.file test.link
ln: failed to create symbolic link 'test.link': File exists
``` 

Now let me also put the context of why the `&& rm -f /data/manage &&` was added at the first place. This file comes as the original `manage` script provided by the `dmwm-base` image [here](https://github.com/dmwm/CMSKubernetes/blob/11a4962c3dde96c639b62cbf5b3046bd06f3882c/docker/pypi/dmwm-base/Dockerfile#L18) and is completely different  from what we create  for the agent at `$WMA_MANAGE_DIR/manage`. So I decided to remove it an convert it from an executable file to a soft link pointing to the actual `manage` script that we are using with the agent. I could have left it untouched, but this would cause confusion as we would have two completely different manage scripts from two completely different sources in the same image, one of which is absolutely not related to the current component/service and impossible to work.

amaltaro:
With the current changes, I could build both `wmagent-base` and `wmagent` images, the former has also been uploaded to the CERN registry as `wmagent-base:pypi-20230703`.

From my side, this is good to go. @todor-ivanov could you please review? Thanks

docker/pypi/wmagent-base/Dockerfile
```diff
@@ -0,0 +1,24 @@
...
+RUN apt-get install -y hostname net-tools iputils-ping procps && apt-get clean
+
+# copy oracle client:
+COPY --from=oracle /usr/lib/oracle /usr/lib/oracle
+ENV LD_LIBRARY_PATH=/usr/lib/oracle
+ENV PATH=$PATH:/usr/lib/oracle
+ENV PKG_CONFIG_PATH=/usr/lib/oracle
+
+# Add all deployment needed directories, install and run scripts
+ADD etc /data/etc
```
todor-ivanov, line 19: Please do not hardcode absolute paths, but do use relative paths originating everything from a single $WMA_ROOT_DIR as it was done in
https://github.com/dmwm/CMSKubernetes/pull/1364/files#diff-6d65f482db9361d53e7c64522f0be43dc64fde7f6999426c0657cb064a74283eR23-R37

We have seen already how we easily start juggling with paths and we eventually end up with few of them unlinked to a single root. And this  is a path to unmanageable situation in the future.

This is also needed for the scripts that are to implement an equivalent to the Docker environment.

todor-ivanov: same for all lines bellow containing `/data` as hardwired path

todor-ivanov, line 19: Are you absolutely sure that the contents in the `etc` you are adding here are coming from the correct source. Meaning from here: https://github.com/dmwm/CMSKubernetes/tree/master/docker/pypi/wmagent/etc 

If we are still copying only the Dockerfile in the CI pipeline I am not sure how could we add it without any error in this `wmagent-base` image. Have you inspected this in the image actually crated/uploaded to CERN registry?

todor-ivanov (changes requested):
Thanks for creating this PR @amaltaro. This one does improve the situation. 
I have left few important comments inline. They are about changes you introduce that would eventually break other pieces depending on those deployments here. Please take a look.

docker/pypi/wmagent-base/Dockerfile
amaltaro: Yes, I did check the `wmagent-base` image and $WMA_ROOT_DIR/etc contains those 3 configuration files that have now been moved under the `wmagent-base` directory.

docker/pypi/wmagent-base/install.sh
```diff
@@ -111,10 +111,10 @@ echo "WMA_BUILD_ID preserved at: $WMA_ROOT_DIR/.dockerBuildId "
 echo "Done $stepMsg!" && echo
 echo "-----------------------------------------------------------------------"
 
-stepMsg="Replace the current /data/manage script coming from 'dmwm-base' image with a symlink link"
+stepMsg="Make a symlink from the original manage script from wmagent-base to the agent expected location"
 echo "-----------------------------------------------------------------------"
 echo "Start $stepMsg"
-[[ -f /data/manage ]] && rm -f /data/manage && ln -s $WMA_MANAGE_DIR/manage /data/manage
+[[ -f /data/manage ]] && ln -s /data/manage $WMA_MANAGE_DIR/manage
```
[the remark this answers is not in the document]
amaltaro: My understanding is that we need to use the `manage` script that is now shipped with CMSKubernetes, right?
With the changes provided in this PR, the `wmagent-base` image will provide that `manage` script under /data/ (or $WMA_ROOT_DIR), so this line is creating a symlink in `$WMA_MANAGE_DIR/manage` which points to the container manage script in `/data/manage`. Let me know if I am missing anything though.

amaltaro:
@todor-ivanov thank you for the prompt review. I should have addressed all your concerns; and/or replied inline to things that I consider correct as is. Please have another look

docker/pypi/wmagent-base/Dockerfile
```diff
@@ -0,0 +1,27 @@
...
+
+# copy oracle client:
+COPY --from=oracle /usr/lib/oracle /usr/lib/oracle
+ENV LD_LIBRARY_PATH=/usr/lib/oracle
+ENV PATH=$PATH:/usr/lib/oracle
+ENV PKG_CONFIG_PATH=/usr/lib/oracle
+
+# Add all deployment needed directories, install and run scripts
+ADD etc $WMA_ROOT_DIR/etc
+ADD manage $WMA_ROOT_DIR/manage
```
todor-ivanov, line 23: Isn't the `manage` script origin (as currently placed in the repository) `bin/manage` ? At least it used to be...  in the original PR. I can see you have moved it out of `bin` now. But it was there for a reason - see my comment about the way how  `bin/` and `etc/` should be distributed from both pypi and docker origins.

docker/pypi/wmagent-base/Dockerfile
```diff
@@ -0,0 +1,27 @@
...
+ADD install.sh $WMA_ROOT_DIR/install.sh
```
todor-ivanov, line 24: Reading more about the it I do not think even the `install.sh` and `run.sh` script placement is good to be in the `wmagent-base` image, since they depend on variables and setup parameters that come later with the other image you suggest.

todor-ivanov (changes requested):
Thanks for suggesting these changes  @amaltaro, but the more I read the code the more confusion and complications I see introduced and pieces moved around only because we  refrain to resolve one disadvantage in another place/component of our system. 

Let me now put my pros and cons for this exact PR:
* Adding the TAG as external parameter to the Docker build process is perfect and great. Thanks  for that!
* Splitting the image is great. I strongly support that. But the criteria and motivation used are wrong. What I mean is:
    * Having OS dependencies resolved in the wma-base image is good, but `install.sh` and `run.sh` are pieces related to the `wmagent` installation and initialization process. They should belong to the `wmagent` image eventually. Possibly same for the `manage` script. 
    * Having the need to reorganize Docker files and structures, just because we do not want to fix an error with a different origin is far from optimal.  
    * Shuffling already established and tested  directory structure in the agent image will only lead to errors at runtime and would require future changes for fixing these deviations. (that could be minor though - hopefully I have recognized the few that could be problematic and gave suggestions inline for them) 
    * Messing up with the structure and how  we distribute configuration and executable files is a drawback and blocks future improvements I have planned for in order to resolve the already hugely accumulated  workarounds in the agent  initialization process.

amaltaro:
Todor, after this review I see that I actually removed something that I shouldn't, which is related to:
```
ADD bin $WMA_DEPLOY_DIR/bin
ADD etc $WMA_DEPLOY_DIR/etc
```

I will get it fixed soon.

Now about your concerns:

> Having OS dependencies resolved in the wma-base image is good, but install.sh and run.sh are pieces related to the wmagent installation and initialization process. They should belong to the wmagent image eventually. Possibly same for the manage script.

Honestly, I does not really matter IMO if it is in the wmagent or wmagent-base images. The way I see the separation of images is the following: a) wmagent-base to be rebuilt in every WMAgent upgrade release cycle (such that we can pull any updates to those underlying libraries); b) while the wmagent image will be rebuilt for every new tag in WMCore. Having said that, install.sh, run.sh and manage will hardly change and I don't think it's makes any difference to say that it should belong to one or another.

> Having the need to reorganize Docker files and structures, just because we do not want to fix an error with a different origin is far from optimal.

My same argument as above. I don't understand why such a fuss about this.

> Shuffling already established and tested directory structure in the agent image will only lead to errors at runtime and would require future changes for fixing these deviations. (that could be minor though - hopefully I have recognized the few that could be problematic and gave suggestions inline for them)

Other than the mistake I just mentioned above - which I will fix soon - ALL the changes I made here should be compatible with your recent changes as well. It's just a step earlier than it was!

> Messing up with the structure and how we distribute configuration and executable files is a drawback and blocks future improvements I have planned for in order to resolve the already hugely accumulated workarounds in the agent initialization process.

I think I answered this above as well. There is no "mess up", data now is where it was before, the only difference is that it comes with the wmagent-base image instead of being added by the wmagent image.

I see you are frustrated with these changes though, so if you prefer, I am happy to:
* roll back your CMSKubernetes changes
* reopen the WMCore GH issues (one more in addition to the one opened)
* we reorganize plans for Q3
* and you take your time to to the optimal development that you have in mind

This way we can move forward with the testbed validation that cannot start because we cannot even build images right now! Let me stress it once again that production upgrade is scheduled for the next Tuesday, with enough heads up to make sure we don't delay things!

amaltaro:
Actually the `bin` directory is something that I got rid off, given that right now it only provides the manage script. IMO, if it's a single script, we better provide it in the base directory as it is done for the other WM images. But if you prefer I can revert that as well.

docker/pypi/wmagent-base/Dockerfile
```diff
@@ -0,0 +1,27 @@
+FROM registry.cern.ch/cmsweb/oracle:21_5-stable as oracle
+FROM registry.cern.ch/cmsweb/dmwm-base:pypi-20230525
+MAINTAINER Valentin Kuznetsov vkuznet@gmail.com
+
+# directory where we place all the relevant scripts and configurations for WMAgent docker image
+ENV WMA_ROOT_DIR=/data
```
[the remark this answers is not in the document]
amaltaro: I think we can keep these under WMA_ROOT_DIR, then in the wmagent image we move them under WMA_DEPLOY_DIR. I am making this change.

docker/pypi/wmagent-base/Dockerfile
amaltaro: I just replied in the main pull request on this. But can you please clarify why `manage` must be in the `bin` directory? It is a single script, that's why I moved it out. It will be taken care in the wmagent image though with the same logic that you implemented (other than contained in bin).

docker/pypi/wmagent-base/install.sh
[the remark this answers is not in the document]
amaltaro: I am going to revert these changes then, but from what I checked it was correct.

docker/pypi/wmagent/Dockerfile
```diff
@@ -48,16 +33,6 @@ RUN rm -f /etc/mysql/mariadb.conf.d/50-server.cnf
 # Add WMA_USER to sudoers
 RUN echo "${WMA_USER} ALL=(ALL) NOPASSWD:ALL" >> /etc/sudoers
 
-# Add all deployment needed directories
-ADD bin $WMA_DEPLOY_DIR/bin
```
amaltaro, line 52: @todor-ivanov by the way, why is the manage script under the bin directory? It should have been WMA_MANAGE_DIR, if you wanted to keep a similar structure to what we have now.

todor-ivanov:
@amaltaro please take a look at this PR https://github.com/dmwm/WMCore/pull/11639, which is intended to fix only the broken GitHub actions. We should deal with all those quite intrusive and large developments on splitting the `wmagent` image under a separate issue. 

With the current PR we can still adopt the `TAG` parametrisation, if we still want it though.

todor-ivanov:
hi @amaltaro 
>  a) wmagent-base to be rebuilt in every WMAgent upgrade release cycle (such that we can pull any updates to those underlying libraries); 
> b) while the wmagent image will be rebuilt for every new tag in WMCore

This was yet another hing I was concerned and wanted to mention earlier but simply forgot. The fact that we are going to use different methods for building those two images:
* manual for `wmagent-base` 
* automated for `wmagent` 

Is just another source of confusion. I am pretty much in favor of them both being properly covered by the CI/CD actions. 

Let me suggest something here:
* We eventually converge on the suggested  fix I have made with this PR: https://github.com/dmwm/WMCor/pull/11639  It will fix only the broken part and nothing more.
* We do NOT rollback the CMSkubernetes changes, but keep them, because the above is definitely going to fix the issue
* We create a GH issue on splitting the `wmagent` image the way you suggest here and make it part of the WMAgent containerization  meta issue: https://github.com/dmwm/WMCore/issues/11314 We can easily resolve this one later during Q3 with medium priority taking the part of this PR that relates to the image splitting 
* We create a GH issue just for adding the `TAG` as a parameter to the docker build process
* We resolve the `TAG` related issue with this PR (leaving behind the part of the image splitting - this should go as a resolution to the previously mentioned issue). And we can resolve this one here on our pace. e.g. upon the start of the validation process for Monday.

todor-ivanov:
If we agree, I can put the above plan in motion immediately.

amaltaro:
@todor-ivanov after the latest updates in WMCore, I didn't want to lose the initial work towards the Dockerfile separation. So I took the opportunity to refactor this branch. Initial description has been updated as well. 
Please have a look whenever it's convenient to you.

todor-ivanov:
thanks @amaltaro 
I am looking again at this PR

todor-ivanov (approved):
Thanks @amaltaro  for completing this separation of the `wmagent` image. 

Things look really good now with your last commit. This PR here slices the construction at the exact place now. The comment I made above is just for informative purposes, it is not a request for change. 

p.s. This PR is currently solving an already resolved issue. You may want to quickly create  a short issue related only to the splitting of the wmagent image as part of the meta issue for wmagent containerization, By resolve it when merging this PR we will not miss to account this work in Q3, as it should be. Because, even though it originates from a fix of another issue, the final result is good and is worth keeping track of it.

amaltaro:
Having a fresh look at the image layers/sizes, here is the new `wmagent-base` image:
```
$ docker image ls
REPOSITORY                                 TAG                   IMAGE ID            CREATED             SIZE
registry.cern.ch/cmsweb/wmagent-base       pypi-20230706         a63c35d9737f        2 minutes ago       1.64GB
```

and this is a break down of the layers:
```
$ docker history a63c35d9737f
IMAGE               CREATED             CREATED BY                                      SIZE                COMMENT
a63c35d9737f        3 minutes ago       /bin/sh -c #(nop)  ENV PKG_CONFIG_PATH=/usr/…   0B                  
0da4a97a9570        3 minutes ago       /bin/sh -c #(nop)  ENV PATH=/usr/local/bin:/…   0B                  
cb6ad593a67d        3 minutes ago       /bin/sh -c #(nop)  ENV LD_LIBRARY_PATH=/usr/…   0B                  
9026e2fe9a19        3 minutes ago       /bin/sh -c #(nop) COPY dir:d84ae0cee1c8e3468…   258MB               
17859b8748d2        3 minutes ago       /bin/sh -c apt-get install -y hostname net-t…   4.02MB              
4642b4a59bcd        3 minutes ago       /bin/sh -c apt-get install -y libmariadb-dev…   344MB               
c9bc41dbda1d        4 minutes ago       /bin/sh -c apt-get update                       700kB               
6ad7eee5d71b        6 weeks ago         /bin/sh -c #(nop) WORKDIR /data                 0B                  
9d65b3fbd076        6 weeks ago         /bin/sh -c #(nop) ADD file:62b40ab409929052d…   2.62kB              
6e95acab4053        6 weeks ago         /bin/sh -c #(nop) ADD file:d542a6348f3d47c04…   1.43kB              
8cf698dcd460        6 weeks ago         /bin/sh -c #(nop) ADD file:4b52f467960680022…   2.81kB              
90aefde5fd5f        6 weeks ago         /bin/sh -c pip install dbs3-client              10MB                
36a61becd113        6 weeks ago         /bin/sh -c #(nop) COPY file:179dde413893af6e…   8.22MB              
acb82929238d        6 weeks ago         /bin/sh -c #(nop) COPY file:0efefb1ce6f445b6…   8.04MB              
56fc2eccd834        6 weeks ago         /bin/sh -c #(nop) COPY file:bdd094e203553774…   4.53MB              
b2caff57abe0        6 weeks ago         /bin/sh -c #(nop) COPY dir:fdfb134dd7bef922f…   184B                
96374bb478a7        6 weeks ago         /bin/sh -c #(nop) COPY dir:58a0efbd777b5881b…   198B                
66402d648958        6 weeks ago         /bin/sh -c #(nop) COPY dir:4b189a3d1eebc67f3…   431kB               
185f219e84a2        6 weeks ago         /bin/sh -c mkdir /etc/grid-security             0B                  
102152270db3        6 weeks ago         /bin/sh -c ln -s /usr/bin/python3 /usr/bin/p…   16B                 
bed23823e718        6 weeks ago         /bin/sh -c apt-get install -y curl vim libcu…   79.3MB              
bc9f4d7dbc30        6 weeks ago         /bin/sh -c apt-get update                       18MB                
5a1eb91236e3        6 weeks ago         /bin/sh -c #(nop)  MAINTAINER Valentin Kuzne…   0B                  
0869a4e92f01        6 weeks ago         CMD ["python3"]                                 0B                  buildkit.dockerfile.v0
<missing>           6 weeks ago         RUN /bin/sh -c set -eux;   wget -O get-pip.p…   10.3MB              buildkit.dockerfile.v0
<missing>           6 weeks ago         ENV PYTHON_GET_PIP_SHA256=96461deced5c2a487d…   0B                  buildkit.dockerfile.v0
<missing>           6 weeks ago         ENV PYTHON_GET_PIP_URL=https://github.com/py…   0B                  buildkit.dockerfile.v0
<missing>           6 weeks ago         ENV PYTHON_SETUPTOOLS_VERSION=57.5.0            0B                  buildkit.dockerfile.v0
<missing>           6 weeks ago         ENV PYTHON_PIP_VERSION=22.0.4                   0B                  buildkit.dockerfile.v0
<missing>           6 weeks ago         RUN /bin/sh -c set -eux;  for src in idle3 p…   32B                 buildkit.dockerfile.v0
<missing>           6 weeks ago         RUN /bin/sh -c set -eux;   wget -O python.ta…   42.7MB              buildkit.dockerfile.v0
<missing>           6 weeks ago         ENV PYTHON_VERSION=3.8.16                       0B                  buildkit.dockerfile.v0
<missing>           6 weeks ago         ENV GPG_KEY=E3FF2839C048B25C084DEBE9B26995E3…   0B                  buildkit.dockerfile.v0
<missing>           6 weeks ago         RUN /bin/sh -c set -eux;  apt-get update;  a…   18.5MB              buildkit.dockerfile.v0
<missing>           6 weeks ago         ENV LANG=C.UTF-8                                0B                  buildkit.dockerfile.v0
<missing>           6 weeks ago         ENV PATH=/usr/local/bin:/usr/local/sbin:/usr…   0B                  buildkit.dockerfile.v0
<missing>           6 weeks ago         /bin/sh -c set -ex;  apt-get update;  apt-ge…   529MB               
<missing>           6 weeks ago         /bin/sh -c apt-get update && apt-get install…   152MB               
<missing>           6 weeks ago         /bin/sh -c set -eux;  apt-get update;  apt-g…   28.6MB              
<missing>           6 weeks ago         /bin/sh -c #(nop)  CMD ["bash"]                 0B                  
<missing>           6 weeks ago         /bin/sh -c #(nop) ADD file:150a6453ab2258061…   124MB   
```

eventually I need to research what the "missing" tag is :)

I also tested moving all those packages back under the same `apt-get install`, and the size of the final image is exactly the same.

Checking the image size of the final product `wmagent`, it seems to be unchanged:
```
$ docker image ls
REPOSITORY                                 TAG                   IMAGE ID            CREATED             SIZE
registry.cern.ch/cmsweb/wmagent            2.2.2rc5              7685798d6309        8 minutes ago       2.03GB
```
@todor-ivanov

todor-ivanov:
Thanks @amaltaro
I did not expect the final product size to be altered, because we fully include the `wmaget-base` image inside the `wmagent` image. On the other hand it is now well distinguished which are the OS dependencies and which are the wmagent related pieces. Which I do consider very good.

The line that I was quite interested in was:
```
7859b8748d2        3 minutes ago       /bin/sh -c apt-get install -y hostname net-t…   4.02MB 
```
Which means basically those packages add up no more than 4 MB, which to me seems quite acceptable, given the fact that those provide basic capabilities in the running agent.

amaltaro:
Yes I agree. It's great to see that the auxiliary libraries are pretty small. Okay, I am going to get this one merged then.

PR #1405, opened 2023-07-24
Modifications to the secret deployment to only delete the decrypted secrets.

arooshap:


todor-ivanov (commented):
Thanks @arooshap 
Things look much better than before. This change would definitely prevent the deletion of the configuration files, together with the decrypted password files. 

One thing remains, though. Given the fact that the `config` files are altered in place with `set` command, this leads to yet another situation. The final `services_config` clone still ends up with altered working tree upon the use of the `deploy-secrets.sh` script. This prevents the multiple use of the script with the same clone of the repository. People must remember, they need to do `git checkout .` upon every run of `deploy-secrets.sh`. This is quite error prone. 
If I may suggest, in order to work around this situation, the script should actually copy the relevant service config area (the one related to the service it works with during the current execution) in a temporary place, and then perform all the needed actions from there, while preserving the original files untouched in the local clone of the `service_config` repository.

PR #1413, opened 2023-09-06
Add dev dbs2go-global migration services to deploy-srv

amaltaro:
Dennis and I found out today that, depending on the dbs2go instance that is deployed in a development environment (k8s cluster different than prod and preprod), the service manifest file does not have the same string replacements (sed) as the other read and write services, affecting especially the livenessprobe setup and causing the service to be constantly restarted.

With this PR, we treat the migrate (m) and migration dbs2go services equally to the read (r) and write (w) instances

@arooshap could you please review it?
FYI @d-ylee @vkuznet

### DIFF
--- a/kubernetes/cmsweb/scripts/deploy-srv.sh
+++ b/kubernetes/cmsweb/scripts/deploy-srv.sh
@@ -192,7 +192,7 @@ elif  [ "$cmsweb_env" == "k8s-preprod" ] ; then
 
             cat $srv.yaml | sed -e "s,1 #PROD#,,g" | sed -e "s,#PROD#,      ,g" |  sed -e "s,logs-cephfs-claim,$cmsweb_log,g" | sed -e "s, #imagetag,$cmsweb_image_tag,g" | sed -e "s,k8s #k8s#,$cmsweb_env,g" | sed -e 's+crabserver/prod+crabserver/preprod+g' |  kubectl apply -f -
  
-        elif [[ "$srv" == "dbs2go-global-r" || "$srv" == "dbs2go-global-w" || "$srv" == "dbs2go-phys03-r" || "$srv" == "dbs2go-phys03-w" || "$srv" == "dbs2go-global-m" || "$srv" == "dbs2go-phys03-m" || "$srv" == "dbs2go-global-migration" || "$srv" == "dbs2go-phys03-migration" ]] ; then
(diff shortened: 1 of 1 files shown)


muhammadimranfarooqi:
@amaltaro 

If dbs python services are not required to be deployed anymore in any k8s cluster, then this PR looks fine.

muhammadimranfarooqi:
> The changes look good to me provided we no longer need to deploy old dbs images! :)

Thats what I said but with a longer sentence :laughing:

arooshap:
> > The changes look good to me provided we no longer need to deploy old dbs images! :)
> 
> Thats what I said but with a longer sentence 😆

Oh, I missed your reply.

amaltaro:
@arooshap @muhammadimranfarooqi yes, and I am not sure about that :-)

Are we still running any of the python-based instances? If not, when where they stopped?
From my perspective, it's been more than a year that we migrated to dbs2go and I do not recall of ever going back to the python-based for anything. So I would be in favor of not deploying it anymore and soon (end of the year?) remove everything related to it from CMSKubernetes and services_config.

If you prefer to have those changes back, I can update this PR while we keep discussing it as well.

muhammadimranfarooqi:
@amaltaro 

During migration, we intentionally kept dbs python services and added redirect rules in FE too for them. Its question for @vkuznet whether dbs python related stuff should be removed permanently now?

vkuznet:
The presence of DBS Python was requested by CRAB team at a time. And, if I recall correctly we do not need phys03 migration. Therefore, I suggest that you consult further with this with @belforte and the team before you'll merge.

belforte:
CRAB does not use python-based DBS servers any more. We only use dbs20 at cmsweb.cern.ch/dbs.
We still need to be able to migrate datasets from global to phys03. This migration functionality will be needed "forever" until/unless there is some redesigning of the way user datasets are handled.

amaltaro:
Even though nobody seems to be using the old implementation of the DBS server (python based), we are still running them in production. So I decided to leave that discussion/decision outside of this PR.

PR has been updated and the python-based dbs services have been recovered in deploy-srv. 
Please review this pull request again and merge it if it's fine for you.

amaltaro:
Please let me know if there is anything left pending with this PR. From my side, it is good to go.

arooshap (approved):

PR #1414, opened 2023-09-20
Update MSUnmerged resource limits

todor-ivanov:
Fixes https://github.com/dmwm/WMCore/issues/11701


amaltaro:
Just a note, I don't think it really fixes the issue mentioned, but at least it should keep the service more stable, in case it is getting restarted because it runs out of resources.

PR #1427, opened 2023-10-13
crabserver helm - test liveness readiness probes

mapellidario:
We use specific oracle schema owners for test clusters, which differ from the one we use for preprod. 

In order not to use the preprod schema for the liveness and readiness probe in test clusters, we need to use new liveness and readiness probes endpoints just for test clusters, identified as all the clusters that are not `prod` or `preprod`.

tested in `test11`.


mapellidario:
fyi: @novicecpp

PR #1435, opened 2023-10-24
Updated dbs2go Dockerfile

d-ylee:
Since oracle:21_5 is not available in the registry, the dbs2go Dockerfile has updated the image version to oracle:21_5-stable


d-ylee:
@arooshap I think this is ready.

PR #1436, opened 2023-10-26
Mongo eos

arooshap:


docker/mongodb/source/run.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+### This script relies on provided keytab file which will be
+### be mounted to /etc/secrets area
+if [ -f /tmp/krb ]; then
+  export keytab=/tmp/krb/cmsweb.keytab
```
vkuznet, line 5: I'm not sure I understand logic here. The comment says it relies on /etc/secrets area while you use /tmp/rkb, and I checked k8s manifest which also provides keytab file in /tmp/krb. Is there are any reason why it should be in /tmp/krb and not in /etc/krb for instance?

arooshap: @vkuznet, sorry that is a wrong path in the comment. I will modify it.

vkuznet:
@arooshap , now everything is ok, please merge and proceed.

PR #1446, opened 2024-01-11
Remove monitoring MongoDB

nikodemas:
https://cms-dm-monitoring.cern.ch/ tool update to stop using custom monitoring MongoDB and start using common cmsweb MongoDB.

- Delete old MongoDB docker and yaml files
- Change database connection commands accordingly
- Modify deployment files and image versions to align with MongoDB changes
- Unrelated change: update image version for `cron-spark-jobs.yaml` to reflect changes in https://github.com/dmwm/CMSSpark/pull/155

FYI @leggerf @brij01


leggerf (approved):

nikodemas:
We couldn't achieve the same performance with the common cmsweb MongoDB, so I am closing this ticket and returning to the custom MongoDB instance running as a side-container. Related JIRA ticket - `CMSMONIT-516`, related PR - https://github.com/dmwm/CMSKubernetes/pull/1473.

vkuznet:
@nikodemas since I don't know details of benchmark metrics you or someone else tried to compare the performance, but MongoDB performance depends mostly on two factors:
1. the underlying database engine to use, e.g. [WiredTiger](https://www.mongodb.com/docs/manual/core/wiredtiger/)
2. your query results performance really depends on fitting indexes into node RAM
   - here the injection will depend on storage, e.g. SSD vs HDD
   - data lookup will depend on index used to query the data, i.e. if there is no index the full table scan will be performed.
   
Therefore, it is likely that you may not compare apples to apples where your custom MongoDB may reside on better storage, or use different engine, or use indexes which may not be present in common CMSWEB instance.

It would be interesting if you'll post details of comparison as many groups in CMS uses MongoDB and any details can be shared across groups.

nikodemas:
@vkuznet we didn't even try to get any benchmark metrics as we had to wait almost minutes for the new setup to load (compared to seconds in the initial setup). Regarding the indexing and other configs - we used exactly the same setup as we had a separate database in the cmsweb MongoDB instance (you can see the indices [here](https://github.com/dmwm/CMSMonitoring/blob/master/rucio-dataset-monitoring/spark/createindexes.js)). The only clear difference that I could see was that in our setup we simply were using [Pod's memory](https://github.com/dmwm/CMSKubernetes/blob/master/kubernetes/monitoring/services/datasetmon/cmsmon-mongo.yaml#L112-L113) as our storage without attaching it to any PersistentVolume (since we are dropping the data every data and refilling it with new information, this volatility isn't an issue), whereas in the cmsweb setup all of the data is saved in the PersistentVolumes, so I assume that this is causing the delay.

PR #1454, opened 2024-03-07
Changes to the frontend docker image and kubernetes manifest to decrease load on the CRIC servers.

arooshap:
- Tweak the cronjobs to run a bit more "randomly" in the sense that they now run every 4 minutes but based on a clock so it's always 10:04-10:08-10:12 e.t.c. If we could somehow distribute these calls it would maybe stabilize traffic to CRIC.
- For production and testbed: 15 minutes, and random sleep of 10 minutes. For development: 30 minutes and 15 minutes random sleep.


Panos512:
Looks good to me, thanks @arooshap !

PR #1455, opened 2024-03-11
Add SqlCl client

todor-ivanov:
This  PR is resolving the WMAgent related issue for adding sqlcl client: https://github.com/dmwm/WMCore/issues/11927


amaltaro (commented):
Given the dependency on Java Runtime and increase in image size, I would say we should hold off with the adoption of this tool into our stack until we can actually get some experience with the (first) migration to a container solution.

vkuznet:
You may consider package I wrote 2 years ago called [sqlshell](https://github.com/vkuznet/sqlshell). It uses Go DB drivers to provide similar functionality and work across all DB flavors, sqlite, mysql, oracle. Said that, to access DB it requires DB libraries, i.e. for oracle it will use oracle dynamic libraries, but it does not require separate Java environment. Feel free to explore it as alternative.

vkuznet (commented):
Todor, it would be useful to know exact number of increase in image size and number of vulnerabilities when we our intention only to bring a single tool. I provided an alternative https://github.com/vkuznet/sqlshell which is Go based solution and does not require Java. Since we will need ORACLE libraries anyway, I think it will bring nothing in terms of image size and vulnerabilities and at the same time provide similar functionality.

PR #1456, opened 2024-03-13
Fix a MariaDB related bug, caused by the previous PR.

todor-ivanov:
This is a PR meant to resolve a bug caused by a typo in this PR:  https://github.com/dmwm/CMSKubernetes/pull/1451

PR #1457, opened 2024-03-17
init.sh and manage-common.sh fixes for successful initialization

LinaresToine:
PR to fix initialization issues pointed out by @todor-ivanov in https://github.com/dmwm/WMCore/issues/11890#issuecomment-2000075089


todor-ivanov:
@LinaresToine your PR is trying to revert changes that from another one:
https://github.com/dmwm/CMSKubernetes/pull/1456/files 

This may have happened because you forgot to update you local branch before creating the PR.
Could you please first fetch the master branch and fork from there before you create your PR so that you do not try to revert other changes.

LinaresToine:
@todor-ivanov I fixed the PR.

docker/pypi/wmagent/init.sh
LinaresToine, line 1: The _check_mounts() function crashes without the return true statement in the beginning of the function. 

It was one of the changes I was attempting to revert in the original PR. @todor-ivanov

LinaresToine: Indeed, we never talked about this part.  What I see when running the init.sh script in this PR is that the _check_mounts() function is returning false. However, the version of the script we were using during our last meeting has that function returning true by force, so we didnt see the issue then

todor-ivanov: hi @LinaresToine  thanks for reminding me that ... I'll see what can be done.

amaltaro:
@todor-ivanov @LinaresToine is this PR still relevant? Perhaps it has all been integrated into https://github.com/dmwm/CMSKubernetes/pull/1466?

LinaresToine:
The changes in this PR have been implemented. I believe we can close this one.

PR #1462, opened 2024-04-13
Addition of the opportunistic resources in the initialization script (retrying to avoid false authors)

anpicci:
This PR is intended to leverage the inclusion of opportunistic resources in the WMAgent.secrets template in the initialization of WMAgents when deployed with containers. This relates to the https://github.com/dmwm/WMCore/issues/11722, and is closely linked to the https://github.com/dmwm/WMCore/pull/11966.

In `manage-common.sh`, the hardcoded loading and parsing of the WMAgent.secrets are replaced with a dynamical approach, with a dedicated section for managing the variables containing the information about the opportunistic sites.

In `init.sh`, a new function is introduced to populate the WMAgent opportunistic resource control. It parses the `RESOURCE_*` env variables introduced in the WMAgent.secrets with the dmwm/WMCore PR mentioned above, according to their structure. This needs to exploit `eval` to properly access the `RESOURCE_*` env vars and the information stored in them, which are the name of the site, the running slots, and the pending slots allocated as opportunistic resources.

NB:
I recreated the PR with a new branch to avoid false contributions to the PR

### DIFF
--- a/docker/pypi/wmagent/bin/manage-common.sh
+++ b/docker/pypi/wmagent/bin/manage-common.sh
@@ -368,17 +368,11 @@ _load_wmasecrets(){
     local secretsFile=${1:-$WMA_SECRETS_FILE}
     local varsToLoad=${2:-""}
 
-    # # Initial parsing of the $WMA_SECRETS_FILE
-    # _parse_wmasecrets $secretsFile || { echo "$FUNCNAME: WARNING: Not loading raw or not updated secrets file at $secretsFile"; return $(false) ;}
-
     [[ -f $secretsFile ]] || {
         echo "$FUNCNAME: ERROR: Password file: $secretsFile does not exist"
         echo "$FUNCNAME: ERROR: Either set WMA_SECRETS_FILE environment variable to a valid file or check that $HOME/WMAgent.secrets exists"
         return $(false)
     }
-
-    # # All variables need to be fetched in lowercase through: ${varName,,}
-    # local badValuesReg="(update-me|updateme|<update-me>|<updateme>|fix-me|fixme|<fix-me>|<fixme>|^$)"
 
     # If no list of variables to be loaded was given assume all of them.
     # Building the list by parsing the secrets file itself.
--- a/docker/pypi/wmagent/bin/manage-common.sh
+++ b/docker/pypi/wmagent/bin/manage-common.sh
@@ -346,117 +346,80 @@ _parse_wmasecrets(){
 # Passwords/Secrets handling
 #
 _load_wmasecrets(){
-    if [ "x$WMA_SECRETS_FILE" == "x" ]; then
-        WMA_SECRETS_FILE=$HOME/WMAgent.secrets;
-    fi
-    if [ ! -e $WMA_SECRETS_FILE ]; then
-        echo "$FUNCNAME: Password file: $WMA_SECRETS_FILE does not exist"
-        echo "$FUNCNAME: Either set WMA_SECRETS_FILE environment variable to a valid file or check that $HOME/WMAgent.secrets exists"
-        return 1;
-    fi
 
-    _parse_wmasecrets $WMA_SECRETS_FILE || { echo "$FUNCNAME: WARNING: Not loading raw or not updated secrets file at $WMA_SECRETS_FILE"; return $(false) ;}
+    # Auxiliary function to parse WMAgent.secrets or MariaDB.secrets files
+    # and load a set of variables from them
+    # :param $1: Path to WMAgent.secrets or file (Default: $WMA_SECRETS_FILE)
+    # :param $2: String with variable names to be checked (Default: *)
+    # :retrun:   Error value if one or more values have been left unset in the secrets file
+    local errVal=0
+    local value=""
+    local secretsFile=${1:-$WMA_SECRETS_FILE}
+    local varsToLoad=${2:-""}
 
-    local MATCH_ORACLE_USER=`cat $WMA_SECRETS_FILE | grep ORACLE_USER | sed s/ORACLE_USER=//`
-    local MATCH_ORACLE_PASS=`cat $WMA_SECRETS_FILE | grep ORACLE_PASS | sed s/ORACLE_PASS=//`
-    local MATCH_ORACLE_TNS=`cat $WMA_SECRETS_FILE | grep ORACLE_TNS | sed s/ORACLE_TNS=//`
-    local MATCH_GRAFANA_TOKEN=`cat $WMA_SECRETS_FILE | grep GRAFANA_TOKEN | sed s/GRAFANA_TOKEN=//`
-    local MATCH_MDB_USER=`cat $WMA_SECRETS_FILE | grep MDB_USER | sed s/MDB_USER=//`
-    local MATCH_MDB_PASS=`cat $WMA_SECRETS_FILE | grep MDB_PASS | sed s/MDB_PASS=//`
-    local MATCH_MDB_HOST=`cat $WMA_SECRETS_FILE | grep MDB_HOST | sed s/MDB_HOST=//`
-    local MATCH_COUCH_USER=`cat $WMA_SECRETS_FILE | grep COUCH_USER | sed s/COUCH_USER=//`
-    local MATCH_COUCH_PASS=`cat $WMA_SECRETS_FILE | grep COUCH_PASS | sed s/COUCH_PASS=//`
-    local MATCH_COUCH_PORT=`cat $WMA_SECRETS_FILE | grep COUCH_PORT | sed s/COUCH_PORT=//`
-    local MATCH_COUCH_HOST=`cat $WMA_SECRETS_FILE | grep COUCH_HOST | sed s/COUCH_HOST=//`
-    local MATCH_COUCH_CERT_FILE=`cat $WMA_SECRETS_FILE | grep COUCH_CERT_FILE | sed s/COUCH_CERT_FILE=//`
-    local MATCH_COUCH_KEY_FILE=`cat $WMA_SECRETS_FILE | grep COUCH_KEY_FILE | sed s/COUCH_KEY_FILE=//`
-    local MATCH_GLOBAL_WORKQUEUE_URL=`cat $WMA_SECRETS_FILE | grep GLOBAL_WORKQUEUE_URL | sed s/GLOBAL_WORKQUEUE_URL=//`
-    local MATCH_LOCAL_WORKQUEUE_DBNAME=`cat $WMA_SECRETS_FILE | grep LOCAL_WORKQUEUE_DBNAME | sed s/LOCAL_WORKQUEUE_DBNAME=//`
-    local MATCH_WORKLOAD_SUMMARY_URL=`cat $WMA_SECRETS_FILE | grep WORKLOAD_SUMMARY_URL | sed s/WORKLOAD_SUMMARY_URL=//`
-    local MATCH_WORKLOAD_SUMMARY_DBNAME=`cat $WMA_SECRETS_FILE | grep WORKLOAD_SUMMARY_DBNAME | sed s/WORKLOAD_SUMMARY_DBNAME=//`
-    local MATCH_WMSTATS_URL=`cat $WMA_SECRETS_FILE | grep WMSTATS_URL | sed s/WMSTATS_URL=//`
-    local MATCH_REQMGR2_URL=`cat $WMA_SECRETS_FILE | grep REQMGR2_URL | sed s/REQMGR2_URL=//`
-    local MATCH_ACDC_URL=`cat $WMA_SECRETS_FILE | grep ACDC_URL | sed s/ACDC_URL=//`
-    local MATCH_DBS3_URL=`cat $WMA_SECRETS_FILE | grep DBS3_URL | sed s/DBS3_URL=//`
-    local MATCH_DQM_URL=`cat $WMA_SECRETS_FILE | grep DQM_URL | sed s/DQM_URL=//`
-    local MATCH_REQUESTCOUCH_URL=`cat $WMA_SECRETS_FILE | grep REQUESTCOUCH_URL | sed s/REQUESTCOUCH_URL=//`
-    local MATCH_CENTRAL_LOGDB_URL=`cat $WMA_SECRETS_FILE | grep CENTRAL_LOGDB_URL | sed s/CENTRAL_LOGDB_URL=//`
-    local MATCH_WMARCHIVE_URL=`cat $WMA_SECRETS_FILE | grep WMARCHIVE_URL | sed s/WMARCHIVE_URL=//`
-    local MATCH_AMQ_CREDENTIALS=`cat $WMA_SECRETS_FILE | grep AMQ_CREDENTIALS | sed s/AMQ_CREDENTIALS=//`
-    local MATCH_RUCIO_HOST=`cat $WMA_SECRETS_FILE | grep RUCIO_HOST | sed s/RUCIO_HOST=//`
-    local MATCH_RUCIO_AUTH=`cat $WMA_SECRETS_FILE | grep RUCIO_AUTH | sed s/RUCIO_AUTH=//`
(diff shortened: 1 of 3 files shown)


anpicci:
I have tried to run the following test.

After successfully initiating the agent, I changed a parameter in WMAgent.secrets for one of the opportunistic sites, and I got this message when re-initiating the agent with `./wmagent-docker-run.sh -t 2.3.0 && docker logs -f wmagent`

`
Start: Performing agent_resource_opp
agent_resource_opp: triggered.
agent_resource_opp: Populating opportunistic resource-control
I am based at CERN, so I cannot use US opportunistic resources, moving to the next site
execute_command_agent: Executing: wmagent-resource-control --plugin=SimpleCondorPlugin --opportunistic --pending-slots=700 --running-slots=5000 --add-one-site=T3_ES_PIC_BSC ...
It's not possible to change a site's SE name after the site has
been added to the database.
ERROR: Failed to populate WMAgent's opportunistic resource control!
`

This would happen also with the other sites, and the only way to solve is to clean the Oracle database within the agent, kill it, and re-initiate it. Discussing with @todor-ivanov, this is due to the immutable nature of the agent resource control. It would be good to move a mutable resource control model, also taking into account that, with the new deployment model, patching the agent would mean creating a new patch tag and rebuilding a new docker agent image.

For the rest, this development works, so I don't know if we want to consider it closed and work on the resource control model afterward (after the revision), or first work on the resource control model and then close it.

@amaltaro what do you think?

docker/pypi/wmagent/bin/manage-common.sh
```diff
@@ -346,150 +357,100 @@ _parse_wmasecrets(){
...
+    # :param $2: String with variable names to be checked (Default: *)
+    # :retrun:   Error value if one or more values have been left unset in the secrets file
+    local errVal=0
+    local value=""
+    local secretsFile=${1:-$WMA_SECRETS_FILE}
+    local varsToLoad=${2:-""}
 
-    LOCAL_WORKQUEUE_DBNAME=${MATCH_LOCAL_WORKQUEUE_DBNAME:-$LOCAL_WORKQUEUE_DBNAME};
+    # # Initial parsing of the $WMA_SECRETS_FILE
+    # _parse_wmasecrets $secretsFile || { echo "$FUNCNAME: WARNING: Not loading raw or not updated secrets file at $secretsFile"; return $(false) ;}
```
todor-ivanov, line 372: hi @anpicci  this line was a leftover from a previous implemention.  Please remove it.

docker/pypi/wmagent/bin/manage-common.sh
```diff
@@ -346,150 +357,100 @@ _parse_wmasecrets(){
...
-    WORKLOAD_SUMMARY_URL=${MATCH_WORKLOAD_SUMMARY_URL:-$WORKLOAD_SUMMARY_URL};
+    [[ -f $secretsFile ]] || {
+        echo "$FUNCNAME: ERROR: Password file: $secretsFile does not exist"
+        echo "$FUNCNAME: ERROR: Either set WMA_SECRETS_FILE environment variable to a valid file or check that $HOME/WMAgent.secrets exists"
+        return $(false)
+    }
 
-    WMSTATS_URL=${MATCH_WMSTATS_URL:-$WMSTATS_URL}
+    # # All variables need to be fetched in lowercase through: ${varName,,}
+    # local badValuesReg="(update-me|updateme|<update-me>|<updateme>|fix-me|fixme|<fix-me>|<fixme>|^$)"
```
todor-ivanov, line 381: These two line were also a leftover from a previous implementation. Please remove them

todor-ivanov (commented):
Hi @anpicci  I left two comments inline for some minor stuff. You may want to take a look

docker/pypi/wmagent/bin/manage-common.sh
```diff
@@ -326,17 +328,26 @@ _renew_proxy(){
 
 _parse_wmasecrets(){
     # Auxiliary function to provide basic parsing of the WMAgent.secrets file
-    # :param $1: path to WMAgent.secrets file
+    # :param $1: path to WMAgent.secrets file (Default: $WMA_SECRETS_FILE)
+    # :param $2: a particular valuw to check (Default: *)
```
amaltaro, line 332: typo

docker/pypi/wmagent/bin/manage-common.sh
```diff
@@ -346,150 +357,94 @@ _parse_wmasecrets(){
...
-    local MATCH_AGENT_NUMBER=`cat $WMA_SECRETS_FILE | grep AGENT_NUMBER | sed s/AGENT_NUMBER=//`
-
-
-    # database settings (mysql or oracle)
-    if [ "x$MATCH_ORACLE_USER" == "x" ]; then
+    # Auxiliary function to parse WMAgent.secrets or MariaDB.secrets files
+    # and load a set of variables from them
+    # :param $1: Path to WMAgent.secrets or file (Default: $WMA_SECRETS_FILE)
+    # :param $2: String with variable names to be checked (Default: *)
+    # :retrun:   Error value if one or more values have been left unset in the secrets file
```
amaltaro, line 365: typo

docker/pypi/wmagent/bin/manage-common.sh
```diff
@@ -346,150 +357,94 @@ _parse_wmasecrets(){
...
+    local errVal=0
+    local value=""
+    local secretsFile=${1:-$WMA_SECRETS_FILE}
+    local varsToLoad=${2:-""}
+
+    [[ -f $secretsFile ]] || {
+        echo "$FUNCNAME: ERROR: Password file: $secretsFile does not exist"
```
amaltaro, line 372: I would rephrase it to "ERROR: Secrets file..."

docker/pypi/wmagent/bin/manage-common.sh
```diff
@@ -346,150 +357,94 @@ _parse_wmasecrets(){
...
+        }
+    done
+
+    # Now load them all
+    for varName in $varsToLoad
+    do
+        value=`grep -E "^[[:blank:]]*$varName=" $secretsFile | sed "s/ *$varName=//"`
+        [[ $varName =~ ^RESOURCE_ ]] && declare -g -A $varName
+        eval $varName=$value
+        [[ -n $varName ]] || { echo "$FUNCNAME: ERROR: Empty value for: $varName=$value"; let errVal+=1 ;}
```
amaltaro, line 399: Given that we are already validating the secrets parameter a few lines above:
https://github.com/dmwm/CMSKubernetes/pull/1462/files#diff-01c5d9be01b6e6ea0ee4b12d5262b4a641bc3bdc55a79853ca90abb819996f96R386

isn't this L399 redundant?

docker/pypi/wmagent/init.sh
```diff
@@ -425,6 +427,39 @@ agent_resource_control() {
...
+    if _init_valid $wmaInitResourceOpp && \
+       _init_valid $wmaInitSqlDB
+    then
+        echo "$FUNCNAME: Agent Opportunistic Resource control has been populated already."
+    else
+        echo "$FUNCNAME: triggered."
+        local errVal=0
+        ### Populating opportunistic resource-control
+        echo "$FUNCNAME: Populating opportunistic resource-control"
+        for res in ${!RESOURCE_*}
```
amaltaro, line 443: Can you please add a comment above this line saying what is expected to be in RESOURCE_* env variables?

In addition, I am confused with the exclamation mark in this line. What is actually happening here? It iterates over all RESOURCE_* env vars that are not empty?

amaltaro (changes requested):
Thank you for these changes, Andrea and Todor. I left a few comments along the code.

In addition, can you please enhance the PR description with other relevant changes that have been provided with this PR? Such as:
* which issue it's meant to fix
* the overhaul of secrets parsing
* a new init file for opportunistic resources

docker/pypi/wmagent/bin/manage-common.sh
todor-ivanov: no, it is not
because 
* between the moment when you set the value and the moment when you "initially" parse and validate the the whole secrets file  there are  few steps including an extra declaration which may cause an error in the `eval` step due to an obscured value put for those `RESOURCE_*`  (which we  cannot predict what it could be) 
* This is also a precaution, because at some point, if in the future we decide to add  yet another functionality to be parsed, and we work only in the `_parse_wmasecrets` file (s it should be), but we somehow make a mistake and create a bug, we will be extra protected from this line here to avoid empty values loaded.

anpicci: I would leave this as it is

docker/pypi/wmagent/init.sh
anpicci: Yes, and "!" is needed to actually access the information stored in `RESOURCE_*` variables as `res`

anpicci:
@amaltaro I have committed my changes addressing your comments

docker/pypi/wmagent/bin/manage-common.sh
amaltaro: Yes, I don't think it's worth it to change this PR only because of that line.

I still fail to understand the logic for checking these variables. Isn't this line:
```
[[ ${value,,} =~ $badValuesReg ]] && { echo "$FUNCNAME: ERROR: Bad value for: $var=$value"; let errVal+=1 ;}
```
from `_parse_wmasecrets` already catching the case where a variable is left empty?

I am not sure I understand your first bullet. Are you saying that someone might mistype one of the RESOURCE_ variables, which would then fail the `        eval $varName=$value` command?

amaltaro:
Other than the follow up question, these changes are looking good to me.
Andrea, whenever you make changes and need follow up review, please trigger again a review request on the top right corner.

docker/pypi/wmagent/bin/manage-common.sh
todor-ivanov: hi @amaltaro 
> I am not sure I understand your first bullet. Are you saying that someone might mistype one of the RESOURCE_ variables, which would then fail the         eval $varName=$value command?

You got it correctly. With one minor detail. The value may not mandatory be mistyped, but it could be someone actually tried to configure those resources in the form of a regular array, rather than an associative array. e.g. `RESOURCE_OPP3=("T3_US_UCSD" 3000  5000 "normal")` instead of `RESOURCE_OPP3=([name]="T3_US_UCSD" [run]=3000  [pend]=5000 [state]="normal")` This will result in the following error in the evaluation command:

```
:~$ declare -A RESOURCE_OPP3
:~$ RESOURCE_OPP3=("T3_US_UCSD" 3000  5000 "normal")
-bash: RESOURCE_OPP3: "T3_US_UCSD": must use subscript when assigning associative array
-bash: RESOURCE_OPP3: 3000: must use subscript when assigning associative array
-bash: RESOURCE_OPP3: 5000: must use subscript when assigning associative array
-bash: RESOURCE_OPP3: "normal": must use subscript when assigning associative array
:~$ echo ${RESOURCE_OPP3[*]}

```   

Which results in an empty value. While the expected behavior is:

```
:~$ declare -A RESOURCE_OPP3
:~$ RESOURCE_OPP3=([name]="T3_US_UCSD" [run]=3000  [pend]=5000 [state]="normal")
:~$ echo ${RESOURCE_OPP3[*]}
normal 3000 T3_US_UCSD 5000
```
Since the `_parse_wmasecrets` function does not have a section for detailed validation of value  types other than the generic value presence and bad regular expression... this cannot be caught by this function. Hence the presence of the extra check.

amaltaro:
Thank you for the clarification, Andrea and Todor. These changes look good to me!

todor-ivanov:
thanks for approving this @amaltaro , 

But before we merge ....... I hope you did not miss that exact  comment from @anpicci: https://github.com/dmwm/CMSKubernetes/pull/1462#issuecomment-2056033742 

This is an important behavior change here. And let me give you a short explanation how it changes, so we can discuss it:
 **Current behavior:**
 * Until now we were always running with an immutable resource control, which means once  a resource is added  at  initialization time it stays as it is. 
 * We set all the opportunistic resources with a identical resource values and we distinguish only between resources  suitable for FNAL and CERN agents.
 
 **New behavior:**
* Once the agent goes through the init process we set the Opportunistic resources based on the resource configurations put in the WMAgent.secrets file
* We set all the opportunistic resources with its own set of resource values and again preserve the geographical separation as before.
* BUT from now one  if somebody decides  to change a resource configuration he may go and reshape the values  in the WMAgent.secrets file, and restart the wmagent container. What will happen from there on is as follows: 
   * The agent init script will find out that WMASecrets file has a different checksum - meaning somebody has reconfigurred something
   * The init script will repeat the checks for relational database ... and some more.... regenerate wmagent config etc.
   * The init script will also try to reconfigure the Oportunistic resources. And there is  the moment when you observe the Error @anpicci  is mentioning. 

Lets talk ;)

amaltaro:
> BUT from now one if somebody decides to change a resource configuration he may go and reshape the values in the WMAgent.secrets file, and restart the wmagent container. What will happen from there on is as follows:
>
>    The agent init script will find out that WMASecrets file has a different checksum - meaning somebody has reconfigurred something
>    The init script will repeat the checks for relational database ... and some more.... regenerate wmagent config etc.
>    The init script will also try to reconfigure the Oportunistic resources. And there is the moment when you observe the Error @anpicci is mentioning.

I see. Even though we rarely change these resources - and we could as well only change it when upgrading agents - this is really not a great behavior to have.
I have a suggestion to avoid erroring on addition of opportunistic resources, which would be a SELECT statement to check whether the site has already been added to the resource-control, and only add sites not yet inserted. 

It still does not avoid the whole initialization and checks of the system.

todor-ivanov:
To avoid completely re-configuring an already existing site is one way to go.... but now imagine the mid agent lifetime we get the news of a lot  bigger allocation granted to an already existing Resource....  how do we react in order to consume this?

amaltaro:
The same ways we do nowadays, by increasing the number of slots for that given site in SSB.

todor-ivanov:
so if we have those opportunistic resources in SSB such that we can set the values there .... why do we need to have this semi-manual intervention for adding them to the resource control like this for ages ..... Am I missing something here .....where is the missing bit?

amaltaro:
Site and storage names come from CRIC. Given that these are opportunistic resources without storage, they are not available in CRIC.
Slot information is required to creating a new resource in the database, so we need to have reasonable starting values and - IF - the site is properly configured in SSB (site name, slots for production, status for production, etc), the agent is capable of managing the site status through AgentStatusWatcher.

At some point we should revisit these opportunistic sites and/or adopt a different CRIC API to automatically populate everything. One of the reasons I've taken the easy path years ago :)

todor-ivanov:
Thanks @amaltaro,  We will have to revisit all of this. I am willing to work on this. Even though not in this issue here. May I ask you to create a HG populating as much of this information as  you can, and I'll spend some extra time on my own to look at it and give it a thought aside of the current development process. We may also discuss it and later properly include it in the development cycle.... I am aware we are not doing it the optimal way ... and I wanted to have this revisiting process triggered with this very issue here..  but obviously, with great sorrow in my heart, I'll must accept the fact that we will have to take the easy path yet again. 

@anpicci  could you just remove the line https://github.com/dmwm/CMSKubernetes/pull/1462/files#diff-609d980355cae845676b5d5f8859fa63cb459b6cf154dce468690a820fa23f4dR194 : 
```
       rm -f $wmaInitResourceOpp
```
 before we merge, which is responsible for triggering the reconfiguration of the opportunistic resources. Thanks!

anpicci:
@todor-ivanov sure! Only to double-check, is it ok if, for the time being, we keep the implementation of the opportunistic resources as proposed in this PR? @amaltaro

anpicci:
@todor-ivanov done

todor-ivanov (approved):
Thanks @anpicci , looks good to me!

amaltaro (approved):
Thanks! It looks good to me as well.

amaltaro:
I am squashing and merging these changes in. Thanks Todor and Andrea!

anpicci:
Thanks to you @amaltaro @todor-ivanov !

PR #1464, opened 2024-04-17
Add UID:GID options to docker options, remove user/group from Dockerfile

vkuznet:
Fixes: https://github.com/dmwm/WMCore/issues/11944

This PR contains changes to pypi Dockerfiles and run scripts to use UID:GID at run-time, via the following docker options (here is an example of using alpine image and passing specific UID:GID along with group, passwd files to docker run-time environment):
```
$ docker run --rm
    --user $UID:$GID
    --workdir="/home/$USER"
    --volume="/etc/group:/etc/group:ro"
    --volume="/etc/passwd:/etc/passwd:ro"
    --volume="/etc/shadow:/etc/shadow:ro"
    alpine ash -c "whoami"
```


docker/pypi/wmagent-couchdb/Dockerfile
```diff
@@ -80,7 +57,7 @@ ENV PATH="/opt/couchdb/bin:/usr/local/bin/:${PATH}"
 ENV CRYPTOGRAPHY_ALLOW_OPENSSL_102=true
 
 # RUN <<EOF cat >> /opt/couchdb/.bashrc
-RUN <<EOF cat >> /home/$USER/.bashrc
+RUN <<EOF cat >> $HOME/.bashrc
```
todor-ivanov, line 60: this I think will break in case somebody tries to build with an user other than the one, who is about to later run the container. Maybe escaping the `\$` sign before Home may help ... but I am still skeptical.....

It is really important, to my opinion, to have the command prompt correctly set (for both regular and root users inside the container). The best way how it was done was in the MariaDB container.

https://github.com/dmwm/CMSKubernetes/blob/326de33e3769e2f66e088c91093d3179a8d1cc12/docker/pypi/wmagent-mariadb/Dockerfile#L55-L64 

This must be set to all of the 3 containers the same way as  it was done there. And to cover both `regular user` and `root user` command prompts.

I do not know how this would work now. I did waste quite a lot of time in the near past to implement it in a generic way and to put it in `/etc/profile`, this way we will not depend on the $HOME valu to be resolved  at build time, but I have lost the change. Let me try to find it again, and come back to you with a concrete suggestion.

docker/pypi/wmagent-couchdb/couchdb-docker-run.sh
```diff
@@ -69,6 +69,11 @@ dockerOpts="
 --detach \
 --network=host \
 --rm \
+--user $UID:$GID \
+--volume='/etc/group:/etc/group:ro' \
+--volume='/etc/passwd:/etc/passwd:ro' \
+--volume='/etc/shadow:/etc/shadow:ro' \
+--volume='/etc/sudoers:/etc/sudoers:ro' \
```
todor-ivanov, line 76: maybe we should also bind mount `/etc/sudoers.d`   because especially in the CERN machines, there are the actual lists of `sudo`  users (puppet managed of course)

docker/pypi/wmagent-mariadb/Dockerfile
```diff
@@ -63,9 +59,6 @@ alias manage=$MDB_MANAGE_DIR/manage
 export PS1="(MariaDB-$MDB_TAG) [\u@\h:\W]\$([[ \$(id -u) -eq 0 ]] && echo \# || echo \$) "
 EOF
```
todor-ivanov, line 60: look at my comment referring the respective lines at the CouchDB container.

todor-ivanov (commented):
Thanks @vkuznet  I left few comments inline and I am about to make direct suggestion in a while for fixing 2 of them, based on my previous experience. Just stay put for a while. Other than that I find this solution as incredibly good.

docker/pypi/wmagent-couchdb/couchdb-docker-run.sh
vkuznet: ok, that's fine, added

docker/pypi/wmagent-couchdb/Dockerfile
vkuznet: ok, done

docker/pypi/wmagent-mariadb/Dockerfile
vkuznet: Todor, I resolved previous comments, but I'm not sure about this one, please clarify it since I didn't touch mariadb prompt, and only added CouchDB one.

todor-ivanov:
@vkuznet  two more comments.

* since we do not create docker volumes the so added docker runtime options fail with:
```
cmst1@vocms0290:wmagent-mariadb $ ./mariadb-docker-run.sh -t 10.6.5
Starting the local/mariadb:10.6.5 docker container with the following parameters:  --user cmst1 -e USER=cmst1
docker: Error response from daemon: invalid mode: ro'.
See 'docker run --help'.
```

 **This is fixable** by downgrading again to bind mounts like this: 

```
--mount type=bind,source=/etc/group,target=/etc/group,readonly \
--mount type=bind,source=/etc/passwd,target=/etc/passwd,readonly \
--mount type=bind,source=/etc/shadow,target=/etc/shadow,readonly \
--mount type=bind,source=/etc/sudoers,target=/etc/sudoers,readonly \
--mount type=bind,source=/etc/sudoers.d,target=/etc/sudoers.d,readonly \
--mount type=bind,source=/etc/profile,target=/etc/profile,readonly \
```

* but we fail yet again .... badly...., in the next stage, when we realize that the users we are running with (especially at CERN) are not coming from `/etc/passowrds` by from `ldap` : 

```
cmst1@vocms0290:wmagent-mariadb $ ./mariadb-docker-run.sh -t 10.6.5
Starting the local/mariadb:10.6.5 docker container with the following parameters:  --user cmst1 -e USER=cmst1
31a5fca1a3f24e8dbc7f209f36891984004d079b145317e1c5a075105edbe351
docker: Error response from daemon: unable to find user cmst1: no matching entries in passwd file.
```
and indeed the user is missing there: 
```
cmst1@vocms0290:wmagent-mariadb $ grep cmst1 /etc/passwd

```

amaltaro:
I was reading about the `--mount` and `--volume` options and the [documentation](https://docs.docker.com/storage/bind-mounts/#choose-the--v-or---mount-flag) suggests going with `--mount`. I didn't see if there is anything provided by the volume option which is not by the mount one(?)
 
>    but we fail yet again .... badly...., in the next stage, when we realize that the users we are running with (especially at CERN) are not coming from /etc/passowrds by from ldap :

This is bad! We do have at least one alternative though, which would be to puppetize these two files /etc/passwd and /etc/group such that custom - for the WMAgent containers - files are provided (such that we have this ldap user in).
Farrukh mentioned they have some similar setup at FNAL and he is supportive of this idea at the FNAL schedds, if needed.

Do you see different approaches to this?

vkuznet:
Todor, I added mount points you provided.

Alan, I think it is the only solution we have, i.e. puppetize files /etc/passwd and /etc/group to include proper user/group we need.

todor-ivanov:
hi @amaltaro 

let me stand myself correctd .... we are not using ldap (at least not directly) we are going through sssd etc.... long path to the user details...... 
Then: 
> We do have at least one alternative though, which would be to puppetize these two files /etc/passwd and /etc/group such that custom - for the WMAgent containers - files are provided (such that we have this ldap user in). 

This is the reeeeaaaallllyyy last resort.... We should not involve puppet in any way here. 

But long story short: 
> Do you see different approaches to this?

Yes, we can obtain all users  details at runtime. I am looking into something here. Just a heads up. Go in one of the agents and query for a singel user all the user databases configured at the current machine's in `nsswitch`. This is simple:

`getent passwd <username>` 

I am looking deeper into this rabbit  whole.

todor-ivanov:
@amaltaro  @vkuznet 

And something more .... 
> getent passwd <username>
>I am looking deeper into this rabbit whole.

If you go and look at  the output fro few users we usually use you will see their home to be in `afs` ... so we will have to work arround this one as well.

vkuznet:
So, how do we need to proceed with this issue? In my view, if there is no way to pass user id/groups to the container we can't use UID:GID at run time because those must be available within run-time environment of the container. And, we can't hard-code them during docker image creation due to security concern of FNAL team. I don't know any mechanism how to pass afs user names/ids to the container unless they are in /etc/passwd. The only way seems to reproduce entire afs setup within docker image and bring in a similar fashion the afs groups/users. But this seems to be overkill. Any other ideas/thoughts?

docker/pypi/wmagent-mariadb/mariadb-docker-run.sh
```diff
@@ -69,6 +69,7 @@ dockerOpts="
 --detach \
 --network=host \
 --rm \
+--user $UID:$GID \
```
todor-ivanov, line 72: @vkuznet  just wanted to mention, since we are using bash as  default shell in our agents, the `$GUID` env variable is empty. The best possible way to get the group id would be `id -g`.

vkuznet: Todor, I applied your suggestion.

docker/pypi/wmagent/wmagent-docker-run.sh
```diff
@@ -55,7 +58,19 @@ HOST_MOUNT_DIR=/data/dockerMount
 [[ -h /data/srv/wmagent ]] && rm -f /data/srv/wmagent
 ln -s $HOST_MOUNT_DIR/srv/wmagent /data/srv/wmagent
 
+# create the passwd and group mount point dynamically at runtime
+passwdEntry=`getent passwd $wmaUser  |awk -F : -v wmaUser=$wmaUser '{print $1":"$2":"$3":"$4":"$5":""/home/"wmaUser":"$7}'`
+groupEntry=`getent group $wmaGroup`
 
+[[ -d $HOST_MOUNT_DIR/admin/ ]] || (mkdir -p $HOST_MOUNT_DIR/admin) || exit $?
+
+getent passwd > $HOST_MOUNT_DIR/admin/passwd
```
todor-ivanov, line 67: Ok @vkuznet , I made a minor mistake here. I just found  it myself. Here the 3 containers would overlap and would overwrite their files one another..... so we should have this line for the 3 container flavours in  3 different configurations ( with the respective sources for the mount points bellow):

```
[[ -d $HOST_MOUNT_DIR/admin/wmagent/etc ]] || (mkdir -p $HOST_MOUNT_DIR/admin/wmagent/etc) || exit $?
getent passwd > $HOST_MOUNT_DIR/admin/wmagent/etc/passwd
```

todor-ivanov: Nope, exactly the opposite: 

We should recreate the them 3 times in the 3 different hostmount  areas:

* docker/pypi/wmagent-couchdb/couchdb-docker-run.sh
```
[[ -d $HOST_MOUNT_DIR/admin/couchdb/etc ]] || (mkdir -p $HOST_MOUNT_DIR/admin/couchdb/etc) || exit $?
getent passwd > $HOST_MOUNT_DIR/admin/couchdb/etc/passwd
...
--mount type=bind,source=$HOST_MOUNT_DIR/admin/couchdb/etc/passwd,target=/etc/passwd,readonly \
```
* docker/pypi/wmagent-mariadb/mariadb-docker-run.sh
```
[[ -d $HOST_MOUNT_DIR/admin/mariadb/etc ]] || (mkdir -p $HOST_MOUNT_DIR/admin/mariadb/etc) || exit $?
getent passwd > $HOST_MOUNT_DIR/admin/mariadb/etc/passwd
...
--mount type=bind,source=$HOST_MOUNT_DIR/admin/mariadb/etc/passwd,target=/etc/passwd,readonly \
```
* docker/pypi/wmagent/wmagent-docker-run.sh

```
[[ -d $HOST_MOUNT_DIR/admin/wmagent/etc ]] || (mkdir -p $HOST_MOUNT_DIR/admin/wmagent/etc) || exit $?
getent passwd > $HOST_MOUNT_DIR/admin/wmagent/etc/passwd
...
--mount type=bind,source=$HOST_MOUNT_DIR/admin/wmagent/etc/passwd,target=/etc/passwd,readonly \
```


But I am not sure if we are not to take the other path @amaltaro  was  suggesting here: https://github.com/dmwm/CMSKubernetes/pull/1468 which should in general make this step unnneded

amaltaro:
@vkuznet Valentin, everything that is provided in this PR has also been incorporated to https://github.com/dmwm/CMSKubernetes/pull/1469. It also has a bunch of other changes on top of that, but I don't think it's in a final state, as moving the user from the image to the container opened up other ownership/permissions issues.

I think we can follow this development up in https://github.com/dmwm/CMSKubernetes/pull/1469 and close this one. What do you think?

PR #1466, opened 2024-04-18
Adding init_tier0 function

todor-ivanov:
Fixes https://github.com/dmwm/WMCore/issues/11890

With this PR we try to address the differences  in the agent initialization between production and tier0 agents.

* Avoid checks of mount points for Tier0 agents 
* Build the initial agent configuration template by merging the T0 config template  + WMA config template
* Add a separate agent initialization function for T0 including options relevant for T0 and excluding those speciffic to production agents
* Parsing and adding additional (T0 specific) variables from the WMAgent.secrets
* Add the `t0-mod-config`  step for modifying the temporarily produced  WMAgent configuration file to remove all config sections irrelevant to T0
* Prevent moving runtime files from the deployment to a non existing mount point, which was needed  only for the docker model, where HTCondor was not able to reach the runtime files and the workflow sandbox during submission
* Add a T0 specific `_renew_proxy` method (to be called through `manage renew-proxy`)
* Fixing few bugs left unnoticed in parsing `WMAgent.secrets` file, which were breaking the support of `#` comments in the file
* Avoid uploading agent configuration to central services  for Tier0 agents
* Make `agent_resource_opp` more universal - stop translating option names in the code, instead use a direct map of configuration key to option between the so configured resource record in the `WMAgent.secrets` file and set of options recognized by the `wmagent-resource-control` script: 

https://github.com/dmwm/WMCore/blob/e4628ce9ab25f5015b71197a42eb09d630286318/bin/wmagent-resource-control#L48-L96

This is needed for both 
* avoiding extra hard coded variables in the parsing process and making the code more agnostic to the contents of the configuration/secrets files and also for 
* reusing the same function to implement the initialization step for populating the resource control for Tier0 agents: https://github.com/dmwm/T0/blob/97f9cded77b0d91c8f77b9949a567c7ea4883d6f/bin/00_deploy_replay.sh#L226-L251 
This step needs to be able to add  resource parameters by request type. 

So, from now on, the key names in the `WMAgent.secrets` file should be a one to one map of the parameter names recognized by `wmagent-resource-control` like this:
```
RESOURCE_OPP1=([add-one-site]=T3_US_NERSC   [pending-slots]=3000 [running-slots]=750 [plugin]=SimpleCondorPlugin [opportunistic]=true [normal]=true)
```
And respectively for T0 agents, like this:
```
RESOURCE_T2CERN=([site-name]=T2_CH_CERN [cms-name]=T2_CH_CERN [pnn]=T0_CH_CERN_Disk [ce-name]=T2_CH_CERN [pending-slots]=20000 [running-slots]=20000 [plugin]=SimpleCondorPlugin)

RESOURCE_T2CERN_Repack=([site-name]=T2_CH_CERN [task-type]=Repack [pending-slots]=5000 [running-slots]=5000)
```


docker/pypi/wmagent/bin/manage
```diff
@@ -158,10 +163,69 @@ init_wmagent(){
...
+
+# generate the tier0 config from the basic template
+init_tier0(){
+    local errVal=0
+    wmagent-mod-config --input=$WMA_CONFIG_DIR/config-template.py \
+                       --output=$WMA_CONFIG_DIR/config-agent.py \
+                       --coredb_url=oracle://$ORACLE_USER:$ORACLE_PASS@$ORACLE_TNS \
+                       --working_dir=$WMA_INSTALL_DIR \
+                       --couch_url=http://$COUCH_USER:$COUCH_PASS@$COUCH_HOST:$COUCH_PORT \
+                       --workqueue_db_name=$LOCAL_WORKQUEUE_DBNAME \
```
anpicci, line 179: For my education, some of these env vars are extracted by WMAgent secrets, right?

docker/pypi/wmagent/bin/manage
```diff
@@ -158,10 +163,69 @@ init_wmagent(){
...
+    export WMAGENT_CONFIG=$WMA_CONFIG_DIR/config.py
+    wmagent-couchapp-init;
+    let errVal+=$?
+    unset WMAGENT_CONFIG
+}
+
+
 init_agent(){
     if _init_valid $wmaInitActive ; then
         echo "Initialising Agent..."
```
anpicci, line 223: Could this echo line be moved in the if-else statement, together with the specification of what type of agent is going to be initialized? Only to enforce the confidence that an operator has about deploying the right agent

docker/pypi/wmagent/init.sh
```diff
@@ -115,9 +116,15 @@ deploy_to_host(){
     echo "-------------------------------------------------------"
     echo "Start: $stepMsg"
 
-    # TODO: This is to be removed once we decide to run it directly from the deploy area
-    echo "$FUNCNAME: Copy the proper manage file"
-    cp -fv $WMA_DEPLOY_DIR/bin/manage $WMA_MANAGE_DIR/manage && chmod 755 $WMA_MANAGE_DIR/manage
+    # TODO: This is to be removed once we decide to run it only from the deploy area
```
anpicci, line 119: Is there already an issue concerning this? And if not, would a new issue be needed for future reference?

anpicci (commented):
@todor-ivanov thank you for your effort! I have left only some minor comments along the scripts

docker/pypi/wmagent/bin/manage
todor-ivanov: Anything starting with `WMA_*`  is coming from the environment, the rest from  WMAgent.secrets

docker/pypi/wmagent/init.sh
todor-ivanov: This comment concerns only the creation of the soft link pointing to the correct `manage` script from `$WMA_CONFIG_DIR`, and as long as we use the rpm based agents, we cannot remove it, because it would require changing the  `$WMA_DEPLOY_DIR/bin/clean-oracle`, which would break backwards compatibility.

docker/pypi/wmagent/bin/manage
todor-ivanov: good point

docker/pypi/wmagent/bin/manage
todor-ivanov: And the origin of the full set of variables used in the current script is explained here: https://github.com/dmwm/CMSKubernetes/blob/e2d764a0a4c970b3ceff58f913e9bdee10328b25/docker/pypi/wmagent/bin/manage#L11-L58

docker/pypi/wmagent/init.sh
anpicci: ah ok, thank you for the clarification!

anpicci:
For me, everything looks ok now (I'm not sure if @todor-ivanov needs to require a new revision or not)

docker/pypi/wmagent/bin/manage
```diff
@@ -74,7 +74,12 @@ activate_agent(){
     # The very first step of activating the WMAgent upon all needed checks and
     # validations performed during the initialization process.
     # No .init* flags are checked at this stage.
-    cp -vf $WMA_DEPLOY_DIR/etc/WMAgentConfig.py $WMA_CONFIG_DIR/config-template.py
+    if [[ "$TEAMNAME" == Tier0* ]]; then
```
amaltaro, line 77: Instead of hardcoding a team, how about checking if `$WMA_DEPLOY_DIR/etc/Tier0Config.py` exists?

docker/pypi/wmagent/bin/manage
```diff
@@ -158,10 +163,70 @@ init_wmagent(){
...
+                       --workload_summary_url=$WORKLOAD_SUMMARY_URL \
+                       --grafana_token=$GRAFANA_TOKEN \
+                       --wmstats_url=$WMSTATS_URL \
+                       --dbs3_url=$DBS3_URL \
+                       --dqm_url=$DQM_URL \
+                       --requestcouch_url=$REQUESTCOUCH_URL \
+                       --central_logdb_url=$CENTRAL_LOGDB_URL \
+                       --wmarchive_url=$WMARCHIVE_URL \
+                       --amq_credentials=$AMQ_CREDENTIALS \
+	                   --rucio_account=$RUCIO_ACCOUNT \
```
amaltaro, line 189: Given that you are rewriting this, please fix the indentation of these lines

amaltaro (commented):
I really wish we could separate Tier0 from the WMAgent deployment, but I guess that would require a bunch of code duplication, wouldn't it?

I left a few comments along the code, and in general, I would suggest a different approach for detecting if the agent is for production or Tier0. Also, the concept of team name is blurred for T0. Perhaps we create a new environment variable to be used in the T0 system (in the secrets) to tag T0 agents?

In addition, please provide some details in the PR description. It will help us in the near future to understand why and where things were done.

docker/pypi/wmagent/bin/manage
todor-ivanov: hi @amaltaro  The `Teamname` has been discussed with T0 team and is a valid  criteria for the purpose of distinguishing the production agents  from the Tier0 ones, exactly the same way how we distinguish the `testbed` from `prod` agents.  The `Tier0Config.py` though is not such a reliable criteria mostly because it get installed quite  late in the process, and it comes with the T0 package.

todor-ivanov:
Hi @amaltaro , please take another look at the code changes here. 
I have addressed all your comments. Plus I have added a full PR description detailing every change. I think we may consider the changes  ready for final review. all of them have been tested and are working producing the expected result in deploying the agent. Both for the T0 agents and for the Production development environment if somebody wants too test it. 

There were a set of workflows injected in thee production system in testbed in order to test  this setup, and all of them finished successfully: 
* https://cmsweb-testbed.cern.ch/reqmgr2/fetch?rid=tivanov_SC_MultiPU_WMAgentVirtEnvTest_v11_240510_124352_6357
* https://cmsweb-testbed.cern.ch/reqmgr2/fetch?rid=tivanov_ReReco_RunBlockWhite_WMAgentVirtEnvTest_v11_240510_124338_6941
* https://cmsweb-testbed.cern.ch/reqmgr2/fetch?rid=tivanov_ReReco_Run2022C_LumiMask_WMAgentVirtEnvTest_v11_240510_124345_8233

The last step of tests needed  is yet to be confirmed - with a full validation replay in one of the T0 agents. This we plan to finish by the end of the day with T0 team. The issues we have met last time with that were  mostly T0 code speciffic, so I am confident we will be able to close that issue with our next test and no further deployment code changes  would be needed. So feel free  to start a final review of yours. I'll keep you posted about the results from these  tests

todor-ivanov:
Hi @anpicci @amaltaro I have rebased this PR on top of  our latest bugfixes from last week So it is ok for you to do  your final review on that.

docker/pypi/wmagent/init.sh
```diff
@@ -521,20 +552,20 @@ agent_resource_opp() {
 
 agent_upload_config(){
     # A function to be used for uploading WMAgentConfig to AuxDB at Central CouchDB
-    # NOTE: The final config/.dockerInit is to be set here after full agent initialisation.
+    # NOTE: The final config/.initUsing is to be set in the main() function after full agent initialization.
     #       Once we have reached to the step of uploading the agent config to Cerntral CouchDB
-    #       we consider all previous reintialisation and config steps as successfully complpeted.
+    #       we consider all previous re-internationalist and config steps as successfully completed.
```
amaltaro, line 557: re-internationalist?

docker/pypi/wmagent/init.sh
```diff
@@ -546,6 +577,10 @@ agent_upload_config(){
         elif [[ "$TEAMNAME" == *dev* ]]; then
             echo "$FUNCNAME: Dev agent, setting MaxRetries to 0..."
             agentExtraConfig='{"MaxRetries":0}'
+        elif [[ "$TEAMNAME" == Tier0* ]]; then
```
amaltaro, line 580: I would move this elif statement to the second position (2nd to be evaluated), given that the other 2 could be part of a Tier0 agent setup.

docker/pypi/wmagent/install.sh
```diff
@@ -105,9 +105,9 @@ echo "-----------------------------------------------------------------------"
 stepMsg="Generating and preserving current build id"
 echo "-----------------------------------------------------------------------"
 echo "Start $stepMsg"
-echo $RANDOM |sha256sum |awk '{print $1}' > $WMA_ROOT_DIR/.dockerBuildId
-echo "WMA_BUILD_ID:`cat $WMA_ROOT_DIR/.dockerBuildId`"
-echo "WMA_BUILD_ID preserved at: $WMA_ROOT_DIR/.dockerBuildId "
+echo $RANDOM |sha256sum |awk '{print $1}' > $WMA_ROOT_DIR/.wmaBuildId
```
amaltaro, line 108: Won't this reinitialize the whole deployment process whenever a new installation is performed?
This might be the case for Tier0, which is using venv for deploying the service, but not for production, as provided in your fix here: https://github.com/dmwm/CMSKubernetes/pull/1485

amaltaro (changes requested):
@todor-ivanov I left a bunch of comments along the code. Let me mention some of them here as well:
* do we really need to provide the renew proxy functionality now? If this is not a problem to be fixed, I would leave it for the future;
* I would not make any further changes to the opportunistic resources handling. It's already making the secrets file very overloaded with configuration and parameters.

In addition, having 100s of lines of change at this stage is worrisome, given that we are already seriously considering this setup for pre-production.

I also don't have all the knowledge necessary to provide a more meaningful review. I do think though that we would profit from documenting this deployment, as I feel like we should separate:
* docker deployment for central production
* from virtual environment deployment for Tier0
but again, we can only evaluate it properly once we know the current steps required for T0 deployment.

docker/pypi/wmagent/bin/manage
```diff
@@ -74,7 +74,12 @@ activate_agent(){
     # The very first step of activating the WMAgent upon all needed checks and
     # validations performed during the initialization process.
     # No .init* flags are checked at this stage.
-    cp -vf $WMA_DEPLOY_DIR/etc/WMAgentConfig.py $WMA_CONFIG_DIR/config-template.py
+    if [[ "$TEAMNAME" == Tier0* ]]; then
+	cat $WMA_DEPLOY_DIR/etc/WMAgentConfig.py $WMA_DEPLOY_DIR/etc/Tier0Config.py > $WMA_CONFIG_DIR/config-template.py
```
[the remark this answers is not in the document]
LinaresToine: Yes. The resulting config file is as expected

docker/pypi/wmagent/bin/manage-common.sh
```diff
@@ -265,8 +275,22 @@ _status_of_oracle(){
 _renew_proxy(){
     # Auxiliary function to renew agent proxy
     local hostName=`hostname -f`
+    _load_wmasecrets
 
     # Here to find out if the agent is CERN or FNAL and use the proper credentials name for _renew_proxy
+    [[ "$TEAMNAME" == Tier0* ]] &&  {
```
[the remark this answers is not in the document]
todor-ivanov: yep, this is needed as  much for them  just as it is for us.

docker/pypi/wmagent/init.sh
```diff
@@ -34,12 +34,12 @@ AGENT_FLAVOR=mysql
 # are to be defined for the first time or otherwise their default values will be used
 [[ -f $WMA_SECRETS_FILE ]] && _load_wmasecrets
 
-# Find the current WMAgent Docker image BuildId:
-# NOTE: The $WMA_BUILD_ID is exported only from $WMA_USER/.bashrc but not from the Dockerfile ENV command
-[[ -n $WMA_BUILD_ID ]] || WMA_BUILD_ID=$(cat $WMA_ROOT_DIR/.dockerBuildId) || { echo "ERROR: Cuold not find/set WMA_UILD_ID"; exit 1 ;}
+# Find the current WMAgent BuildId:
+# NOTE: The $WMA_BUILD_ID is exported from $WMA_ENV_FILE but not from the Dockerfile ENV command
+[[ -n $WMA_BUILD_ID ]] || WMA_BUILD_ID=$(cat $WMA_ROOT_DIR/.wmaBuildId) || { echo "ERROR: Cuold not find/set WMA_UILD_ID"; exit 1 ;}
```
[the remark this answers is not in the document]
todor-ivanov: > Don't we execute the install.sh script for T0 deployment:

Yes

> In addition, the install.sh uses .dockerBuildId while here we have a reference to .wmaBuildId; what is the actual difference between those 2?

Do difference  one should be a rename of the other. I may have forgotten to change that in the install.sh scirpt. I'll double check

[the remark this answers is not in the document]
todor-ivanov: Nope,  it is renamed also in the `install.sh` script, but it does not yet reflect the new deployemtn granularit from our latest Bugfix. I'll rebase this branch yet again.

docker/pypi/wmagent/init.sh
```diff
@@ -98,54 +101,58 @@ basic_checks() {
...
+    # Avoid copying the run time scripts if not running from inside a Docker container
+    _is_venv && {
+        echo $WMA_BUILD_ID > $wmaInitRuntime
+        return
+    }
+
+    # checking if $WMA_DEPLOY_DIR is root path for $pythonLib:
+    if [[ $pythonLib =~ ^$WMA_DEPLOY_DIR ]]; then
+        mkdir -p $WMA_INSTALL_DIR/Docker/
+        cp -rav $pythonLib/WMCore/WMRuntime $WMA_INSTALL_DIR/Docker/
```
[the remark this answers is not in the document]
todor-ivanov: ok

docker/pypi/wmagent/init.sh
```diff
@@ -98,54 +101,58 @@ basic_checks() {
...
+        cp -rav $WMA_DEPLOY_DIR/etc/ $WMA_CONFIG_DIR/
+        echo $WMA_BUILD_ID > $wmaInitRuntime
+    else
+        echo "$FUNCNAME: ERROR: \$WMA_DEPLOY_DIR: $WMA_DEPLOY_DIR is not a root path for \$pythonLib: $pithonLib"
```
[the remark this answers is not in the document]
todor-ivanov: 10x

docker/pypi/wmagent/init.sh
```diff
@@ -329,12 +329,27 @@ EOF
     echo "-----------------------------------------------------------------------"
 }
 
-check_docker_init() {
-    # A function to check all previously populated */.dockerInit files
-    # from all previous steps and compare them with the /data/.dockerBuildId
+check_wmagnet_init() {
```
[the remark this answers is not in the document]
todor-ivanov: Fixed. Also in the relevant calls to this very same function. so at the end it was not breaking and the typo went through unnoticed

docker/pypi/wmagent/init.sh
```diff
@@ -498,18 +519,28 @@ agent_resource_opp() {
...
+                    resOpt="$resOpt --$opt=${res[$opt]}"
+                fi
+            done
+
+            if [[ ${res[name]} =~ .*_US_.* ]] && [[ $HOSTNAME =~ .*cern\.ch ]]; then
                 echo "I am based at CERN, so I cannot use US opportunistic resources, moving to the next site"
                 continue
             else
-                manage execute-agent wmagent-resource-control --plugin=SimpleCondorPlugin --opportunistic --pending-slots=${site[pend]} --running-slots=${site[run]} --add-one-site=${site[name]} ; let errVal+=$?
+                manage execute-agent wmagent-resource-control $resOpt || let errVal+=$?
```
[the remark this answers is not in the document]
todor-ivanov: That would be pity, since this proved to be quite useful... and the code change here was meant to only fix one error we made last time with @anpicci  - to base the data structure key names on the actual set of parameter names which the `wmagent-resource-control` script supports. So ti is actually intended to fix an overlooked part related to our code. This was not meant to serve T0 deployment needs. The fact that it could be used by the T0 team in their deployment process, came just as a side effect. .....  but ok, I'll remove this part.

docker/pypi/wmagent/init.sh
todor-ivanov: hahaha  :) man this is what happens when one uses a spellchecker at  3 AM

docker/pypi/wmagent/init.sh
todor-ivanov: I do not think so. The Tier0's team names should always start with the Tier0*, meaning there is no inter mixture of all those options. Actually the better construction to implement those checks here would be a `case ` statement, rather than an `elif` sequence.

docker/pypi/wmagent/install.sh
todor-ivanov: No ... actually T0 should follow the same re-initialization granularity as us.  I have already rebased this PR on top of our latest  changes  in this regard

todor-ivanov:
hi @amaltaro  I have addressed your comments with my latest commits. Please  take another look.

docker/pypi/wmagent/bin/manage-common.sh
```diff
@@ -265,8 +275,22 @@ _status_of_oracle(){
...
+	echo "$FUNCNAME: This is a Tier0 agent"
+	local vomsproxyCmd="voms-proxy-init -rfc \
+	            -voms cms:/cms/Role=production -valid 168:00 -bits 2048 \
+		    -cert $X509_USER_CERT -key $X509_USER_KEY \
+                    -out  $X509_USER_PROXY"
+	$vomsproxyCmd || {
+	    echo "$FUNCNAME: ERROR: Failed to renew invalid myproxy"
+	    return $(false)
+	}
+    return
```
amaltaro, line 291: Even though indentation does not matter in bash, it would be great to fix this return.

docker/pypi/wmagent/bin/manage-common.sh
```diff
@@ -359,8 +383,21 @@ _load_wmasecrets(){
     for varName in $varsToLoad
     do
         value=`grep -E "^[[:blank:]]*$varName=" $secretsFile | sed "s/ *$varName=//"`
-        [[ $varName =~ ^RESOURCE_ ]] && declare -g -A $varName
-        eval $varName=$value
+        if [[ $varName =~ ^RESOURCE_ ]]; then
+	        declare -g -A $varName
+            eval $varName=$value || {
```
amaltaro, line 388: Same with all this new block of code, some lines are not properly indented and make it a bit harder to read.

docker/pypi/wmagent/init.sh
```diff
@@ -34,12 +34,12 @@ AGENT_FLAVOR=mysql
 # are to be defined for the first time or otherwise their default values will be used
 [[ -f $WMA_SECRETS_FILE ]] && _load_wmasecrets
 
-# Find the current WMAgent Docker image BuildId:
-# NOTE: The $WMA_BUILD_ID is exported only from $WMA_USER/.bashrc but not from the Dockerfile ENV command
-[[ -n $WMA_BUILD_ID ]] || WMA_BUILD_ID=$(cat $WMA_ROOT_DIR/.dockerBuildId) || { echo "ERROR: Cuold not find/set WMA_UILD_ID"; exit 1 ;}
+# Find the current WMAgent BuildId:
+# NOTE: The $WMA_BUILD_ID is exported from $WMA_ENV_FILE but not from the Dockerfile ENV command
+[[ -n $WMA_BUILD_ID ]] || WMA_BUILD_ID=$(cat $WMA_ROOT_DIR/.wmaBuildId) || { echo "ERROR: Cuold not find/set WMA_BUILD_ID"; exit 1 ;}
```
amaltaro, line 39: typo: Cuold

docker/pypi/wmagent/init.sh
```diff
@@ -98,54 +101,60 @@ basic_checks() {
...
-            cp -rav $pythonLib/WMCore/WMRuntime $WMA_INSTALL_DIR/Docker/
-            cp -rav $WMA_DEPLOY_DIR/etc/ $WMA_CONFIG_DIR/
-            echo $WMA_BUILD_ID > $wmaInitRuntime
-        else
-            echo "$FUNCNAME: ERROR: \$WMA_DEPLOY_DIR: $WMA_DEPLOY_DIR is not a root path for \$pythonLib: $pithonLib"
-            echo "$FUNCNAME: ERROR: We cannot find the correct WMCore/WMRuntime source to copy at the current host!"
-            return $(false)
-        fi
-    }
+    _init_valid $wmaInitRuntime || _copy_runtime
```
amaltaro, line 146: This behavior is not clear to me. If we stop a container running 2.3.3.1 and start a container 2.3.3.2, will the runtime code be copied over once again?

From what I understood, this `$wmaInitRuntime` contains the WMAgent hash build id based on the release cycle. Given that the release cycle is the same, it will not retrigger the runtime copy when upgrading a patch release. Can you please correct/confirm this understanding?

If it does not copy, then we will have to change this behavior. As a patch release might bring changes to the WMRuntime package and we need to ensure that those are properly deployed together with a new docker image version.

docker/pypi/wmagent/init.sh
```diff
@@ -329,12 +329,27 @@ EOF
...
+    # NOTE: On every step we need to check the .init<step> file content. We always compare
+    #       the current $WMA_BUILD_ID with the one previously initialized at the host.
+    #       The WMA_BUILD_ID is generated during the execution of `install.sh`
+    #       at build time (for docker images) or deploy time for virtual env
+    #       There are three levels of comparison we can make:
+    #       * if we want to trigger re-initialization on any WMAgent image rebuild,
+    #         then the $WMA_BUILD_ID should contain a sha256sum of a random variable
+    #       * if we want to trigger re-initialization only on new WMAgent tag builds,
+    #         then the $WMA_BUILD_ID should contain a sha256sum of whole $WMA_TAG
+    #       * if we want to trigger re-initialization only on version change (not on patches),
```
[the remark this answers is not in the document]
amaltaro: Then I would suggest making the bullet points above as an ordered list, just so whoever is working with this code can understand that there are multiple ways that this could have been implemented. Nonetheless, I still think the most important is to say what IS the expected functionality and behavior.

docker/pypi/wmagent/init.sh
```diff
@@ -329,12 +331,32 @@ EOF
...
+    #         then we should split $WMA_TAG in parts: major, minor and patch(release candidate)
+    #         part and the $WMA_BUILD_ID should contain a sha256sum only of the major + minor
+    #         part excluding the patch version, e.g.:
+    #         WMA_TAG=2.3.3.1;
+    #         WMA_VER[release]=2.3.3
+    #         WMA_VER[major]=2.3
+    #         WMA_VER[minor]=3
+    #         WMA_VER[patch]=1
+    #       The current implementation considers the first one - re-initialization on any WMAgent rebuild
```
amaltaro, line 358: This last line in this comment is also confusing to me:
"The current implementation considers the first one - re-initialization on any WMAgent rebuild"

What is "current implementation"? What is already in the repository? Or what is provided in this PR?

In addition, I wonder in practice what is the actual meaning of "re-initialization on any WMAgent rebuild"? Is it that when we run a new container version, we will perform every single step of the initialization again? E.g.:
1) defining cronjobs
2) setup couchapps
3) (re-)create the agent config.py
4) create the SQL database schema
5) and so on?

docker/pypi/wmagent/init.sh
amaltaro: Either one is fine with me. However, I still insist that matching against Tier0* is much more restrict than the dev and testbed. To avoid any "human mistake" in the future like "team=Tier0-dev1", please move it up.

docker/pypi/wmagent/init.sh
```diff
@@ -594,6 +626,13 @@ main(){
         echo "            docker kill wmagent"
         echo "          * Start a fresh instance of wmagent:"
         echo "            ./wmagent-docker-run.sh -t <WMA_TAG> && docker logs -f wmagent"
+        echo
+        echo "     d) If you are deploying inside a virtual environment"
```
amaltaro, line 630: Where is c)? :)

amaltaro (changes requested):
@todor-ivanov please find further comments along the code.

docker/pypi/wmagent/init.sh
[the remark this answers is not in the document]
todor-ivanov: ok

docker/pypi/wmagent/init.sh
todor-ivanov: done

docker/pypi/wmagent/bin/manage-common.sh
todor-ivanov: because they were copiy-pasted from old Tier0 code, where somebody long long ago....  has used tabs the indentation, which is supper annoying, I admit. I'll fix it wherever I can find it, but if you notice some more, please point them as well.

docker/pypi/wmagent/bin/manage-common.sh
todor-ivanov: k

docker/pypi/wmagent/init.sh
todor-ivanov: fixed

docker/pypi/wmagent/init.sh
todor-ivanov: No, it meant that whenever somebody goes and executes `install.sh` which is a step during the docker image build process for the production agents and an equivalent step during the creation of the virtual env for the agents who are about to use it, the resultant docker image or virtual env root would have their own unique to them id and they must go through the initialization process at least once. But not on every restart. 
The current implementation is not this one any more. We switched to the model where we restart the initialization process only whenever someone changes  the release version.

docker/pypi/wmagent/init.sh
todor-ivanov: > This behavior is not clear to me. If we stop a container running 2.3.3.1 and start a container 2.3.3.2, will the  runtime code be copied over once again?

Not any more

> From what I understood, this $wmaInitRuntime contains the WMAgent hash build id based on the release cycle. > Given that the release cycle is the same, it will not retrigger the runtime copy when upgrading a patch release.  Can you please correct/confirm this understanding?

That is correct. It won't re-trigger the runtime code copy step.

> If it does not copy, then we will have to change this behavior.

That's  going to be tricky, because it is not about to go through the initialization steps at all on a patch version or release candidate change, so the process would never land here. I'll have to think of something.

docker/pypi/wmagent/init.sh
todor-ivanov: Not needed in the new Latin alphabet ;)
fixing it

docker/pypi/wmagent/init.sh
todor-ivanov: ok, I think I have an idea on how to fix this behavior, but I really want to stress ... this is something we should have caught while resolving this issue: https://github.com/dmwm/WMCore/issues/11990 and the change which will follow does not relate in any way to Tier0. I am just going to use this PR to fix it.

todor-ivanov: ok ... the intervention would be too intrusive and we better do it separately. I'll reopen the other issue and will follow with a new PR just for this.

amaltaro: Yes, it's probably a better idea. Thanks

todor-ivanov:
@amaltaro  please go ahead and take another look. Modulo the change of behavior in regards to the Runtime files (which is not related  to the T0 deployment). The later I'll fix with a separate PR.

docker/pypi/wmagent/bin/manage-common.sh
amaltaro: Todor, did you forget pushing your changes in? This whole block above is inside `[[ "$TEAMNAME" == Tier0* ]] &&  {`, so I expected it to be further indented.

docker/pypi/wmagent/bin/manage-common.sh
amaltaro: Thank you for fixing it. I think the `fi` statement in L399 needs one extra indentation (same level as the else clause)

amaltaro (approved):
@todor-ivanov changes look good to me. However, if you can fix those 2 minor mis-indentantions in the next few hours, I would wait for it. Thanks

amaltaro:
Don't worry with those leftover comments, I addressed them in https://github.com/dmwm/CMSKubernetes/pull/1489
Thank you and enjoy your holiday!

todor-ivanov:
> Don't worry with those leftover comments, I addressed them in https://github.com/dmwm/CMSKubernetes/pull/1489
> Thank you and enjoy your holiday!

Thanks  @amaltaro , but I am not even close to my holiday. For the last few hours I was actually having quite a fight in the match for the title in heavy weight against  Docker Engine. It took me a while to teach him how to parse a variable of the form `[0-9]+.[0-9]+.[0-9]+(.[0-9]+|rc[0-9]+)*`. At the end, I some how managed to prevail, but I cannot express with words the level of  rudimentality I faced in the Dockerfile parsing routines when it comes to parameter expansion. Nothing like the old bash. The result, of course, was  far from beautiful. We both (me and the Dockerfile of the WMAgent containers) will carry the scars from this cheerful encounter for life ;) Unfortunately this so engaging endeavor took all my attention and I missed your request for these extra refinements. Thanks for addressing them. I'll post the summary of this so crucial for the season sports event in the form of a PR shortly, and I will call for your review. Stay tuned.

PR #1467, opened 2024-04-23
Rebuild /etc/passwd and /etc/group files at runtime

todor-ivanov:
One out of the two suggested  solutions in this comments: https://github.com/dmwm/WMCore/issues/11944#issuecomment-2071376685 . The second one in particular.

### DIFF
--- a/docker/pypi/wmagent/Dockerfile
+++ b/docker/pypi/wmagent/Dockerfile
@@ -58,6 +58,11 @@ ADD init.sh ${WMA_ROOT_DIR}/init.sh
 RUN ${WMA_ROOT_DIR}/install.sh -t ${WMA_TAG}
 RUN chown -R ${WMA_USER}:${WMA_GID} ${WMA_ROOT_DIR}
 
+# Set command prompt for root inside the container
+RUN <<EOF cat >>/root/.bashrc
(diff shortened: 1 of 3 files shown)


todor-ivanov:
@vkuznet  @amaltaro I am about to add  few more lines to this related to fixing the command prompt for the root and the regular user... It is quite important to have it, in order to be distinguishable the mode one runs  into (iehter inside or outside the container). And it is now a must to have it setup through `/etc/profile` file at the container, that's the reason why I removed it from the mount points. Also we need the profile file from the container for the agent setup as well, so we  cannot mount it from the host

vkuznet:
Todor, I'm confused with usage of `getent` command. Let me clarify. When we create a fresh image there is no users/groups in it. Therefore, in order for `getent` to provide us with cmst1 user name/group we should provide at run-time an admin database where `getent` will read from. If it is `/etc/passwd` and if cmst1 user not in it (on a host node) how `getent` will get this user, i.e. from which DB. In other words what should we supply to the image at run-time from the host node which will have this user?

amaltaro:
@todor-ivanov I think this PR has also been superseeded by https://github.com/dmwm/CMSKubernetes/pull/1469. Should we close this out?

todor-ivanov:
correct

PR #1468, opened 2024-04-24
Start WMAgent container with dynamic account

amaltaro:
Proposal to fix https://github.com/dmwm/WMCore/issues/11944

Similar changes are required for MariaDB and CouchDB, but it will only be done if we decide to move forward with this alternative to detect the user name during execution of the containers.

Reference: https://github.com/Enchufa2/docker-host-auth?tab=readme-ov-file

### DIFF
--- a/docker/pypi/wmagent-base/Dockerfile
+++ b/docker/pypi/wmagent-base/Dockerfile
@@ -3,7 +3,8 @@ FROM registry.cern.ch/cmsweb/dmwm-base:pypi-20230525
 
 # Install basic OS package dependencies
 RUN apt-get update
-RUN apt-get install -y libmariadb-dev-compat libmariadb-dev apache2-utils cron mariadb-server myproxy voms-clients rlwrap libaio1 && apt-get clean
+RUN apt-get install -y libmariadb-dev-compat libmariadb-dev apache2-utils cron mariadb-server \
+   myproxy voms-clients rlwrap libaio1 sssd && apt-get clean
 
 # Install some debugging tools
 RUN apt-get install -y hostname net-tools iputils-ping procps && apt-get clean
@@ -13,3 +14,11 @@ COPY --from=oracle /usr/lib/oracle /usr/lib/oracle
 ENV LD_LIBRARY_PATH=/usr/lib/oracle
 ENV PATH=$PATH:/usr/lib/oracle
 ENV PKG_CONFIG_PATH=/usr/lib/oracle
+
+# overwrite /etc/pam.d/login such that we can access accounts at CERN
+RUN printf "\n\
+auth required pam_sss.so\n\
+account required pam_sss.so\n\
+password required pam_sss.so\n\
+session required pam_sss.so\n\
+" > /etc/pam.d/login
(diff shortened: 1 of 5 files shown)


docker/pypi/wmagent-base/Dockerfile
```diff
@@ -13,3 +14,11 @@ COPY --from=oracle /usr/lib/oracle /usr/lib/oracle
 ENV LD_LIBRARY_PATH=/usr/lib/oracle
 ENV PATH=$PATH:/usr/lib/oracle
 ENV PKG_CONFIG_PATH=/usr/lib/oracle
+
+# overwrite /etc/pam.d/login such that we can access accounts at CERN
+RUN printf "\n\
```
todor-ivanov, line 19: Can we use the same pattern with a heredoc everywhere we need to edit documents at the image? 

like: 
```
RUN <<EOF cat >> /etc/pam.d/login
auth required pam_sss.so
account required pam_sss.so
password required pam_sss.so
session required pam_sss.so
EOF
```
It avoids syntactic complications and workarounds  with escape characters etc.

docker/pypi/wmagent/run.sh
```diff
@@ -1,11 +1,22 @@
...
 
 ### Basic initialization wrapper for WMAgent to serve as the main entry point for the WMAgent Docker container
+wmaUser=`id -un`
+wmaGroup=`id -gn`
+wmaUserID=`id -u`
+wmaGroupID=`id -g`
+export WMA_USER=$wmaUser
+echo "Running WMAgent container with user: $wmaUser (ID: $wmaUserID) and group: $wmaGroup (ID: $wmaGroupID)"
+
+echo "Correcting ownership for WMA_ROOT_DIR: $WMA_ROOT_DIR"
```
todor-ivanov, line 11: Same comment as to the PR with the alternative approach

docker/pypi/wmagent/run.sh
```diff
@@ -1,11 +1,22 @@
...
+find $WMA_ROOT_DIR \! \( -user $wmaUser -group $wmaGroup \) -exec chown -f $wmaUser:$wmaGroup '{}' +;
+
+# append the WMAgent user to the mysql group
+usermod -aG mysql ${WMA_USER}
```
todor-ivanov, line 15: Same comment as to the PR with the alternative approach

docker/pypi/wmagent/wmagent-docker-run.sh
```diff
@@ -44,8 +44,8 @@ while getopts ":t:hp" opt; do
 done
 
 
-wmaUser=cmst1
-wmaOpts=" --user $wmaUser"
+wmaUser=$(id -un)
+userOpts=" --user $(id -u):$(id -g)"
```
todor-ivanov, line 48: Well technically speaking, those are again `docker options` so the proper way is for them to be concatenated to the `dockerOpt` string variable.  The existence of the two variables:  `wmaOpts` and `dockerOpts` has different origin - it is their application target, rather their functional separation: 

* `dockerOpts` is the set of parameters passed to the docker engine 
* `wmaOpts` is the set of parameters passed to the service running inside the image (i.e. the entry point).

(The same applies to the MariaDB and CouchDB containers)

PR #1469, opened 2024-04-24
Set WMAgent/MariaDB/CouchDB user during container runtime

amaltaro:
Fixes https://github.com/dmwm/WMCore/issues/11944
Superseeds https://github.com/dmwm/CMSKubernetes/pull/1464

First commit contains all the changes that Valentin has in #1464; while 2nd commit onward contains additional changes to get the whole setup working.

Short summary of the changes are:
* create a template for bashrc which will be copied over to the user running the container (remove bashrc juggling from the `install.sh` script)
* workaround account access based on SSSD daemon by juggling with /etc/passwd and /etc/group and mounting them into the container. Implemented in `wmagent-docker-run.sh` script.
* remove any referece to service accounts in the dockerfiles
* use the docker `--user` parameter to start up the container with the current/correct user.
* move cronjob creation to the `init.sh` script (to properly create it against the correct user)

Similar fixes have been provided to the `wmagent-couchdb` and `wmagent-mariadb` images.

TODO: at some point we should release a `latest` tag for CouchDB, to ensure the image does not get deleted by the automatic cmsweb retention policy.

### DIFF
--- a/docker/pypi/wmagent-couchdb/Dockerfile
+++ b/docker/pypi/wmagent-couchdb/Dockerfile
@@ -13,16 +13,6 @@ RUN apt-get install -y hostname net-tools iputils-ping procps emacs-nox tcpdump 
 
 RUN pip install CMSCouchapp
 
-# ENV USER=couchdb
-# ENV GROUP=couchdb
-ENV USER=cmst1
-ENV GROUP=zh
-ENV COUCH_UID=31961
-ENV COUCH_GID=1399
-ENV COUCH_PORT=5984
-# ENV COUCH_UID=5984
-# ENV COUCH_GID=5984
-
 ENV COUCH_ROOT_DIR=/data
 
 ENV COUCH_BASE_DIR=$COUCH_ROOT_DIR/srv/couchdb
@@ -46,16 +36,6 @@ ENV WMA_SECRETS_FILE=$WMA_ADMIN_DIR/WMAgent.secrets
 
 # RUN mkdir -p /etc/grid-security
 
-# # Setting up users and previleges
-# # THIS MUST HAPPEN MANUALLY ON THE HOST:
-RUN groupadd -g ${COUCH_GID} ${GROUP}
-RUN useradd -u ${COUCH_UID} -g ${COUCH_GID} -m ${USER}
-RUN install -o ${USER} -g ${COUCH_GID} -d ${COUCH_ROOT_DIR}
-
-
-# add user to sudoers file
-RUN echo "$USER ALL=(ALL) NOPASSWD:ALL" >> /etc/sudoers
-
 # start the setup
 RUN mkdir -p $COUCH_ROOT_DIR
 
@@ -63,9 +43,6 @@ ENV PATH="${COUCH_ROOT_DIR}:${PATH}"
 
 RUN mkdir -p $COUCH_CURRENT_DIR $COUCH_CONFIG_DIR $COUCH_MANAGE_DIR $COUCH_LOG_DIR $COUCH_DATABASE_DIR $COUCH_STATE_DIR $COUCH_AUTH_DIR
 RUN ln -s $COUCH_CURRENT_DIR $COUCH_BASE_DIR/current
-
-# ENV COUCHDB_USER=admin
-# ENV COUCHDB_PASSWORD=adminpass
 
 # add necessary scripts
 ADD run.sh ${COUCH_ROOT_DIR}/
@@ -79,8 +56,7 @@ RUN ln -s ${COUCH_CONFIG_DIR}/local.ini /opt/couchdb/etc/local.d/
 ENV PATH="/opt/couchdb/bin:/usr/local/bin/:${PATH}"
 ENV CRYPTOGRAPHY_ALLOW_OPENSSL_102=true
 
-# RUN <<EOF cat >> /opt/couchdb/.bashrc
-RUN <<EOF cat >> /home/$USER/.bashrc
+RUN <<EOF cat >> ~/.bashrc
 
 alias lll="ls -lathr"
 alias ls="ls --color=auto"
@@ -92,9 +68,17 @@ alias manage=$COUCH_MANAGE_DIR/manage
 export PS1="(CouchDB-$TAG) [\u@\h:\W]\$ "
 EOF
 
-RUN chown -R ${USER}:${GROUP} ${COUCH_ROOT_DIR}
+# set CouchDB docker specific bash prompt and manage alias for all users:
+RUN <<EOF cat >>/root/.bashrc
+alias manage=$COUCH_MANAGE_DIR/manage
+export PS1="(CouchDB-$TAG) [\u@\h:\W]\$([[ \$(id -u) -eq 0 ]] && echo \# || echo \$) "
+EOF
+
+RUN <<EOF cat >> ~/.bashrc
+alias manage=$COUCH_MANAGE_DIR/manage
+export PS1="(CouchDB-$TAG) [\u@\h:\W]\$([[ \$(id -u) -eq 0 ]] && echo \# || echo \$) "
+EOF
 
 # setup final environment
-USER $USER
 WORKDIR $COUCH_ROOT_DIR
 ENTRYPOINT ["./run.sh"]
--- a/docker/pypi/wmagent-couchdb/Dockerfile
+++ b/docker/pypi/wmagent-couchdb/Dockerfile
@@ -13,6 +13,7 @@ RUN apt-get install -y hostname net-tools iputils-ping procps emacs-nox tcpdump 
 
 RUN pip install CMSCouchapp
 
+ENV COUCH_PORT=5984
 ENV COUCH_ROOT_DIR=/data
 
 ENV COUCH_BASE_DIR=$COUCH_ROOT_DIR/srv/couchdb
@@ -50,7 +51,7 @@ ADD manage ${COUCH_MANAGE_DIR}/manage
 RUN ln -s ${COUCH_MANAGE_DIR}/manage ${COUCH_ROOT_DIR}/manage
 
 # The $COUCH_CONFIG_DIR is to be mounted from the host and locla.ini read from there
-ADD local.ini ${COUCH_CONFIG_DIR}/local.ini
+ADD local.ini ${COUCH_DEPLOY_DIR}/local.ini
 RUN ln -s ${COUCH_CONFIG_DIR}/local.ini /opt/couchdb/etc/local.d/
 
 ENV PATH="/opt/couchdb/bin:/usr/local/bin/:${PATH}"
--- a/docker/pypi/wmagent-couchdb/Dockerfile
+++ b/docker/pypi/wmagent-couchdb/Dockerfile
@@ -56,28 +56,13 @@ RUN ln -s ${COUCH_CONFIG_DIR}/local.ini /opt/couchdb/etc/local.d/
 ENV PATH="/opt/couchdb/bin:/usr/local/bin/:${PATH}"
 ENV CRYPTOGRAPHY_ALLOW_OPENSSL_102=true
 
-RUN <<EOF cat >> ~/.bashrc
-
-alias lll="ls -lathr"
-alias ls="ls --color=auto"
-alias ll='ls -la --color=auto'
-
-alias manage=$COUCH_MANAGE_DIR/manage
-
-# set CouchDB docker specific bash prompt:
-export PS1="(CouchDB-$TAG) [\u@\h:\W]\$ "
+# Set command prompt for root
+RUN <<EOF cat >> /root/.bashrc
+export PS1="(CouchDB-$TAG) [\u@\h:\W]# "
 EOF
 
-# set CouchDB docker specific bash prompt and manage alias for all users:
-RUN <<EOF cat >>/root/.bashrc
-alias manage=$COUCH_MANAGE_DIR/manage
-export PS1="(CouchDB-$TAG) [\u@\h:\W]\$([[ \$(id -u) -eq 0 ]] && echo \# || echo \$) "
-EOF
-
-RUN <<EOF cat >> ~/.bashrc
-alias manage=$COUCH_MANAGE_DIR/manage
-export PS1="(CouchDB-$TAG) [\u@\h:\W]\$([[ \$(id -u) -eq 0 ]] && echo \# || echo \$) "
-EOF
+# allow dynamic users to create homefolders and .bashrc
+RUN chmod 777 /home
 
 # setup final environment
 WORKDIR $COUCH_ROOT_DIR
--- a/docker/pypi/wmagent-couchdb/couchdb-docker-run.sh
+++ b/docker/pypi/wmagent-couchdb/couchdb-docker-run.sh
@@ -75,9 +75,10 @@ groupEntry=$(getent group $thisGroup)
 
 [[ -d $HOST_MOUNT_DIR/certs ]] || mkdir -p $HOST_MOUNT_DIR/certs || exit $?
 [[ -d $HOST_MOUNT_DIR/admin/couchdb ]] || mkdir -p $HOST_MOUNT_DIR/admin/couchdb || exit $?
-# [[ -d $HOST_MOUNT_DIR/srv/couchdb/$COUCH_TAG/config  ]] || mkdir -p $HOST_MOUNT_DIR/srv/couchdb/$COUCH_TAG/config  || exit $?
-[[ -d $HOST_MOUNT_DIR/srv/couchdb/$COUCH_TAG/install/database ]] || mkdir -p $HOST_MOUNT_DIR/srv/couchdb/$COUCH_TAG/install/database || exit $?
+[[ -d $HOST_MOUNT_DIR/srv/couchdb/$COUCH_TAG/config  ]] || mkdir -p $HOST_MOUNT_DIR/srv/couchdb/$COUCH_TAG/config  || exit $?
+[[ -d $HOST_MOUNT_DIR/srv/couchdb/$COUCH_TAG/install ]] || mkdir -p $HOST_MOUNT_DIR/srv/couchdb/$COUCH_TAG/install || exit $?
 [[ -d $HOST_MOUNT_DIR/srv/couchdb/$COUCH_TAG/logs ]] || mkdir -p $HOST_MOUNT_DIR/srv/couchdb/$COUCH_TAG/logs || exit $?
+[[ -d $HOST_MOUNT_DIR/srv/couchdb/$COUCH_TAG/state ]] || mkdir -p $HOST_MOUNT_DIR/srv/couchdb/$COUCH_TAG/state || exit $?
 
 
 dockerOpts="
@@ -89,8 +90,10 @@ dockerOpts="
 --name=couchdb \
 --mount type=bind,source=/tmp,target=/tmp \
 --mount type=bind,source=$HOST_MOUNT_DIR/certs,target=/data/certs \
---mount type=bind,source=$HOST_MOUNT_DIR/srv/couchdb/$COUCH_TAG/install/database,target=/data/srv/couchdb/current/install/database \
+--mount type=bind,source=$HOST_MOUNT_DIR/srv/couchdb/$COUCH_TAG/install,target=/data/srv/couchdb/current/install \
 --mount type=bind,source=$HOST_MOUNT_DIR/srv/couchdb/$COUCH_TAG/logs,target=/data/srv/couchdb/current/logs \
+--mount type=bind,source=$HOST_MOUNT_DIR/srv/couchdb/$COUCH_TAG/state,target=/data/srv/couchdb/current/state \
+--mount type=bind,source=$HOST_MOUNT_DIR/srv/couchdb/$COUCH_TAG/config,target=/data/srv/couchdb/current/config \
 --mount type=bind,source=$HOST_MOUNT_DIR/admin/wmagent,target=/data/admin/wmagent/ \
 --mount type=bind,source=$HOST_MOUNT_DIR/admin/couchdb,target=/data/admin/couchdb/ \
 --mount type=bind,source=$HOST_MOUNT_DIR/admin/etc/passwd,target=/etc/passwd,readonly \
@@ -100,6 +103,7 @@ dockerOpts="
 "
 
 # --mount type=bind,source=$HOST_MOUNT_DIR/srv/couchdb/$COUCH_TAG/config,target=/data/srv/couchdb/current/config \
+# --mount type=bind,source=$HOST_MOUNT_DIR/srv/couchdb/$COUCH_TAG/install/database,target=/data/srv/couchdb/current/install/database \
 
 registry=local
 repository=wmagent-couchdb
--- a/docker/pypi/wmagent-couchdb/couchdb-docker-run.sh
+++ b/docker/pypi/wmagent-couchdb/couchdb-docker-run.sh
@@ -69,6 +69,7 @@ dockerOpts="
 --detach \
 --network=host \
 --rm \
+--user $UID:`id -g` \
 --hostname=`hostname -f` \
 --name=couchdb \
 --mount type=bind,source=/tmp,target=/tmp \
@@ -77,6 +78,12 @@ dockerOpts="
 --mount type=bind,source=$HOST_MOUNT_DIR/srv/couchdb/$COUCH_TAG/logs,target=/data/srv/couchdb/current/logs \
 --mount type=bind,source=$HOST_MOUNT_DIR/admin/wmagent,target=/data/admin/wmagent/ \
 --mount type=bind,source=$HOST_MOUNT_DIR/admin/couchdb,target=/data/admin/couchdb/ \
+--mount type=bind,source=/etc/group,target=/etc/group,readonly \
+--mount type=bind,source=/etc/passwd,target=/etc/passwd,readonly \
+--mount type=bind,source=/etc/shadow,target=/etc/shadow,readonly \
+--mount type=bind,source=/etc/sudoers,target=/etc/sudoers,readonly \
+--mount type=bind,source=/etc/sudoers.d,target=/etc/sudoers.d,readonly \
+--mount type=bind,source=/etc/profile,target=/etc/profile,readonly \
 "
 
 # --mount type=bind,source=$HOST_MOUNT_DIR/srv/couchdb/$COUCH_TAG/config,target=/data/srv/couchdb/current/config \
--- a/docker/pypi/wmagent-couchdb/couchdb-docker-run.sh
+++ b/docker/pypi/wmagent-couchdb/couchdb-docker-run.sh
@@ -49,13 +49,29 @@ while getopts ":t:hp" opt; do
     esac
 done
 
-
-couchdbUser=cmst1
-couchOpts=" --user $couchdbUser"
-
 # This is the root at the host only, it may differ from the root inside the container.
 # NOTE: this may be parameterized, so that the container can run on a different mount point.
 HOST_MOUNT_DIR=/data/dockerMount
+
+thisUser=$(id -un)
+thisGroup=$(id -gn)
+
+# create the passwd and group mount point dynamically at runtime
+passwdEntry=$(getent passwd $thisUser | awk -F : -v thisHome="/home/$thisUser" '{print $1 ":" $2 ":" $3 ":" $4 ":" $5 ":" thisHome ":" $7}')
+groupEntry=$(getent group $thisGroup)
+
+# workaround case where Unix account is not in the local system (e.g. sssd)
+[[ -d $HOST_MOUNT_DIR/admin/etc/ ]] || (mkdir -p $HOST_MOUNT_DIR/admin/etc) || exit $?
+[[ -f $HOST_MOUNT_DIR/admin/etc/passwd ]] || {
+    echo "Creating passwd file"
+    getent passwd > $HOST_MOUNT_DIR/admin/etc/passwd
+    echo $passwdEntry >> $HOST_MOUNT_DIR/admin/etc/passwd
+}
+[[ -f $HOST_MOUNT_DIR/admin/etc/group ]] || {
+    echo "Creating group file"
+    getent group > $HOST_MOUNT_DIR/admin/etc/group
+    echo $groupEntry >> $HOST_MOUNT_DIR/admin/etc/group
+}
 
 [[ -d $HOST_MOUNT_DIR/certs ]] || mkdir -p $HOST_MOUNT_DIR/certs || exit $?
 [[ -d $HOST_MOUNT_DIR/admin/couchdb ]] || mkdir -p $HOST_MOUNT_DIR/admin/couchdb || exit $?
@@ -63,14 +79,13 @@ HOST_MOUNT_DIR=/data/dockerMount
 [[ -d $HOST_MOUNT_DIR/srv/couchdb/$COUCH_TAG/install/database ]] || mkdir -p $HOST_MOUNT_DIR/srv/couchdb/$COUCH_TAG/install/database || exit $?
 [[ -d $HOST_MOUNT_DIR/srv/couchdb/$COUCH_TAG/logs ]] || mkdir -p $HOST_MOUNT_DIR/srv/couchdb/$COUCH_TAG/logs || exit $?
 
-# sudo chown -R $couchdbUser:$couchdbUser $HOST_MOUNT_DIR/srv/couchdb/$COUCH_TAG
 
 dockerOpts="
 --detach \
 --network=host \
 --rm \
---user $UID:`id -g` \
---hostname=`hostname -f` \
+--hostname=$(hostname -f) \
+--user $(id -u):$(id -g) \
 --name=couchdb \
 --mount type=bind,source=/tmp,target=/tmp \
 --mount type=bind,source=$HOST_MOUNT_DIR/certs,target=/data/certs \
@@ -78,17 +93,14 @@ dockerOpts="
 --mount type=bind,source=$HOST_MOUNT_DIR/srv/couchdb/$COUCH_TAG/logs,target=/data/srv/couchdb/current/logs \
 --mount type=bind,source=$HOST_MOUNT_DIR/admin/wmagent,target=/data/admin/wmagent/ \
 --mount type=bind,source=$HOST_MOUNT_DIR/admin/couchdb,target=/data/admin/couchdb/ \
---mount type=bind,source=/etc/group,target=/etc/group,readonly \
---mount type=bind,source=/etc/passwd,target=/etc/passwd,readonly \
---mount type=bind,source=/etc/shadow,target=/etc/shadow,readonly \
+--mount type=bind,source=$HOST_MOUNT_DIR/admin/etc/passwd,target=/etc/passwd,readonly \
+--mount type=bind,source=$HOST_MOUNT_DIR/admin/etc/group,target=/etc/group,readonly \
 --mount type=bind,source=/etc/sudoers,target=/etc/sudoers,readonly \
 --mount type=bind,source=/etc/sudoers.d,target=/etc/sudoers.d,readonly \
---mount type=bind,source=/etc/profile,target=/etc/profile,readonly \
 "
 
 # --mount type=bind,source=$HOST_MOUNT_DIR/srv/couchdb/$COUCH_TAG/config,target=/data/srv/couchdb/current/config \
-# couchOpts=$*
-# couchOpts="$couchOpts --user couchdb -e COUCHDB_USER=TestAdmin -e COUCHDB_PASSWORD=TestPass"
+
 registry=local
 repository=wmagent-couchdb
 
@@ -101,7 +113,7 @@ $PULL && {
     docker tag  $registry/$project/$repository:$COUCH_TAG $registry/$repository:latest
 }
 
-echo "Starting the couchdb:$COUCH_TAG docker container with the following parameters: $couchOpts"
-docker run $dockerOpts $couchOpts $registry/$repository:$COUCH_TAG && (
+echo "Starting couchdb:$COUCH_TAG docker container with user: $thisUser:$thisGroup"
+docker run $dockerOpts $registry/$repository:$COUCH_TAG && (
     [[ -h $HOST_MOUNT_DIR/srv/couchdb/current ]] && rm -f $HOST_MOUNT_DIR/srv/couchdb/current
     ln -s $HOST_MOUNT_DIR/srv/couchdb/$COUCH_TAG $HOST_MOUNT_DIR/srv/couchdb/current )
--- a/docker/pypi/wmagent-couchdb/run.sh
+++ b/docker/pypi/wmagent-couchdb/run.sh
@@ -1,60 +1,36 @@
 #!/bin/bash
 
+# Basic initialization for CouchDB
+thisUser=$(id -un)
+thisGroup=$(id -gn)
+thisUserID=$(id -u)
+thisGroupID=$(id -g)
+echo "Running CouchDB container with user: $thisUser (ID: $thisUserID) and group: $thisGroup (ID: $thisGroupID)"
+
+export USER=$thisUser
+[[ -d ${HOME} ]] || mkdir -p ${HOME}
+
+<<EOF cat >> ~/.bashrc
+export USER=$thisUser
+
+alias lll="ls -lathr"
+alias ls="ls --color=auto"
+alias ll='ls -la --color=auto'
+alias scurl='curl -k --cert ${COUCH_CERTS_DIR}/servicecert.pem --key ${COUCH_CERTS_DIR}/servicekey.pem'
+
+alias manage=$COUCH_MANAGE_DIR/manage
+
+# Set command prompt for the running user inside the container
+export PS1="(CouchDB-$TAG) [\u@\h:\W]\$ "
+EOF
+source ${HOME}/.bashrc
+
 manage init      | tee -a $COUCH_LOG_DIR/run.log
 manage start     | tee -a $COUCH_LOG_DIR/run.log
 manage pushapps  | tee -a $COUCH_LOG_DIR/run.log
 
 echo "start sleeping....zzz"
-while true; do sleep 10; done
-
+sleep infinity
 
 # # start the service
 # manage start
-
-
-
-
-
-# ###########################################################################################
-# # NOTE: Leftovers - to be adopted/reimplemented in the GH issue dealing with CouchDB setup
-# #       all of those steps were previously done with the old wmagent deployment procedures
-# ###########################################################################################
-
-# DATA_SIZE=`lsblk -bo SIZE,MOUNTPOINT | grep ' /data1' | sort | uniq | awk '{print $1}'`
-# DATA_SIZE_GB=`lsblk -o SIZE,MOUNTPOINT | grep ' /data1' | sort | uniq | awk '{print $1}'`
-# if [[ $DATA_SIZE -gt 200000000000 ]]; then  # greater than ~200GB
-# echo "Partition /data1 available! Total size: $DATA_SIZE_GB"
-# sleep 0.5
-# while true; do
-# read -p "Would you like to deploy couchdb in this /data1 partition (yes/no)? " yn
-# case $yn in
-# [Y/y]* ) DATA1=true; break;;
-# [N/n]* ) DATA1=false; break;;
-# * ) echo "Please answer yes or no.";;
-# esac
-# done
-# else
-# DATA1=false
-# fi && echo
-
-# echo -e "\n*** Applying (for couchdb1.6, etc) cert file permission ***"
-# chmod 600 /data/certs/service{cert,key}.pem
-# echo "Done!"
-
-# echo "*** Checking if couchdb migration is needed ***"
-# echo -e "\n[query_server_config]\nos_process_limit = 50" >> $WMA_CURRENT_DIR/config/couchdb/local.ini
-# if [ "$DATA1" = true ]; then
-# ./manage stop-services
-# sleep 5
-# if [ -d "/data1/database/" ]; then
-# echo "Moving old database away... "
-# mv /data1/database/ /data1/database_old/
-# FINAL_MSG="5) Remove the old database when possible (/data1/database_old/)"
-# fi
-# rsync --remove-source-files -avr /data/srv/wmagent/current/install/couchdb/database /data1
-# sed -i "s+database_dir = .*+database_dir = /data1/database+" $WMA_CURRENT_DIR/config/couchdb/local.ini
-# sed -i "s+view_index_dir = .*+view_index_dir = /data1/database+" $WMA_CURRENT_DIR/config/couchdb/local.ini
-# ./manage start-services
-# fi
-# echo "Done!" && echo
-# ###########################################################################################
--- a/docker/pypi/wmagent-mariadb/Dockerfile
+++ b/docker/pypi/wmagent-mariadb/Dockerfile
@@ -32,10 +32,6 @@ ENV MDB_SECRETS_FILE=$MDB_ADMIN_DIR/MariaDB.secrets
 ENV WMA_SECRETS_FILE=$WMA_ADMIN_DIR/WMAgent.secrets
 ENV WMA_DATABASE=wmagent
 
-# create the system user to run the database
-RUN groupadd -g 1399 zh
-RUN useradd -u 31961  -g 1399 -G 999 -m cmst1
-
 # start the setup
 RUN mkdir -p $MDB_ROOT_DIR $MDB_CURRENT_DIR $MDB_CONFIG_DIR $MDB_MANAGE_DIR \
     $MDB_LOG_DIR $MDB_DATABASE_DIR $MDB_STATE_DIR $MDB_AUTH_DIR
@@ -58,14 +54,11 @@ alias manage=$MDB_MANAGE_DIR/manage
 export PS1="(MariaDB-$MDB_TAG) [\u@\h:\W]\$([[ \$(id -u) -eq 0 ]] && echo \# || echo \$) "
 EOF
 
-RUN <<EOF cat >>/home/cmst1/.bashrc
+RUN <<EOF cat >> ~/.bashrc
 alias manage=$MDB_MANAGE_DIR/manage
 export PS1="(MariaDB-$MDB_TAG) [\u@\h:\W]\$([[ \$(id -u) -eq 0 ]] && echo \# || echo \$) "
 EOF
 
-# RUN chown -R ${USER} ${MDB_ROOT_DIR}
-
 # setup final environment
-# USER $USER
 WORKDIR $MDB_ROOT_DIR
 ENTRYPOINT ["./run.sh", "2>&1"]
--- a/docker/pypi/wmagent-mariadb/Dockerfile
+++ b/docker/pypi/wmagent-mariadb/Dockerfile
@@ -48,16 +48,13 @@ ADD my.cnf ${MDB_CONFIG_DIR}/my.cnf
 
 ENV PATH="/usr/local/bin/:${MDB_ROOT_DIR}:${PATH}"
 
-# set MariaDB docker specific bash prompt and manage alias for all users:
-RUN <<EOF cat >>/root/.bashrc
-alias manage=$MDB_MANAGE_DIR/manage
-export PS1="(MariaDB-$MDB_TAG) [\u@\h:\W]\$([[ \$(id -u) -eq 0 ]] && echo \# || echo \$) "
+# Set command prompt for root
+RUN <<EOF cat >> /root/.bashrc
+export PS1="(MariaDB-$MDB_TAG) [\u@\h:\W]# "
 EOF
 
-RUN <<EOF cat >> ~/.bashrc
-alias manage=$MDB_MANAGE_DIR/manage
-export PS1="(MariaDB-$MDB_TAG) [\u@\h:\W]\$([[ \$(id -u) -eq 0 ]] && echo \# || echo \$) "
-EOF
+# allow dynamic users to create homefolders and .bashrc
+RUN chmod 777 /home
 
 # setup final environment
 WORKDIR $MDB_ROOT_DIR
--- a/docker/pypi/wmagent-mariadb/mariadb-docker-run.sh
+++ b/docker/pypi/wmagent-mariadb/mariadb-docker-run.sh
@@ -49,13 +49,29 @@ while getopts ":t:hp" opt; do
     esac
 done
 
-
-mariadbUser=`id -un`
-mariadbOpts=" --user $mariadbUser -e USER=$mariadbUser"
-
 # This is the root at the host only, it may differ from the root inside the container.
 # NOTE: this may be parametriesed, so that the container can run on a different mount point.
 HOST_MOUNT_DIR=/data/dockerMount
+
+thisUser=$(id -un)
+thisGroup=$(id -gn)
+
+# create the passwd and group mount point dynamically at runtime
+passwdEntry=$(getent passwd $thisUser | awk -F : -v thisHome="/home/$thisUser" '{print $1 ":" $2 ":" $3 ":" $4 ":" $5 ":" thisHome ":" $7}')
+groupEntry=$(getent group $thisGroup)
+
+# workaround case where Unix account is not in the local system (e.g. sssd)
+[[ -d $HOST_MOUNT_DIR/admin/etc/ ]] || (mkdir -p $HOST_MOUNT_DIR/admin/etc) || exit $?
+[[ -f $HOST_MOUNT_DIR/admin/etc/passwd ]] || {
+    echo "Creating passwd file"
+    getent passwd > $HOST_MOUNT_DIR/admin/etc/passwd
+    echo $passwdEntry >> $HOST_MOUNT_DIR/admin/etc/passwd
+}
+[[ -f $HOST_MOUNT_DIR/admin/etc/group ]] || {
+    echo "Creating group file"
+    getent group > $HOST_MOUNT_DIR/admin/etc/group
+    echo $groupEntry >> $HOST_MOUNT_DIR/admin/etc/group
+}
 
 [[ -d $HOST_MOUNT_DIR/certs ]] || (mkdir -p $HOST_MOUNT_DIR/certs) || exit $?
 [[ -d $HOST_MOUNT_DIR/admin/mariadb ]] || (mkdir -p $HOST_MOUNT_DIR/admin/mariadb) || exit $?
@@ -63,14 +79,13 @@ HOST_MOUNT_DIR=/data/dockerMount
 [[ -d $HOST_MOUNT_DIR/srv/mariadb/$MDB_TAG/install/database ]] || { mkdir -p $HOST_MOUNT_DIR/srv/mariadb/$MDB_TAG/install/database ;} || exit $?
 [[ -d $HOST_MOUNT_DIR/srv/mariadb/$MDB_TAG/logs ]] || { mkdir -p $HOST_MOUNT_DIR/srv/mariadb/$MDB_TAG/logs ;} || exit $?
 
-# sudo chown -R $mariadbUser $HOST_MOUNT_DIR/srv/mariadb/$MDB_TAG
 
 dockerOpts="
 --detach \
 --network=host \
 --rm \
---user $UID:`id -g` \
---hostname=`hostname -f` \
+--hostname=$(hostname -f) \
+--user $(id -u):$(id -g) \
 --name=mariadb \
 --mount type=bind,source=/tmp,target=/tmp \
 --mount type=bind,source=$HOST_MOUNT_DIR/certs,target=/data/certs \
@@ -78,12 +93,10 @@ dockerOpts="
 --mount type=bind,source=$HOST_MOUNT_DIR/srv/mariadb/$MDB_TAG/logs,target=/data/srv/mariadb/current/logs \
 --mount type=bind,source=$HOST_MOUNT_DIR/admin/mariadb,target=/data/admin/mariadb/ \
 --mount type=bind,source=$HOST_MOUNT_DIR/admin/wmagent,target=/data/admin/wmagent/ \
---mount type=bind,source=/etc/group,target=/etc/group,readonly \
---mount type=bind,source=/etc/passwd,target=/etc/passwd,readonly \
---mount type=bind,source=/etc/shadow,target=/etc/shadow,readonly \
+--mount type=bind,source=$HOST_MOUNT_DIR/admin/etc/passwd,target=/etc/passwd,readonly \
+--mount type=bind,source=$HOST_MOUNT_DIR/admin/etc/group,target=/etc/group,readonly \
 --mount type=bind,source=/etc/sudoers,target=/etc/sudoers,readonly \
 --mount type=bind,source=/etc/sudoers.d,target=/etc/sudoers.d,readonly \
---mount type=bind,source=/etc/profile,target=/etc/profile,readonly \
 "
 
 # --mount type=bind,source=$HOST_MOUNT_DIR/srv/mariadb/$MDB_TAG/config,target=/data/srv/mariadb/current/config \
@@ -101,7 +114,7 @@ $PULL && {
     docker tag  $registry/$project/$repository:$MDB_TAG $registry/$repository:latest
 }
 
-echo "Starting the $registry/$repository:$MDB_TAG docker container with the following parameters: $mariadbOpts"
-docker run $dockerOpts $mariadbOpts $registry/$repository:$MDB_TAG && (
+echo "Starting $repository:$MDB_TAG docker container with user: $thisUser:$thisGroup"
+docker run $dockerOpts $registry/$repository:$MDB_TAG && (
     [[ -h $HOST_MOUNT_DIR/srv/mariadb/current ]] && rm -f $HOST_MOUNT_DIR/srv/mariadb/current
     ln -s $HOST_MOUNT_DIR/srv/mariadb/$MDB_TAG $HOST_MOUNT_DIR/srv/mariadb/current )
--- a/docker/pypi/wmagent-mariadb/mariadb-docker-run.sh
+++ b/docker/pypi/wmagent-mariadb/mariadb-docker-run.sh
@@ -69,6 +69,7 @@ dockerOpts="
 --detach \
 --network=host \
 --rm \
+--user $UID:`id -g` \
 --hostname=`hostname -f` \
 --name=mariadb \
 --mount type=bind,source=/tmp,target=/tmp \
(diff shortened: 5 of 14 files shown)


amaltaro:
I tested these changes and they still do not work:
```
cmst1@vocms0265:srv $ ./wmagent-docker-run.sh -t 2.3.2rc8-alan2
ln: failed to create symbolic link ‘/data/srv/wmagent/wmagent’: File exists
Checking if there is no other wmagent container running and creating a link to the 2.3.2rc8-alan2 in the host mount area.
Starting the wmagent:2.3.2rc8-alan2 docker container with the following user parameter:  --user xxx:yyy
a1ddd62b636b304e03ccae77fda3352187ff272a171d47f2971ce5a1b6598461
cmst1@vocms0265:srv $ docker logs a1ddd62b636b304e03ccae77fda3352187ff272a171d47f2971ce5a1b6598461
id: cannot find name for user ID xxx
id: cannot find name for group ID yyy
Running WMAgent container with user: xxx (ID: xxx) and group: yyy (ID: yyy)
```

amaltaro:
Now that I am using the right script to start the container, I can confirm that the dynamic user is working fine both at CERN and FNAL. If anyone wants to give this a try, please use image `cmsweb/wmagent:2.3.2rc8-alan5` from the CERN registry.

There is one apparently unrelated issue when running this on FNAL though, the mount points are being changed to be owned by root, for instance, this is inside the container:
```
cmsdataops@cmssrvxxx:/data$ ls -lt
total 48
lrwxrwxrwx 1 root root    40 Apr 25 01:26 manage -> /data/srv/wmagent/2.3.2rc8/config/manage
drwxr-xr-x 3 root root    21 Apr 25 01:26 admin
drwxr-xr-x 3 root root    21 Apr 25 01:26 srv
```

which makes it read the default secrets file and not copy any of the certs. To be investigated further tomorrow.

docker/pypi/wmagent-couchdb/Dockerfile
```diff
@@ -92,9 +68,17 @@ alias manage=$COUCH_MANAGE_DIR/manage
...
 EOF
 
-RUN chown -R ${USER}:${GROUP} ${COUCH_ROOT_DIR}
+# set CouchDB docker specific bash prompt and manage alias for all users:
+RUN <<EOF cat >>/root/.bashrc
+alias manage=$COUCH_MANAGE_DIR/manage
+export PS1="(CouchDB-$TAG) [\u@\h:\W]\$([[ \$(id -u) -eq 0 ]] && echo \# || echo \$) "
+EOF
+
+RUN <<EOF cat >> ~/.bashrc
```
todor-ivanov, line 77: I'll be pretty surprised if I see this to be properly propagated to the correct runtime user. It will go into the home directory of the default user which comes with the couchdb upstream image. 

I am really concerned about those comand prompts, and as long as we have the init.sh scirpt for the WMAgent container, I am calm for the wmagent (we can set those there and forget), but for mariadb and couchdb the closest thing we have is the runt.sh script, but this means reediting `~/.bashrc` constantly on every restart - not the best practice, to my opinion.

docker/pypi/wmagent/run.sh
```diff
@@ -1,11 +1,23 @@
...
 ### Basic initialization wrapper for WMAgent to serve as the main entry point for the WMAgent Docker container
+wmaUser=$(id -un)
+wmaGroup=$(id -gn)
+wmaUserID=$(id -u)
+wmaGroupID=$(id -g)
+export WMA_USER=$wmaUser
+echo "Running WMAgent container with user: $wmaUser (ID: $wmaUserID) and group: $wmaGroup (ID: $wmaGroupID)"
 
+echo "Correcting ownership for WMA_ROOT_DIR: $WMA_ROOT_DIR"
+find $WMA_ROOT_DIR \! \( -user $wmaUser -group $wmaGroup \) -exec chown -f $wmaUser:$wmaGroup '{}' +;
```
todor-ivanov, line 12: @amaltaro I have two questions regarding this line:

* why this complicate command instead just:
```
chown -R $wmaUser:$wmaGroup $WMA_ROOT_DIR
```
* why do we actually need this. The whole $WMA_ROOT_DIR is expected to be created with the proper user permissions outside the container. In our case it is done here:  https://github.com/dmwm/CMSKubernetes/blob/f4a43f8a85bb1be3b59b913b59914f5f66014a34/docker/pypi/wmagent/wmagent-docker-run.sh#L59-L63 
And if it is not a `sudo` command it will break and do nothing. 
 
Just to clarify, in the past I have been seeing many times the mount  dir having files with root privileges, and in case this was the driving force behind this command to be moved here, I am pretty sure this is not the right solution. Because those files show up with `root:root` ownership, only when the container was wrongly run and have not switched to the proper user. Then docker engine is creating what is needed for us to run, and docker engine runs with `root`. This was  something which took me  quite a while in the past to block happening.  What we need to do is to get assured that the container is about to be executed at the agent with the same user every time.  

In addition for T0 their account is not sudo, so for them this was failing all the time. (I mean, I've walked this path already)

docker/pypi/wmagent/wmagent-docker-run.sh
```diff
@@ -55,37 +55,61 @@ HOST_MOUNT_DIR=/data/dockerMount
...
+
+# create regular mount points at runtime
 [[ -d $HOST_MOUNT_DIR/certs ]] || (mkdir -p $HOST_MOUNT_DIR/certs) || exit $?
 [[ -d $HOST_MOUNT_DIR/admin/wmagent ]] || (mkdir -p $HOST_MOUNT_DIR/admin/wmagent) || exit $?
 [[ -d $HOST_MOUNT_DIR/srv/wmagent/$WMA_TAG/install ]] || (mkdir -p $HOST_MOUNT_DIR/srv/wmagent/$WMA_TAG/install) || exit $?
 [[ -d $HOST_MOUNT_DIR/srv/wmagent/$WMA_TAG/config  ]] || (mkdir -p $HOST_MOUNT_DIR/srv/wmagent/$WMA_TAG/config)  || exit $?
 [[ -d $HOST_MOUNT_DIR/srv/wmagent/$WMA_TAG/logs ]] || { mkdir -p $HOST_MOUNT_DIR/srv/wmagent/$WMA_TAG/logs ;} || exit $?
 
-chown -R $wmaUser $HOST_MOUNT_DIR/srv/wmagent/$WMA_TAG || exit $?
+# Check we own everything in $HOST_MOUNT_DIR/srv/wmagent/$WMA_TAG and $HOST_MOUNT_DIR/etc
```
todor-ivanov, line 82: Same comment as above. What is the advantage of searching for all files not having the proper privileges, instead of recursively changing the ownership of the parent directory.  

Again this chown command here was doing nothing even before this change here, I myself was having in my mind the action item of removing it. And the reasons are  again the same: 
* if it is not `sudo` it will break and do nothing.
* if it is actually needed, it means only one thing someone did the mistake to run the container with the wrong user previously, so someone needs to go  there and clean the mess manually (only once, upon the mistake, and not on a regular basis)

docker/pypi/wmagent/run.sh
```diff
@@ -1,11 +1,23 @@
...
+
+# append the WMAgent user to the mysql group
+if getent passwd mysql >/dev/null 2>&1; then
```
todor-ivanov, line 15: Why needed ??????? 
I am pretty sure everything was working before without the need of having the extra step of appending the `WMA__USER`  to the `mysql` group. We are not using sockets inside the container. We are using the standard client, so always connecting to the mysql database through the network interface and the loopback device. If this is needed here, it is a sign to me only for one thing - someone is trying to connect to a local database at the host (possibly a databse from an old agent) through a socket (possibly exposed at the host from the old agent). Which is genuinely wrong. We should not even try this.

And more in very near future we will move to `mariadb` aliases instead of the `mysql` old ones. including the system groups.

docker/pypi/wmagent-couchdb/Dockerfile
todor-ivanov: Ok now,

So, @vkuznet  @amaltaro, I tested few things and here is the final solution to this annoying problem with the command  prompts for user agnostic containers. It was quite trivial at the end: 
https://github.com/dmwm/CMSKubernetes/pull/1467/commits/c7fec388edcc7c26b0a739411065a1465e6172d6

And if a dare to say.... we should expand this and move to `run.sh` all tweaks related to the `~./bashrc` file of the runtime estimated  user. I am talking about all those in particular: 
https://github.com/dmwm/CMSKubernetes/blob/f4a43f8a85bb1be3b59b913b59914f5f66014a34/docker/pypi/wmagent/install.sh#L153-L185 

Or as Alan did here with the `set_cronjob` function, to enclose these all tweaks in a separate function and call it at the end of `init.sh`, this would achieve the same result and at the same time keep the `run.sh` script as clean as possible.

todor-ivanov: BTW, I started to reply in the CouchDB  container related changes when it came to those command prompts... but it is equally valid for the 3 types of containers. Please consider those arguments equally applied to the 3 of them. And the parts that relate the the `wmagent/install.sh` file of course are just an example of the same problem - i.e. not knowing at deploy time which would be the default user at runtime.

amaltaro: Noted Todor. I am making these changes and will be back with updates soon. Thank you.

docker/pypi/wmagent/run.sh
amaltaro: Indeed I saw directories being owned by root yesterday. Let me remove this code then.

docker/pypi/wmagent/run.sh
amaltaro: This logic is already in our dockerfile, see: https://github.com/dmwm/CMSKubernetes/blob/master/docker/pypi/wmagent/Dockerfile#L40

It might be something legacy then, when we considered running MariaDB inside the same image as of WMAgent. I am removing this as well.

todor-ivanov: yep.... that should be legacy.

amaltaro:
@todor-ivanov I think I addressed all of your comments. If you want, feel free to have another look into this PR.
Now I am going to propagate similar fixes to the MariaDB and CouchDB setup.

docker/pypi/wmagent/etc/wmagent_bashrc
```diff
@@ -0,0 +1,25 @@
...
+
+alias runningagent="ps aux | egrep 'couch|wmcore|mysql|beam'"
+alias foldersize="du -h --max-depth=1 | sort -hr"
+
+# Better curl command
+alias scurl='curl -k --cert ${WMA_CERTS_DIR}/servicecert.pem --key ${WMA_CERTS_DIR}/servicekey.pem'
+
+# Set command prompt for the running user inside the container
+export PS1="(WMAgent-\$WMA_TAG) [\u@\h:\W]\$ "
+
```
todor-ivanov, line 20: It is a good way to load this from a file (me likes) , but I feel a line or two are missing. I thought while we were working with Andrea on the opportunistic resources we have added the line: 
```
source $WMA_DEPLOY_DIR/bin/manage-common.sh
```
which is quite helpful, because exposes  all `_*` auxiliary functions defined there, and makes it quite easy during operations to manually call stuff like `_print_settings`, _load_wmagentsecrets` , `_renew_proxy`

docker/pypi/wmagent/wmagent-docker-run.sh
amaltaro: It looks like I have to remove this chown'ing from this file. Will do it later.

docker/pypi/wmagent/etc/wmagent_bashrc
todor-ivanov: and indeed, @amaltaro, those two very important lines are missing:  https://github.com/dmwm/CMSKubernetes/pull/1462/files#diff-fcb5388aefac16c54c533239a64588ccb9bec26a5ab1d818a9d190830144a239R178-R179 

```
source $WMA_ENV_FILE
source $WMA_DEPLOY_DIR/bin/manage-common.sh
```

Could you please add them at the end of the `~/.bashrc` file.

amaltaro: Fixed! Thank you for catching this, Todor.

docker/pypi/wmagent/Dockerfile
```diff
@@ -43,15 +43,9 @@ RUN ${WMA_ROOT_DIR}/install.sh -t ${WMA_TAG}
 # Switch to the runtime directory
 WORKDIR ${WMA_ROOT_DIR}
 
-# Set WMAgent docker specific bash prompt and manage alias for all users:
+# Set command prompt for root
 RUN <<EOF cat >> /root/.bashrc
-alias manage=$WMA_MANAGE_DIR/manage
```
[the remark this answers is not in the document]
amaltaro: Assuming that we should not be using root for these containers, not having manage in the path might be even a protection. So I am in favor of leaving it out.

docker/pypi/wmagent-mariadb/Dockerfile
```diff
@@ -52,20 +48,14 @@ ADD my.cnf ${MDB_CONFIG_DIR}/my.cnf
...
-alias manage=$MDB_MANAGE_DIR/manage
-export PS1="(MariaDB-$MDB_TAG) [\u@\h:\W]\$([[ \$(id -u) -eq 0 ]] && echo \# || echo \$) "
+# Set command prompt for root
+RUN <<EOF cat >> /root/.bashrc
+export PS1="(MariaDB-$MDB_TAG) [\u@\h:\W]# "
 EOF
 
-# RUN chown -R ${USER} ${MDB_ROOT_DIR}
+# allow dynamic users to create homefolders and .bashrc 
+RUN chmod 777 /home
```
amaltaro, line 57: I am happy to hear about better alternatives to be able to create a home folder at run time.
Another alternative would be to define it at the host, and pass to the container via mount point, the only problem is that we actually have variations of bashrc (and manage).

amaltaro:
After swapping the secrets in vocms0265, I managed to proceed with testing the MariaDB image. Nonetheless, it still fails with problems like:
```
cmst1@vocmsxxx:srv $ ./mariadb-docker-run.sh -t 10.6.5-alan8 -p && docker logs -f mariadb
...
mariadb-admin: connect to server at 'localhost' failed
error: 'Can't connect to local server through socket '/var/run/mysqld/mariadb.sock' (2)'
```

which I tried to resolve with [this commit](https://github.com/dmwm/CMSKubernetes/pull/1469/commits/88f4acce8fa226e6571264fa9146a44d08e355cb), without any success.

I do confirm though that the dynamic user is working fine, e.g.:
```
(MariaDB-10.6.5) [cmst1@vocmsxxx:data]$
```

Unless this MariaDB issue is trivial, I think it should be addressed in a different ticket.

amaltaro:
Similar comment applies to the CouchDB image, the user setup is now dynamic, but it caused permission issues in different places. E.g.:
```
cmst1@vocms0265:srv $ ./couchdb-docker-run.sh -t 3.2.2-alan8 -p && docker logs -f couchdb
...
Running CouchDB container with user: xxx (ID: xxx) and group: yyy (ID: yyy)
init_couchdb: Adding a new account: cmst1
sed: couldn't open temporary file /data/srv/couchdb/3.2.2/config/sedPJzZyC: Permission denied
...
mkdir: cannot create directory '/data/srv/couchdb/3.2.2/install/stagingarea': Permission denied
/data/manage: line 403: pushd: /data/srv/couchdb/3.2.2/install/stagingarea/tmp: No such file or directory
```

docker/pypi/wmagent-couchdb/run.sh
```diff
@@ -1,60 +1,34 @@
 #!/bin/bash
 
+# Basic initialization for CouchDB
+thisUser=$(id -un)
+thisGroup=$(id -gn)
+thisUserID=$(id -u)
+thisGroupID=$(id -g)
+echo "Running CouchDB container with user: $thisUser (ID: $thisUserID) and group: $thisGroup (ID: $thisGroupID)"
+
+if ! [ -d ${HOME} ]; then
```
todor-ivanov, line 10: Please remove `if` `then` `!` `fi` .. all this heavy syntactic stuff and extra negations makes things unreadable. Just substitute with:

```
[[ -d ${HOME} ]] ||  mkdir -p ${HOME}
```

docker/pypi/wmagent-mariadb/manage
```diff
@@ -101,12 +101,12 @@ version(){
 
 
 init_mariadb(){
-    # The function to set/check initial database configurations and user preveleges
```
todor-ivanov, line 104: Why did you remove this check and substitute it with the one from above. They do check two different things. 
* The first one checks if the current user is the admin for the database
* The second one checks if somebody has configured one and the same admin user to serve as  admin and databse user.
 
We've have had this conversation in the past. It was long. It was decided to stay with the proper separation between users to drive the database with root privileges and a user to be connecting from outside as a user of  just for the database. Please, lets not mess up with this part, in this commits and changes.

docker/pypi/wmagent-mariadb/manage
```diff
@@ -149,17 +149,10 @@ init_mariadb(){
...
-        echo "$FUNCNAME: WARNING: WMAgent user set is the same as the MariaDB Root user. You must configure a different one!"
-        echo "$FUNCNAME: WARNING: NOT creating WMAgent users and NOT granting priveleges to $WMA_DATABASE database"
-        return 1
-    else
-        mariadb -u $MDB_ROOT --password=$MDB_ROOTPASS --socket=$MDB_SOCKET_FILE --execute "CREATE USER '${MDB_USER}'@'localhost' IDENTIFIED BY '$MDB_PASS'"
-        mariadb -u $MDB_ROOT --password=$MDB_ROOTPASS --socket=$MDB_SOCKET_FILE --execute "GRANT ALL ON $WMA_DATABASE.* TO $MDB_USER@localhost"
-        mariadb -u $MDB_ROOT --password=$MDB_ROOTPASS --socket=$MDB_SOCKET_FILE --execute "CREATE USER '${MDB_USER}'@'127.0.0.1' IDENTIFIED BY '$MDB_PASS'"
-        mariadb -u $MDB_ROOT --password=$MDB_ROOTPASS --socket=$MDB_SOCKET_FILE --execute "GRANT ALL ON $WMA_DATABASE.* TO $MDB_USER@127.0.0.1"
-    fi
+    mariadb -u $MDB_ROOT --password=$MDB_ROOTPASS --socket=$MDB_SOCKET_FILE --execute "CREATE USER '${MDB_USER}'@'localhost' IDENTIFIED BY '$MDB_PASS'"
```
todor-ivanov, line 152: Please revert the changes  to the `init_mariadb()` function. They are absolutely not related to the current issue and there are no driving reasons for them to happen (neither direct nor indirect) in the fact that we are about to change the user inside the container to be estimated at runtime. On top of that, they are also in contradiction with the decision we have already made in that  regards  months ago. And they are about to break something which was already heavily tested and is working rock solid, just as expected.

docker/pypi/wmagent-mariadb/mariadb-docker-run.sh
```diff
@@ -49,34 +49,55 @@ while getopts ":t:hp" opt; do
...
+thisUser=$(id -un)
+thisGroup=$(id -gn)
+
+# create the passwd and group mount point dynamically at runtime
+passwdEntry=$(getent passwd $thisUser | awk -F : -v thisHome="/home/$thisUser" '{print $1 ":" $2 ":" $3 ":" $4 ":" $5 ":" thisHome ":" $7}')
+groupEntry=$(getent group $thisGroup)
+
+# workaround case where Unix account is not in the local system (e.g. sssd)
+[[ -d $HOST_MOUNT_DIR/etc/ ]] || (mkdir -p $HOST_MOUNT_DIR/etc) || exit $?
+if ! [ -f $HOST_MOUNT_DIR/etc/passwd ]; then
```
todor-ivanov, line 65: Same comment as the one related  to the couchdb caontainers

docker/pypi/wmagent-mariadb/mariadb-docker-run.sh
```diff
@@ -49,34 +49,55 @@ while getopts ":t:hp" opt; do
...
+--hostname=$(hostname -f) \
+--user $(id -u):$(id -g) \
 --name=mariadb \
 --mount type=bind,source=/tmp,target=/tmp \
 --mount type=bind,source=$HOST_MOUNT_DIR/certs,target=/data/certs \
 --mount type=bind,source=$HOST_MOUNT_DIR/srv/mariadb/$MDB_TAG/install/database,target=/data/srv/mariadb/current/install/database \
 --mount type=bind,source=$HOST_MOUNT_DIR/srv/mariadb/$MDB_TAG/logs,target=/data/srv/mariadb/current/logs \
 --mount type=bind,source=$HOST_MOUNT_DIR/admin/mariadb,target=/data/admin/mariadb/ \
 --mount type=bind,source=$HOST_MOUNT_DIR/admin/wmagent,target=/data/admin/wmagent/ \
+--mount type=bind,source=$HOST_MOUNT_DIR/etc/passwd,target=/etc/passwd,readonly \
```
todor-ivanov, line 97: as said earlier: we should separate those mount points on a per container basis, so that we have proper isolation

docker/pypi/wmagent/bin/manage-common.sh
```diff
@@ -317,7 +317,7 @@ _renew_proxy(){
                 echo "$FUNCNAME: ERROR: Failed to renew expired myproxy"; return $(false) ;}
 
         # Stay safe and always change the service {cert,key} and myproxy mode here:
-        sudo chmod 400 $WMA_CERTS_DIR/*
+        chmod 400 $WMA_CERTS_DIR/*
```
todor-ivanov, line 320: good catch... this one I've forgot to remove long ago, and was driving me crazy for so long, to the point I was eager for all other changes to finish so that I can create a separate issue just to fix it.

docker/pypi/wmagent/run.sh
```diff
@@ -1,11 +1,20 @@
 #!/bin/bash
 
 ### Basic initialization wrapper for WMAgent to serve as the main entry point for the WMAgent Docker container
+wmaUser=$(id -un)
+wmaGroup=$(id -gn)
+wmaUserID=$(id -u)
+wmaGroupID=$(id -g)
+export WMA_USER=$wmaUser
```
todor-ivanov, line 8: didn't you have to create the `/home` folder for the user in the `run.sh` script for the wmagent container as well?

docker/pypi/wmagent/wmagent-docker-run.sh
```diff
@@ -55,37 +54,55 @@ HOST_MOUNT_DIR=/data/dockerMount
...
 ln -s $HOST_MOUNT_DIR/srv/wmagent /data/srv/wmagent
 
-
+# create the passwd and group mount point dynamically at runtime
+passwdEntry=$(getent passwd $wmaUser | awk -F : -v wmaHome="/home/$wmaUser" '{print $1 ":" $2 ":" $3 ":" $4 ":" $5 ":" wmaHome ":" $7}')
+groupEntry=$(getent group $wmaGroup)
+
+# workaround case where Unix account is not in the local system (e.g. sssd)
+[[ -d $HOST_MOUNT_DIR/etc/ ]] || (mkdir -p $HOST_MOUNT_DIR/etc) || exit $?
+if ! [ -f $HOST_MOUNT_DIR/etc/passwd ]; then
```
todor-ivanov, line 63: same comment as all over the place when it comes to single  conditioned  `if` expressions with extra negation and heavily overloaded syntax.

todor-ivanov (changes requested):
Thanks @amaltaro 

I have left  several comments inline, which I think they should be addressed. 

The fact that you experience this error:
> error: 'Can't connect to local server through socket '/var/run/mysqld/mariadb.sock' (2)'

Means that your mariaDB initialization has failed. could be also because of the changes of `init_mariadb` function or due to permission issues at the mount point. Never the less, it tells one thing - mariaDB is trying to connect through the default socket, which is `/var/run/mysqld/mariadb.sock` and not the one we've  configured in our `manage` script, which should be referred through the variable `$MDB_SOCKET_FILE` 

Similar comment applies to the couchdb container and  the permission issues there. I'll checkout  your branch and will test, to see from first hand what could be the reason behind them both. Have you checked if you observe similar issues with the wmagent container as well?

docker/pypi/wmagent/Dockerfile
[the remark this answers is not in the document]
todor-ivanov: ok

amaltaro:
@todor-ivanov Todor, about MariaDB, I think the secrets files were backwards in vocms0265. Can you please confirm it again to me:
* MDB_ROOT (under the new secrets file) is supposed to have a root account for running the service
* MDB_USER (under the traditional secrets file) is supposed to have an account for "using" the service (to my understanding, this is the same account that we use to run WMAgent container, even though it does not have to be).

In a general note, I would like to avoid the proliferation of file/mounting/copying/accounts between the host and the container(s). If we have a minor suspicious that things can change in the future, I would not add (safety) complexity at this moment for something that might or not ever happened.

Said that, creating 3 passwd and 3 groups for these services that will be doing exactly the same thing and consuming exactly the same file, is one of the changes that I would not make. IF we ever get to a point that we need their separation, we can then do that.

In addition, you have command line options to be passed to the services through the entrypoint. Can you please clarify whether you have used that so far? For what? And whether we (still) need them?

I really think we should keep this whole setup as thin as possible, unless we have strong evidence that railguards are better to be defined here and there.

docker/pypi/wmagent-couchdb/run.sh
amaltaro: Honestly, I didn't remember that the extra square brackets was a negation. It indeed looks better in the code, but to my brain the `if` and classical `!` negation symbol are easier to digest.

Nonetheless, I am going to update soon all these inconsistent if statements.

amaltaro: Now that I decided to think I managed to understand your suggestion.

amaltaro:
> Means that your mariaDB initialization has failed. could be also because of the changes of init_mariadb function or due to permission issues at the mount point. Never the less, it tells one thing - mariaDB is trying to connect through the default socket, which is /var/run/mysqld/mariadb.sock and not the one we've configured in our manage script, which should be referred through the variable $MDB_SOCKET_FILE

@todor-ivanov the manage script isn't defining this variable, it is only consumed over there.
It is actually defined here:
https://github.com/dmwm/CMSKubernetes/blob/master/docker/pypi/wmagent-mariadb/Dockerfile#L30

How do you recommend to proceed here? Mount (not as ro) `/var/run/mysqld` in the mariadb-docker-run.sh script?

docker/pypi/wmagent-couchdb/run.sh
todor-ivanov: No, the extra square brackets are not negation `[[]]` the `!` is the extra negation which is the confusing part. The `[[]]` are just a notion for a est condition, while  the `||` from the logical chain is the part that states we are about to follow if the condition inside the `[[]]` is broken (evaluates to `false`). But, if we wanted to react only in the cases when the condition is actually not broken but rather is satisfied  (evaluates to `true`) then we should proceed with an `&&`logical chain:

* `[[ $cond ]] || $(action)` - for acting on false conditions
* `[[ $cond ]] && $(action)` - for acting on true conditions

todor-ivanov:
> It is actually defined here:
https://github.com/dmwm/CMSKubernetes/blob/master/docker/pypi/wmagent-mariadb/Dockerfile#L30

You are correct. It was different before -trying to run from the host mount area. I changed it at some point, to match the default fro the container.

>  How do you recommend to proceed here? Mount (not as ro) /var/run/mysqld in the mariadb-docker-run.sh script?

No. We should not mount it from the host actually. The error tells me it is not able to start the server, because obviously the mariadb initialization has failed on an earlier stage. Most probably during the setup with `manage init-mariadb`

todor-ivanov:
hi @amaltaro, 

I just found the origin of your problems with the mariadb container... and I suspect it is the same for CouchDB as well: 

Look, from inside the container:
```
(MariaDB-10.6.5) [cmst1@vocms0290:data]$ manage init-mariadb
init_mariadb: ERROR: The current user does not match the MariaDB root user from /data/admin/mariadb/MariaDB.secrets.
init_mariadb: ERROR: Cannot continue. Exit...

```

And the reason is simple the `USER` variable is missing from the environment. If you export it in the `run.sh` it should be ok.  

```
(MariaDB-10.6.5) [cmst1@vocms0290:data]$ echo $USER
```

todor-ivanov:
@amaltaro,  indeed this is the fix to  our problems:

adding here: https://github.com/dmwm/CMSKubernetes/pull/1469/files#diff-29c3f6976cd381a4365a703aa3c3fa3ab3a34161a258a63380080371b035079cR8 

```
export USER=$thisUser
```

solves the problem with the container. It immediately lights up and shines:

```
cmst1@vocms0290:wmagent-mariadb $ ./mariadb-docker-run.sh -t 10.6.5 && docker logs -f mariadb
Starting mariadb:10.6.5 docker container with user: cmst1:zh
bedf64c73bb8a881d29ad0830dab24cdc15e49fd08dc88acd1dec040dfd10987
Running MariaDB container with user: cmst1 (ID: 31961) and group: zh (ID: 1399)
-------------------------------------------------------------------------
init_mariadb: Stopping any previously running mariadb server
mariadb-admin: connect to server at 'localhost' failed
error: 'Can't connect to local server through socket '/var/run/mysqld/mariadb.sock' (2)'
Check that mariadbd is running and that the socket: '/var/run/mysqld/mariadb.sock' exists!

-------------------------------------------------------------------------
init_mariadb: Trying to install system database with user: cmst1 (if it is not already present)
chown: cannot access '/usr/lib/mysql/plugin/auth_pam_tool_dir/auth_pam_tool': Permission denied
chown: changing ownership of '/usr/lib/mysql/plugin/auth_pam_tool_dir': Operation not permitted
2024-04-26 16:40:45 0 [Warning] You need to use --log-bin to make --expire-logs-days or --binlog-expire-logs-seconds work.
Couldn't set an owner to '/usr/lib/mysql/plugin/auth_pam_tool_dir/auth_pam_tool'. It must be root, the PAM authentication plugin doesn't work otherwise.. Cannot change ownership of the '/usr/lib/mysql/plugin/auth_pam_tool_dir' directory to the 'cmst1' user. Check that you have the necessary permissions and try again. Installing MariaDB/MySQL system tables in '/data/srv/mariadb/10.6.5/install/database' ... OK To start mysqld at boot time you have to copy support-files/mysql.server to the right place for your system Two all-privilege accounts were created. One is root@localhost, it has no password, but you need to be system 'root' user to connect. Use, for example, sudo mysql The second is cmst1@localhost, it has no password either, but you need to be the system 'cmst1' user to connect. After connecting you can set the password, if you would need to be able to connect as any of these users with a password and without sudo See the MariaDB Knowledgebase at https://mariadb.com/kb or the MySQL manual for more instructions. You can start the MariaDB daemon with: cd '/usr' ; /usr/bin/mysqld_safe --datadir='/data/srv/mariadb/10.6.5/install/database' You can test the MariaDB daemon with mysql-test-run.pl cd '/usr/mysql-test' ; perl mysql-test-run.pl Please report any problems at https://mariadb.org/jira The latest information about MariaDB is available at https://mariadb.org/. You can find additional information about the MySQL part at: https://dev.mysql.com Consider joining MariaDB's strong and vibrant community: https://mariadb.org/get-involved/
-------------------------------------------------------------------------
init_mariadb: Starting MariaDB server
start_mariadb: Starting MariaDB server
...
240426 16:40:51 mysqld_safe Can't log to error log and syslog at the same time.  Remove all --log-error configuration options for --syslog to take effect.
240426 16:40:51 mysqld_safe Logging to '/data/srv/mariadb/10.6.5/logs/error.log'.
240426 16:40:51 mysqld_safe Starting mariadbd daemon with databases from /data/srv/mariadb/10.6.5/install/database
mariadb-admin  Ver 9.1 Distrib 10.6.5-MariaDB, for debian-linux-gnu on x86_64
Copyright (c) 2000, 2018, Oracle, MariaDB Corporation Ab and others.

Server version		10.6.5-MariaDB-1:10.6.5+maria~focal-log
Protocol version	10
Connection		Localhost via UNIX socket
UNIX socket		/var/run/mysqld/mariadb.sock
Uptime:			10 sec

Threads: 2  Questions: 1  Slow queries: 0  Opens: 16  Open tables: 10  Queries per second avg: 0.100

-------------------------------------------------------------------------
init_mariadb: Securing MariaDB Root user

-------------------------------------------------------------------------
init_mariadb: Creating WMAgent databases
init_mariadb: Installing WMAgent Database: wmagent

-------------------------------------------------------------------------
init_mariadb: Creating WMAgent user and setting grants
-------------------------------------------------------------------------
start_mariadb: WARNING: MariaDB Server already running on --socket=/var/run/mysqld/mariadb.sock
Start sleeping....zzz
```


And later inside the container:
```
cmst1@vocms0290:wmagent-mariadb $ docker exec -it mariadb bash
(MariaDB-10.6.5) [cmst1@vocms0290:data]$ manage status 
mariadb-admin  Ver 9.1 Distrib 10.6.5-MariaDB, for debian-linux-gnu on x86_64
Copyright (c) 2000, 2018, Oracle, MariaDB Corporation Ab and others.

Server version		10.6.5-MariaDB-1:10.6.5+maria~focal-log
Protocol version	10
Connection		Localhost via UNIX socket
UNIX socket		/var/run/mysqld/mariadb.sock
Uptime:			3 min 27 sec

Threads: 2  Questions: 17  Slow queries: 0  Opens: 16  Open tables: 10  Queries per second avg: 0.082

```

like a star! 
Congrats!!!!!

amaltaro:
Thanks for looking into this. I still see a few errors in your log which make me wonder how the service managed to start.

The issue with CouchDB is different, given that it actually has no access to the CouchDB installation/work area.

todor-ivanov:
>  I still see a few errors in your log which make me wonder how the service managed to start

If you refer about this

```
init_mariadb: Stopping any previously running mariadb server
mariadb-admin: connect to server at 'localhost' failed
error: 'Can't connect to local server through socket '/var/run/mysqld/mariadb.sock' (2)'
Check that mariadbd is running and that the socket: '/var/run/mysqld/mariadb.sock' exists!
```

This is because the very first action during the init process is  to try to stop any already (previously) running mariaDB server. and it fails because there is none (since this is the first start of the container) . It is just a precaution in case  someone decides to rerun the init process later with `manage ini-mariadb` from inside the container.

amaltaro:
Not only those errors (thank you for clarifying that those are trying to stop anything that might be leftover), but these permission issues:
```
init_mariadb: Trying to install system database with user: cmst1 (if it is not already present)
chown: cannot access '/usr/lib/mysql/plugin/auth_pam_tool_dir/auth_pam_tool': Permission denied
chown: changing ownership of '/usr/lib/mysql/plugin/auth_pam_tool_dir': Operation not permitted
2024-04-26 16:40:45 0 [Warning] You need to use --log-bin to make --expire-logs-days or --binlog-expire-logs-seconds work.
Couldn't set an owner to '/usr/lib/mysql/plugin/auth_pam_tool_dir/auth_pam_tool'. It must be root, the PAM authentication plugin doesn't work otherwise.. Cannot change ownership of the '/usr/lib/mysql/plugin/auth_pam_tool_dir' directory to the 'cmst1' user. Check that you have the necessary permissions and try again. Installing MariaDB/MySQL system tables in '/data/srv/mariadb/10.6.5/install/database' ... OK To start mysqld at boot time you have to copy support-files/mysql.server to the right place for your system Two all-privilege accounts were created. One is root@localhost, it has no password, but you need to be system 'root' user to connect. Use, for example, sudo mysql The second is cmst1@localhost, it has no password either, but you need to be the system 'cmst1' user to connect. After connecting you can set the password, if you would need to be able to connect as any of these users with a password and without sudo See the MariaDB Knowledgebase at https://mariadb.com/kb or the MySQL manual for more instructions. You can start the MariaDB daemon with: cd '/usr' ; /usr/bin/mysqld_safe --datadir='/data/srv/mariadb/10.6.5/install/database' You can test the MariaDB daemon with mysql-test-run.pl cd '/usr/mysql-test' ; perl mysql-test-run.pl Please report any problems at https://mariadb.org/jira The latest information about MariaDB is available at https://mariadb.org/. You can find additional information about the MySQL part at: https://dev.mysql.com Consider joining MariaDB's strong and vibrant community: https://mariadb.org/get-involved/
```

I get those with `./mariadb-docker-run.sh -t 10.6.5-alan8 -p && docker logs -f mariadb` and the service does not start. 
The same happens if I `exec` into the container and try to initialize it manually, it fails out and MariaDB isn't started.

As suggested in the error message, cmst1 user has no permission to access that area inside the container:
```
(MariaDB-10.6.5) [cmst1@vocms0265:data]$ ls -lt /usr/lib/mysql/plugin/
total 1208
drwx------. 2 polkitd root     27 Feb  2  2022 auth_pam_tool_dir
```

I am puzzled to see it works for you(!) Do you have any difference? Can you please pull the image I mentioned in this log and try to run it in vocms0290?

todor-ivanov:
@amaltaro , 

Those errors:
```
init_mariadb: Trying to install system database with user: cmst1 (if it is not already present)
chown: cannot access '/usr/lib/mysql/plugin/auth_pam_tool_dir/auth_pam_tool': Permission denied
chown: changing ownership of '/usr/lib/mysql/plugin/auth_pam_tool_dir': Operation not permitted
2024-04-26 16:40:45 0 [Warning] You need to use --log-bin to make --expire-logs-days or --binlog-expire-logs-seconds work.
Couldn't set an owner to '/usr/lib/mysql/plugin/auth_pam_tool_dir/auth_pam_tool'. It must be root, the PAM authentication plugin doesn't work otherwise.. Cannot change ownership of the '/usr/lib/mysql/plugin/auth_pam_tool_dir' directory to the 'cmst1' user. Check that you have the necessary permissions and try again. Installing MariaDB/MySQL system tables in '/data/srv/mariadb/10.6.5/install/database' ... OK To start mysqld at boot time you have to copy support-files/mysql.server to the right place for your system Two all-privilege accounts were created. One is root@localhost, it has no password, but you need to be system 'root' user to connect. Use, for example, sudo mysql The second is cmst1@localhost, it has no password either, but you need to be the system 'cmst1' user to connect. After connecting you can set the password, if you would need to be able to connect as any of these users with a password and without sudo See the MariaDB Knowledgebase at https://mariadb.com/kb or the MySQL manual for more instructions. You can start the MariaDB daemon with: cd '/usr' ; /usr/bin/mysqld_safe --datadir='/data/srv/mariadb/10.6.5/install/database' You can test the MariaDB daemon with mysql-test-run.pl cd '/usr/mysql-test' ; perl mysql-test-run.pl Please report any problems at https://mariadb.org/jira The latest information about MariaDB is available at https://mariadb.org/. You can find additional information about the MySQL part at: https://dev.mysql.com Consider joining MariaDB's strong and vibrant community: https://mariadb.org/get-involved/

``` 

Are unrelated. 

We do not use pam to authenticate with the server. (I have not found an easy way to tell the server to not even try to start/use it). MariaDB is capable of running without this auth module just fine. If a client end up authenticating with pam module it means the `init_mariadb` has gone wrong and the manage script has fallen to create the proper set of passwords for the database for some reason. This can explain why the server fails to start, but not for the lack of the pam module. I've seen such things happening after many many tests with the same tag recompiled and changed multiple times and tested on the same machine. The origin of the problems are usually left overs from previously wrongly configured databases at the host. The solution is to completely delete `mariadb` directories at the host and rerun. 

BTW your image runs perfectly fine for me:
```
cmst1@vocms0290:wmagent-mariadb $ ./mariadb-docker-run.sh -t 10.6.5-alan8 -p && docker logs -f mariadb
Pulling Docker image: registry.cern.ch/cmsweb/mariadb:10.6.5-alan8
10.6.5-alan8: Pulling from cmsweb/mariadb
08c01a0ec47e: Already exists 
a2bcb14c13a1: Already exists 
29c56760f879: Already exists 
a95000a218fc: Already exists 
a765d76e68d9: Already exists 
c6945738f085: Already exists 
62787b7c58c5: Already exists 
d2987a30cfe4: Already exists 
cbc65983d8b5: Already exists 
db216f91595e: Already exists 
88e173126053: Already exists 
a391f67d1cb5: Already exists 
fddbf0017781: Already exists 
a33af69faa48: Pull complete 
0b1e6f515ba8: Pull complete 
356754afeb33: Pull complete 
e8e9fc7ca2de: Pull complete 
d86de9ab39d7: Pull complete 
e53d004c686f: Pull complete 
Digest: sha256:0826d5f7198ba36a8dca86e402546747e1288c816e65bd5492d73a46ba1deec9
Status: Downloaded newer image for registry.cern.ch/cmsweb/mariadb:10.6.5-alan8
registry.cern.ch/cmsweb/mariadb:10.6.5-alan8
Starting mariadb:10.6.5-alan8 docker container with user: cmst1:zh
ba697d3278cf38fb9a9b4ccee04087157cf3d0c403c62bf3df45d8a1f6e532cc
Running MariaDB container with user: cmst1 (ID: 31961) and group: zh (ID: 1399)
-------------------------------------------------------------------------
init_mariadb: Stopping any previously running mariadb server
mariadb-admin: connect to server at 'localhost' failed
error: 'Can't connect to local server through socket '/var/run/mysqld/mariadb.sock' (2)'
Check that mariadbd is running and that the socket: '/var/run/mysqld/mariadb.sock' exists!

-------------------------------------------------------------------------
init_mariadb: Trying to install system database with user: cmst1 (if it is not already present)
chown: cannot access '/usr/lib/mysql/plugin/auth_pam_tool_dir/auth_pam_tool': Permission denied
chown: changing ownership of '/usr/lib/mysql/plugin/auth_pam_tool_dir': Operation not permitted
2024-04-27  3:04:29 0 [Warning] You need to use --log-bin to make --expire-logs-days or --binlog-expire-logs-seconds work.
Couldn't set an owner to '/usr/lib/mysql/plugin/auth_pam_tool_dir/auth_pam_tool'. It must be root, the PAM authentication plugin doesn't work otherwise.. Cannot change ownership of the '/usr/lib/mysql/plugin/auth_pam_tool_dir' directory to the 'cmst1' user. Check that you have the necessary permissions and try again. Installing MariaDB/MySQL system tables in '/data/srv/mariadb/10.6.5/install/database' ... OK To start mysqld at boot time you have to copy support-files/mysql.server to the right place for your system Two all-privilege accounts were created. One is root@localhost, it has no password, but you need to be system 'root' user to connect. Use, for example, sudo mysql The second is cmst1@localhost, it has no password either, but you need to be the system 'cmst1' user to connect. After connecting you can set the password, if you would need to be able to connect as any of these users with a password and without sudo See the MariaDB Knowledgebase at https://mariadb.com/kb or the MySQL manual for more instructions. You can start the MariaDB daemon with: cd '/usr' ; /usr/bin/mysqld_safe --datadir='/data/srv/mariadb/10.6.5/install/database' You can test the MariaDB daemon with mysql-test-run.pl cd '/usr/mysql-test' ; perl mysql-test-run.pl Please report any problems at https://mariadb.org/jira The latest information about MariaDB is available at https://mariadb.org/. You can find additional information about the MySQL part at: https://dev.mysql.com Consider joining MariaDB's strong and vibrant community: https://mariadb.org/get-involved/
-------------------------------------------------------------------------
init_mariadb: Starting MariaDB server
start_mariadb: Starting MariaDB server
...
240427 03:04:36 mysqld_safe Can't log to error log and syslog at the same time.  Remove all --log-error configuration options for --syslog to take effect.
240427 03:04:36 mysqld_safe Logging to '/data/srv/mariadb/10.6.5/logs/error.log'.
240427 03:04:36 mysqld_safe Starting mariadbd daemon with databases from /data/srv/mariadb/10.6.5/install/database
mariadb-admin  Ver 9.1 Distrib 10.6.5-MariaDB, for debian-linux-gnu on x86_64
Copyright (c) 2000, 2018, Oracle, MariaDB Corporation Ab and others.

Server version		10.6.5-MariaDB-1:10.6.5+maria~focal-log
Protocol version	10
Connection		Localhost via UNIX socket
UNIX socket		/var/run/mysqld/mariadb.sock
Uptime:			10 sec

Threads: 2  Questions: 1  Slow queries: 0  Opens: 16  Open tables: 10  Queries per second avg: 0.100

-------------------------------------------------------------------------
init_mariadb: Securing MariaDB Root user

-------------------------------------------------------------------------
init_mariadb: Creating WMAgent databases
init_mariadb: Installing WMAgent Database: wmagent

-------------------------------------------------------------------------
init_mariadb: Creating WMAgent user and setting grants
-------------------------------------------------------------------------
start_mariadb: WARNING: MariaDB Server already running on --socket=/var/run/mysqld/mariadb.sock
Start sleeping....zzz

```

And from inside the container:

```
cmst1@vocms0290:wmagent-mariadb $ docker exec -it mariadb bash

(MariaDB-10.6.5) [cmst1@vocms0290:data]$ manage status 
mariadb-admin  Ver 9.1 Distrib 10.6.5-MariaDB, for debian-linux-gnu on x86_64
Copyright (c) 2000, 2018, Oracle, MariaDB Corporation Ab and others.

Server version		10.6.5-MariaDB-1:10.6.5+maria~focal-log
Protocol version	10
Connection		Localhost via UNIX socket
UNIX socket		/var/run/mysqld/mariadb.sock
Uptime:			14 min 52 sec

Threads: 2  Questions: 17  Slow queries: 0  Opens: 16  Open tables: 10  Queries per second avg: 0.019
```

docker/pypi/wmagent-couchdb/couchdb-docker-run.sh
```diff
@@ -49,39 +49,58 @@ while getopts ":t:hp" opt; do
...
+# create the passwd and group mount point dynamically at runtime
+passwdEntry=$(getent passwd $thisUser | awk -F : -v thisHome="/home/$thisUser" '{print $1 ":" $2 ":" $3 ":" $4 ":" $5 ":" thisHome ":" $7}')
+groupEntry=$(getent group $thisGroup)
+
+# workaround case where Unix account is not in the local system (e.g. sssd)
+[[ -d $HOST_MOUNT_DIR/etc/ ]] || (mkdir -p $HOST_MOUNT_DIR/etc) || exit $?
+[[ -f $HOST_MOUNT_DIR/etc/passwd ]] || {
+    echo "Creating passwd file"
+    getent passwd > $HOST_MOUNT_DIR/etc/passwd
+    echo $passwdEntry >> $HOST_MOUNT_DIR/etc/passwd
```
todor-ivanov, line 68: please move this under:
```
$HOST_MOUNT_DIR/couchdb/admin/etc/passwd
```

docker/pypi/wmagent-mariadb/mariadb-docker-run.sh
```diff
@@ -49,34 +49,54 @@ while getopts ":t:hp" opt; do
...
+# create the passwd and group mount point dynamically at runtime
+passwdEntry=$(getent passwd $thisUser | awk -F : -v thisHome="/home/$thisUser" '{print $1 ":" $2 ":" $3 ":" $4 ":" $5 ":" thisHome ":" $7}')
+groupEntry=$(getent group $thisGroup)
+
+# workaround case where Unix account is not in the local system (e.g. sssd)
+[[ -d $HOST_MOUNT_DIR/etc/ ]] || (mkdir -p $HOST_MOUNT_DIR/etc) || exit $?
+[[ -f $HOST_MOUNT_DIR/etc/passwd ]] || {
+    echo "Creating passwd file"
+    getent passwd > $HOST_MOUNT_DIR/etc/passwd
+    echo $passwdEntry >> $HOST_MOUNT_DIR/etc/passwd
```
todor-ivanov, line 68: please move this under 

```
$HOST_MOUNT_DIR/mariadb/admin/etc/passwd
```

docker/pypi/wmagent/wmagent-docker-run.sh
```diff
@@ -55,37 +54,55 @@ HOST_MOUNT_DIR=/data/dockerMount
...
+    echo "Creating passwd file"
+    getent passwd > $HOST_MOUNT_DIR/etc/passwd
+    echo $passwdEntry >> $HOST_MOUNT_DIR/etc/passwd
```
todor-ivanov, line 66: please move this under:
```
$HOST_MOUNT_DIR/admin/wmagent/etc/passwd
```

vkuznet:
There were lots of discussions in this PR, but let me give you insight of MariaDB (MySQL) installation. Once executable and libraries are installed someone must run initialization and configuration steps. These steps are required to setup MariaDB under specific account. For example this [blog](https://www.digitalocean.com/community/tutorials/how-to-install-mariadb-on-ubuntu-20-04) nicely describe them. In particular, we may need to pay attention to step 2, 3. Now, I think that described permissions problems caused by missing these steps. But here is a problem. We have executable and libraries and would like to run them under specific account which we know only at run-time. Therefore, the `run.sh` or `init.sh` scripts should perform these steps once we start container. If you'll try to install MariaDB on your own node you'll see them, and they can be performed differently. Basically someone will need to run `chown` on bunch of directories (logs, db, etc) and have proper settings of that account in /etc/passwd. For instance, this area `/data/srv/mariadb` should be owned by account which will run MariaDB, etc.

amaltaro:
Hi Valentin, I think those initialization steps have already been taken care by Todor.

Todor, you were right, after removing the "old" data for previous MariaDB runs, basically deleting this directory:
```
$ ls /data/dockerMount/srv/mariadb/
```

I was able to run MariaDB container and have the service initialized and in run state:
```
(MariaDB-10.6.5) [xxx@vocms0265:data]$ manage version
mariadb-admin  Ver 9.1 Distrib 10.6.5-MariaDB, for debian-linux-gnu on x86_64
Copyright (c) 2000, 2018, Oracle, MariaDB Corporation Ab and others.

Server version		10.6.5-MariaDB-1:10.6.5+maria~focal-log
Protocol version	10
Connection		Localhost via UNIX socket
UNIX socket		/var/run/mysqld/mariadb.sock
Uptime:			1 min 4 sec
```

If anyone wants to give this a try, here is my latest MariaDB tag in CERN registry:
```
Pulling Docker image: REGISTRY/cmsweb/mariadb:10.6.5-alan9
```

Building the CouchDB one now...

todor-ivanov:
Hi @vkuznet All these all steps and configurations, have been done and tested  many many times at least 6 months ago. Lets not go back to MariaDB initialization process. It has been tested and is working just fine.

This:
> If you'll try to install MariaDB on your own node you'll see them, and they can be performed differently. Basically someone will need to run chown on bunch of directories (logs, db, etc) and have proper settings of that account in /etc/passwd. For instance, this area /data/srv/mariadb should be owned by account which will run MariaDB, etc. 

is not going to happen in case you the user trying to run the container and to use a mount point at the host already has write permissions at the parent directory where the mount point would be. And this is the expected usage indeed. The user who is the owner of the database space at the host to be also the user to run the container. Nothing abnormal. I think we are on the right direction.

amaltaro:
I built a fresh new CouchDB image (in CERN registry) and run it with this tag:
xxx@vocms0265:srv $ ./couchdb-docker-run.sh -t 3.2.2-alan9 -p && docker logs -f couchdb

And I can confirm that the permission issue remains there, as previously pointed out in this comment: https://github.com/dmwm/CMSKubernetes/pull/1469#issuecomment-2078020806

From Todor's comment above, should we then change the mount point from:
```
--mount type=bind,source=$HOST_MOUNT_DIR/srv/couchdb/$COUCH_TAG/install/database,target=/data/srv/couchdb/current/install/database \
--mount type=bind,source=$HOST_MOUNT_DIR/srv/couchdb/$COUCH_TAG/logs,target=/data/srv/couchdb/current/logs \
```
to (the whole installation area)
```
--mount type=bind,source=$HOST_MOUNT_DIR/srv/couchdb/$COUCH_TAG/,target=/data/srv/couchdb/current/ \
```
?

vkuznet:
I tried to run latest Alan's container but no luck:
- step 1: download `mariadb-docker-run.sh` from this PR
- step 2: clean-up `data/dockerMount` area
```
node(14:56:06)> rm -rf /data/dockerMount/*
```
- step 3 run the script
```
node(14:56:18)> ./mariadb-docker-run.sh -t 10.6.5-alan9 -p
Creating passwd file
Creating group file
Pulling Docker image: registry.cern.ch/cmsweb/mariadb:10.6.5-alan9
10.6.5-alan9: Pulling from cmsweb/mariadb
Digest: sha256:aad351662fc7c04584f8514e8cfada46268e3887bf5528d3b232f4742c926574
Status: Image is up to date for registry.cern.ch/cmsweb/mariadb:10.6.5-alan9
registry.cern.ch/cmsweb/mariadb:10.6.5-alan9
Starting mariadb:10.6.5-alan9 docker container with user: valya:zh
docker: Error response from daemon: invalid mount config for type "bind": bind source path does not exist: /data/dockerMount/admin/wmagent.
See 'docker run --help'.
```
If you run it, could you please share how you do it.

todor-ivanov:
@vkuznet, Actually, your container is missing to find the `WMAgent.secrets` file. Which is needed for the mariadb initialization, to set up the wmagent user at the database.

vkuznet:
> @vkuznet, Actually, your container is missing to find the `WMAgent.secrets` file which should reside inside `/data/dockerMount/admin/wmagent/`. This is needed for the mariadb initialization, to set up the `wmagent` user at the database.

Todor, can we come up with minimum set of requirements needed to run the image on VM node, e.g.
- `make -p /data/dockerMount/admin/wmagent`
- put `WMAgent.secrets` inside of it, what is minimum context of this file?

todor-ivanov:
Hi @vkuznet please do not try to run these containers in a non WMAgent machine. We will get to that point, but we are not there yet.  This is the regular `WMAgent.secrets` file. In the case of the mariaDB container it needs only three variables from there. Take a look at my reply to Alan in the WMCore Team MM chat, where I have provided the minimal contents of the two secrets files needed for this container to run.

todor-ivanov:
hi @amaltaro 

Here is the root cause of our problems with the couchdb container: 
```
(CouchDB-3.2.2) [cmst1@vocms0260:data]$ manage  init
+ /data/srv/couchdb/3.2.2/manage init
init_couchdb: Adding a new account: cmst1
sed: couldn't open temporary file /data/srv/couchdb/3.2.2/config/sedYiCcGf: Permission denied
(CouchDB-3.2.2) [cmst1@vocms0260:data]$ id 
+ id
uid=31961(cmst1) gid=1399(zh) groups=1399(zh)
(CouchDB-3.2.2) [cmst1@vocms0260:data]$ ll /data/srv/couchdb/3.2.2/config                   
+ ls --color=auto -la --color=auto /data/srv/couchdb/3.2.2/config
total 4
drwxr-xr-x. 1 root root   23 Apr 29 12:31 .
drwxr-xr-x. 1 root root   20 Apr 29 12:31 ..
-rw-r--r--. 1 root root 2197 Apr 29 11:58 local.ini
```

So the `local.ini` file inside the container is with root privileges, and it cannot be opened for editing, so the server cannot be configured  with the password coming from the secrets file.  And from there on all the rest follows up.

docker/pypi/wmagent-couchdb/Dockerfile
```diff
@@ -79,22 +56,14 @@ RUN ln -s ${COUCH_CONFIG_DIR}/local.ini /opt/couchdb/etc/local.d/
...
-alias manage=$COUCH_MANAGE_DIR/manage
-
-# set CouchDB docker specific bash prompt:
-export PS1="(CouchDB-$TAG) [\u@\h:\W]\$ "
+# Set command prompt for root
+RUN <<EOF cat >> /root/.bashrc
+export PS1="(CouchDB-$TAG) [\u@\h:\W]# "
 EOF
 
-RUN chown -R ${USER}:${GROUP} ${COUCH_ROOT_DIR}
```
todor-ivanov, line 95: HI @amaltaro  the removal of this line here is the origin of the problem explained in https://github.com/dmwm/CMSKubernetes/pull/1469#issuecomment-2083139403 

Unfortunately we cannot afford to keep it.  I am thinking about a workaround here.

todor-ivanov: hi @amaltaro 
And I just tested  a workaround and suggested a PR to your branch: 
https://github.com/amaltaro/CMSKubernetes/pull/1 

There is still one error during `couchapp` installation, but now at least the database  user and admin accounts are set correctly:
```
cmst1@vocms0260:wmagent-couchdb $ ./couchdb-docker-run.sh  -t 3.2.2 && docker logs -f couchdb 
Starting couchdb:3.2.2 docker container with user: cmst1:zh
626c2f8650c8cd2cb529abeb2a13cfb4940f4dd9606e388ead9ecefeeecde1b1
Running CouchDB container with user: cmst1 (ID: 31961) and group: zh (ID: 1399)
'/usr/local/local.ini' -> '/data/srv/couchdb/3.2.2/config/local.ini'
init_couchdb: Adding a new account: cmst1
Which couchdb: /opt/couchdb/bin/couchdb
  With configuration directory: /data/srv/couchdb/3.2.2/config
  With logdir: /data/srv/couchdb/3.2.2/logs
  nohup couchdb -couch_ini /data/srv/couchdb/3.2.2/config >> /data/srv/couchdb/3.2.2/logs/couch.log 2>&1 &
Couchapps not found. Installing from latest WMCore tag.
mkdir: cannot create directory '/data/srv/couchdb/3.2.2/install/stagingarea': Permission denied
/data/manage: line 403: pushd: /data/srv/couchdb/3.2.2/install/stagingarea/tmp: No such file or directory
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100  4824  100  4824    0     0  26360      0 --:--:-- --:--:-- --:--:-- 26360

Pulling couchapps version latest from Github...
2.3.2.tar.gz: Permission denied
Cannot write to '2.3.2.tar.gz' (Permission denied).
Error pulling couchapps version latest from Github
start sleeping....zzz
```
I think this one is also solvable in the same manner. I' ll create yet another commit for moving/searching  the  `/data/srv/couchdb/3.2.2/install/` directory to the right place.

todor-ivanov: And I have it all fixed with my PR I'll  give details in the main thread and not in the comment here

todor-ivanov:
Hi @amaltaro,

I have fixed the CouchDB container issue you report in your comment https://github.com/dmwm/CMSKubernetes/pull/1469#issuecomment-2078020806  with my PR here:  https://github.com/amaltaro/CMSKubernetes/pull/1 

So the problems were actually 3 (and two of them originating from the same source, hile the third one I beieve was just a typo):
* We were having `local.ini` (and other things) owned by root inside the container, which was  preventing the user who is currently running the container to actually finilise prioperly the `manage init` command and populate the `[admin]` credentials/section in the file. The fix was to mount  `$COUCH_CONFIG_DIR` from the host and to copy the original `local.ini` file  from the container if needed during `manage init` 
* We were mounting the `/data/dockerMount/srv/couchdb/current/inistall/database` directory instead of the parent   `/data/dockerMount/srv/couchdb/current/inistall/` and due to the nonexistence of the parent inside the container, it was forcing docker engine to create it for us, but since it runs with `root`, inside the container we were  ending up with the following ownership (same for the `satate` btw): 
```
(CouchDB-3.2.2) [cmst1@vocms0260:data]$ ll /data/srv/couchdb/current/        
total 44
drwxr-xr-x. 1 root  root    20 Apr 30 22:00 .
drwxr-xr-x. 1 root  root    19 Apr 30 22:00 ..
drwxr-xr-x. 2 *** **    4096 Apr 30 21:26 config
drwxr-xr-x. 4 root root    4096 Apr 30 22:00 install
drwxr-xr-x. 2 *** **    4096 Apr 30 21:26 logs
-rwxr-xr-x. 1 root  root 28557 Apr 30 21:09 manage
drwxr-xr-x. 2 root root    4096 Apr 30 21:26 state
``` 
while inside the install directory the database  was indeed owned by the current user. This was failing the `update_couchapps` function from the  `manage` script to fail, because it was not able to create any temporary directory inside `install`. So the fix was easy as well - to properly mount both `install` and `state` directories from the host.

* The third error was I believe just a typo. The `$COUCH_PORT` environment variable was removed from the `Dockerfile` so the CouchDB process was there, up and running, but `manage status ` command was not returning anything because  it was having the wrong credentials url built (it requires the port in it) 

Here are the logs from my built:
```
cmst1@vocms0260:wmagent-couchdb $ ./couchdb-docker-run.sh  -t 3.2.2  && docker logs -f couchdb 
Starting couchdb:3.2.2 docker container with user: cmst1:zh
699dbeec5ac9f02e8333f8ca705ffe4244c0ac04fa1dcef770b380a42d0846aa
Running CouchDB container with user: cmst1 (ID: 31961) and group: zh (ID: 1399)
init_couchdb: The local.ini file at the container is already properly configured.
init_couchdb: NOT adding any new admin accounts from the secrets files!
Which couchdb: /opt/couchdb/bin/couchdb
  With configuration directory: /data/srv/couchdb/3.2.2/config
  With logdir: /data/srv/couchdb/3.2.2/logs
  nohup couchdb -couch_ini /data/srv/couchdb/3.2.2/config >> /data/srv/couchdb/3.2.2/logs/couch.log 2>&1 &
Couchapps not found. Installing from latest WMCore tag.
/data/srv/couchdb/3.2.2/install/stagingarea/tmp /data
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100  4824  100  4824    0     0  28376      0 --:--:-- --:--:-- --:--:-- 28376

Pulling couchapps version latest from Github...
2024-04-30 22:00:37 URL:https://codeload.github.com/dmwm/WMCore/tar.gz/refs/tags/2.3.2 [18004734] -> "2.3.2.tar.gz" [1]

Pulling additional reqmon and t0_reqmon dependencies...

Downloading jquery-ui.min.js...
2024-04-30 22:00:38 URL:https://ajax.googleapis.com/ajax/libs/jqueryui/1.8.18/jquery-ui.min.js [201842/201842] -> "jquery-ui.min.js" [1]

Downloading jquery.min.js...
2024-04-30 22:00:38 URL:http://code.jquery.com/jquery-1.7.2.min.js [94840/94840] -> "jquery-1.7.2.min.js" [1]

Downloading Datatables...
2024-04-30 22:00:39 URL:https://datatables.net/releases/DataTables-1.9.1.zip [2415658/2415658] -> "DataTables-1.9.1.zip" [1]

Downloading YUI...
2024-04-30 22:00:40 URL:https://yui.github.io/yui2/archives/yui_2.9.0.zip [14294111/14294111] -> "yui_2.9.0.zip" [1]
/data/srv/couchdb/3.2.2/install/stagingarea/tmp/yui /data/srv/couchdb/3.2.2/install/stagingarea/tmp /data
/data/srv/couchdb/3.2.2/install/stagingarea/tmp /data
Removing old couchapps...
Installing new couchapps...
/data
Cleaning up!
Installing ACDC app into database: http://localhost:5984/acdcserver
Installing GroupUser app into database: http://localhost:5984/acdcserver
Installing ReqMgrAux app into database: http://localhost:5984/reqmgr_auxiliary
Installing ReqMgr app into database: http://localhost:5984/reqmgr_workload_cache
Installing ConfigCache app into database: http://localhost:5984/reqmgr_config_cache
Installing WorkloadSummary app into database: http://localhost:5984/workloadsummary
Installing LogDB app into database: http://localhost:5984/wmstats_logdb
Installing WMStats app into database: http://localhost:5984/wmstats
Installing WMStatsErl app into database: http://localhost:5984/wmstats
Installing WMStatsErl1 app into database: http://localhost:5984/wmstats
Installing WMStatsErl2 app into database: http://localhost:5984/wmstats
Installing WMStatsErl3 app into database: http://localhost:5984/wmstats
Installing WMStatsErl4 app into database: http://localhost:5984/wmstats
Installing WMStatsErl5 app into database: http://localhost:5984/wmstats
Installing WMStatsErl6 app into database: http://localhost:5984/wmstats
Installing WMStatsErl7 app into database: http://localhost:5984/wmstats
Installing T0Request app into database: http://localhost:5984/t0_request
Installing WorkloadSummary app into database: http://localhost:5984/t0_workloadsummary
Installing LogDB app into database: http://localhost:5984/t0_logdb
Installing WMStats app into database: http://localhost:5984/tier0_wmstats
Installing WMStatsErl app into database: http://localhost:5984/tier0_wmstats
Installing WMStatsErl1 app into database: http://localhost:5984/tier0_wmstats
Installing WMStatsErl2 app into database: http://localhost:5984/tier0_wmstats
Installing WMStatsErl3 app into database: http://localhost:5984/tier0_wmstats
Installing WMStatsErl4 app into database: http://localhost:5984/tier0_wmstats
Installing WMStatsErl5 app into database: http://localhost:5984/tier0_wmstats
Installing WMStatsErl6 app into database: http://localhost:5984/tier0_wmstats
Installing WMStatsErl7 app into database: http://localhost:5984/tier0_wmstats
Installing WorkQueue app into database: http://localhost:5984/workqueue
Installing WorkQueue app into database: http://localhost:5984/workqueue_inbox
start sleeping....zzz
```      

And inside the container: 

```
cmst1@vocms0260:wmagent-couchdb $ docker exec -it couchdb bash 
(CouchDB-3.2.2) [cmst1@vocms0260:data]$ manage status 
3.2.2 is RUNNING
No active tasks (e.g. compactions)
```

Please  go ahead and merge my PR into your branch and test.

docker/pypi/wmagent-couchdb/couchdb-docker-run.sh
```diff
@@ -76,7 +76,7 @@ groupEntry=$(getent group $thisGroup)
 [[ -d $HOST_MOUNT_DIR/certs ]] || mkdir -p $HOST_MOUNT_DIR/certs || exit $?
 [[ -d $HOST_MOUNT_DIR/admin/couchdb ]] || mkdir -p $HOST_MOUNT_DIR/admin/couchdb || exit $?
 [[ -d $HOST_MOUNT_DIR/srv/couchdb/$COUCH_TAG/config  ]] || mkdir -p $HOST_MOUNT_DIR/srv/couchdb/$COUCH_TAG/config  || exit $?
-[[ -d $HOST_MOUNT_DIR/srv/couchdb/$COUCH_TAG/install/database ]] || mkdir -p $HOST_MOUNT_DIR/srv/couchdb/$COUCH_TAG/install/database || exit $?
```
todor-ivanov, line 63: Why did you remove this stricture here. The database location is configured quite early, in the Dockerfile at built time: https://github.com/dmwm/CMSKubernetes/blob/f4a43f8a85bb1be3b59b913b59914f5f66014a34/docker/pypi/wmagent-couchdb/Dockerfile#L38 

If you remove this  directory creation here (even though we mount one level up) it would break the container later.

todor-ivanov:
Hi @amaltaro  your suggestion here: https://github.com/dmwm/CMSKubernetes/pull/1469#issuecomment-2082626585

To move from this mount structure:
```
--mount type=bind,source=$HOST_MOUNT_DIR/srv/couchdb/$COUCH_TAG/install/database,target=/data/srv/couchdb/current/install/database \
--mount type=bind,source=$HOST_MOUNT_DIR/srv/couchdb/$COUCH_TAG/logs,target=/data/srv/couchdb/current/logs \
```

to higher in the tree:
```
--mount type=bind,source=$HOST_MOUNT_DIR/srv/couchdb/$COUCH_TAG/,target=/data/srv/couchdb/current/ \
```

Makes a lot of sense. Provided we properly create the whole directory structure at the host as expected from the service later inside the container. As we do here:  https://github.com/dmwm/CMSKubernetes/blob/f4a43f8a85bb1be3b59b913b59914f5f66014a34/docker/pypi/wmagent-couchdb/couchdb-docker-run.sh#L60-L64 

This would have avoided the issues we were experiencing with different ownership at different depth of the mount  points, as I explained with my comment here: https://github.com/dmwm/CMSKubernetes/pull/1469#issuecomment-2087581495

todor-ivanov:
BTW, this  comment: https://github.com/dmwm/CMSKubernetes/pull/1469#issuecomment-2088173245 is valid more or less also for the `mariaDB` and `wmagent` containers as well, although a big portion of the internal structure for the `wmagent` container is or can be created during the init process. We may create a separate issue just for this - to cut the mount paths as high as possible for all containers, so that we do not accumulate many changes in this issue here and we do not deviate from the main purpose of the current one.

amaltaro:
I can confirm that all images now run at CERN with an user provided during runtime (now deployed to vocms0262).
Images have been built and uploaded to CERN registry and this is how I ran each one of them:
```
./mariadb-docker-run.sh -t 10.6.5-alan9 -p && docker logs -f mariadb
./couchdb-docker-run.sh -t 3.2.2-alan9 -p && docker logs -f couchdb
./wmagent-docker-run.sh -t 2.3.3-alan9 -p && docker logs -f wmagent
```

From my side, these is as far as we should go with these changes to provide dynamic user during container runtime.

I appreciate any further feedback you might have on this @todor-ivanov @vkuznet . 

Once we have these changes in, I think we should release a "more" stable release in CERN registry such that others can use that as a baseline for further experiments/validation.

todor-ivanov (approved):
Things are looking good to me @amaltaro I totally agree with you, this is the furthest we should go here. I have tested the containers myslef and they were working as expected.  Please merge it as soon as you consider it ready for production. I'd like to have this in master, so we can rebase the latest changes for T0 on top of it.  Thanks!

PR #1474, opened 2024-05-08
Copy bashrc template instead of moving it, for wmagent

amaltaro:
If we run the `wmagent` container multiple times with the same tag, it fails to create the `.bashrc` file, as can be seen in this log error:
```
mv: cannot stat '/data/srv/wmagent/2.3.3/config/etc/wmagent_bashrc': No such file or directory
./run.sh: line 17: /home/cmst1/.bashrc: No such file or directory
```

so instead of moving it out of the mounted directory, just copy it to ensure future docker run will be able to create it under the home directory.

### DIFF
--- a/docker/pypi/wmagent/run.sh
+++ b/docker/pypi/wmagent/run.sh
@@ -13,7 +13,7 @@ export WMA_USER=$wmaUser
 export USER=$wmaUser
 [[ -d ${HOME} ]] || mkdir -p ${HOME}
 
-mv ${WMA_CONFIG_DIR}/etc/wmagent_bashrc $HOME/.bashrc
+cp -f ${WMA_DEPLOY_DIR}/etc/wmagent_bashrc $HOME/.bashrc
 source $HOME/.bashrc
 
 echo "Start initialization"


todor-ivanov (approved):
looking good. This actually explains what we have observed during out hands on session.

docker/pypi/wmagent/run.sh
```diff
@@ -13,7 +13,7 @@ export WMA_USER=$wmaUser
 export USER=$wmaUser
 [[ -d ${HOME} ]] || mkdir -p ${HOME}
 
-mv ${WMA_CONFIG_DIR}/etc/wmagent_bashrc $HOME/.bashrc
+cp -f ${WMA_CONFIG_DIR}/etc/wmagent_bashrc $HOME/.bashrc
```
vkuznet, line 16: It means that each time run.sh will run it will copy, how about checking first if this file exist? For example:
```
if [ ! -f $HOME/.bashrc ]; then
cp ${WMA_CONFIG_DIR}/etc/wmagent_bashrc $HOME/.bashrc
fi
```

amaltaro: yes, run.sh is supposed to run only when starting up the container, where .bashrc is not yet available. I see no need for this extra check.

vkuznet: and my point is that nothing prevent anyone to login to the pod and run `run.sh` again. If you don't want to check if file exists we open up a door for unexpected behavior that's it.

amaltaro: If someone runs it again, there is no problem at all, the .bashrc file will be again overriden. We can always check things first, but there are cases where it is no needed. Adding more lines of unneeded code only increases maintenance and complexity. Am I missing anything?

vkuznet: I find the entire solution questionable because run.sh has nothing to do with bashrc for the container. In fact, the later should exists without any run.sh script. May be you need to mount your host `${WMA_CONFIG_DIR}/etc/wmagent_bashrc` directly to `$HOME/.bashrc` within any copying. I only pointed out that copy can lead to unexpected behavior which we may not foresee right now.

amaltaro: No, WMA_CONFIG_DIR does not have anything to do with the home directory of the user running the container. In addition, if you followed these developments, we no longer define user accounts at the image level, only at the container during runtime.

So the only possibility would be to mount the user home directory into the container - but for cmst1 account, that would mean mounting the AFS area(!!!). What is implemented is a simple solution to provide a static bashrc for the user running the container.

todor-ivanov: Well..., now..., @amaltaro, since we started talking about `WMA_CONFIG_DIR`, I wonder why would we need to move that file back and forth several times: 
* Copying it once (together with the whole `etc/` parent directory ) to the host during the agent initialization - `init.sh` execution: https://github.com/dmwm/CMSKubernetes/blob/e2d764a0a4c970b3ceff58f913e9bdee10328b25/docker/pypi/wmagent/init.sh#L128 
* Subsequently copying it again from the host area back to the container area during container startup. And risking to lose it meanwhile.

We are not making any tweaks to this file worth saving between container restarts. The only thing that may change at runtime is the destination the file needs to go, namely the `$HOME` variable. I'd suggest we simply switch to coping this file directly from the container's deployment area like this:

```
cp -f ${WMA_DEPLOY_DIR}/etc/wmagent_bashrc $HOME/.bashrc
```
BTW if we need to be strict and run by the book, we should do the same with the `$HOME/.bash_profile`

@vkuznet this file we create in advance: https://github.com/dmwm/CMSKubernetes/blob/master/docker/pypi/wmagent/etc/wmagent_bashrc and we have it inside the container already. But it is not at the proper place, it is just at the generic Deployment area of the container: `$WMA_DEPLOY_DIR/etc/`, and the user may change at runtime, so we need to position it correctly. Having it copied to the host is just a side effect, of the need to copy the job runtime environment, which we had to add recently to those containers. Said that - I can say mounting the file alone would be more difficult, 
* first -  because the file starts to exists at the host as part of the agent initialization process, so it is missing at container startup, when it starts for the first time,  
* second - because it would cause nested mounts from the host mount area (i.e. `/data/dockerMount/srv/wmagent/current/config/etc/wmagent_bashrc`, which is already mounted few levels higher from `/data/dockerMount/srv/wmagent/current/config`). This is would possibly break the original mount or vice versa (the `$HOME/.bashrc` mount). I've experienced similar problems in the past.

amaltaro: It is not that we copy it back and forth, but we were moving this file to the expected destination. Upon same tag container restart, we mount the `config` dir:
https://github.com/dmwm/CMSKubernetes/blob/master/docker/pypi/wmagent/wmagent-docker-run.sh#L97C1-L98C1
, which is already populated - and without the actual `wmagent_bashrc`, which fails.

Yes, in this case copying it from `WMA_DEPLOY_DIR` look a better option. Let me update this PR.

todor-ivanov (approved):
looks good

amaltaro:
It looks like we converged on this issue. @vkuznet if you still have any concerns with this, please let me know. Otherwise I will get this merged after our meeting today.

PR #1477, opened 2024-05-16
Update values-test2.yaml

belforte:
bump crabserver version in cmsweb-test2

### DIFF
--- a/helm/crabserver/values-test2.yaml
+++ b/helm/crabserver/values-test2.yaml
@@ -1,4 +1,4 @@
 environment: "test"
 
 image:
-  tag: "v3.240325"
+  tag: "v3.240515"


arooshap (changes requested):
Hi @belforte,

Is it possible for you to also increment the version similar to https://github.com/dmwm/CMSKubernetes/commit/c2b643cb95e502c7773e16ab6953c9b920476e60? It will make sure that the build for the helm chart does not fail. Thanks! :)

arooshap (approved):
I see what you mean. I had a discussion with Dario and in this case, it should be fine.

mapellidario:
So, I had a chat with Aroosha and explained our reasoning. 

She wants us to upgrade the `.Chart.version` every time that we increase the crabserver image tag because that triggers a github action that deploys the new helm chart to the registry [1]. This makes sure that the cmsweb operator scripts always get the latest helm chart with the latest image tags.

Increasing the `.Chart.version`/`.Chart.appVersion` every time there is an image tag change is not enforced by helm and is an open topic and there is discussion going on whether helm should support automatic `.Chart.appVersion` increase at every helm upgrade.

With Aroosha we agreed on a middle ground:

- crab team increases only the image tag in the `values-*.yaml` file every time it needs to deploy a new image
- when the cmsweb team needs to make a new cluster and need to deploy crabserver from scratch, they bump the `.Chart.appVersion` in order to please their scripts.

[1] https://registry.cern.ch/harbor/projects/1771/helm-charts/crabserver/versions
[2] https://github.com/helm/helm/issues/8194#issuecomment-658715462

belforte:
I am glad to see that I am not the only one confused by chart.version vs. app.version. Given that [2] above is a few years old and marked as locked/resolved... I will not hold my breath waiting for a breakthrough.

PR #1478, opened 2024-05-16
bump crabserver version in values-testbed.yaml

belforte:


belforte:
already deployed. But it is good to keep consistency with GH

PR #1480, opened 2024-05-17
WMAgent: change certs mount point

amaltaro:
Fixes https://github.com/dmwm/WMCore/issues/11987

Instead of mounting the certificates from `HOST_MOUNT_DIR` environment variable (set to /data/dockerMount), use the upper directory, such that it reflects the same proxy path between the host and the container.

### DIFF
--- a/docker/pypi/wmagent/wmagent-docker-run.sh
+++ b/docker/pypi/wmagent/wmagent-docker-run.sh
@@ -92,7 +92,7 @@ dockerOpts=" \
 $tnsMount \
 --mount type=bind,source=/etc/condor,target=/etc/condor,readonly \
 --mount type=bind,source=/tmp,target=/tmp \
---mount type=bind,source=$HOST_MOUNT_DIR/certs,target=/data/certs \
+--mount type=bind,source=/data/certs,target=/data/certs \
 --mount type=bind,source=$HOST_MOUNT_DIR/srv/wmagent/$WMA_TAG/install,target=/data/srv/wmagent/current/install \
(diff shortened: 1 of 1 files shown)


docker/pypi/wmagent/wmagent-docker-run.sh
```diff
@@ -92,7 +92,7 @@ dockerOpts=" \
 $tnsMount \
 --mount type=bind,source=/etc/condor,target=/etc/condor,readonly \
 --mount type=bind,source=/tmp,target=/tmp \
---mount type=bind,source=$HOST_MOUNT_DIR/certs,target=/data/certs \
+--mount type=bind,source=$HOST_MOUNT_DIR/../certs,target=/data/certs \
```
todor-ivanov, line 95: hi @amaltaro 
Provided we now know there is requirement for absolute match of the mount point and host path, I am more in favor of: 
```
--mount type=bind,source=/data/certs,target=/data/certs \
```

todor-ivanov (changes requested):
Alan, I left one comment inline. Please take a look

amaltaro:
I totally agree with that! But whenever I do so, people complain that I am hard-coding things :P
@todor-ivanov code has been updated.

todor-ivanov:
thanks @amaltaro 
> I totally agree with that! But whenever I do so, people complain that I am hard-coding things :P

I  am usually the one to complain the most about this, but this time I think we have a strong case.

amaltaro:
@todor-ivanov I think you forgot to approve this, but please let me know if there was anything else missing.
I am merging it such that we can release a new version with all the fixes that went in in the last 48h or so. Thanks

PR #1481, opened 2024-05-17
crabserver helm - deploy script that checks cluster name.

mapellidario:
As asked at [1] and in private discussions, we would like to avoid deploying crabserver to the wrong cluster when using `helm template`.

I added a script that matches the current cluster name in the kubectl config with the desired cluster provided to the script. This script is compatible with both cmsweb user configs (officially supported way of accessing the k8s clusters) and with local configurations managed by kubectx (as done by dario and wa). I upgraded instructions at [2]

Thanks @novicecpp for the skeleton of the script :)

[1] https://github.com/dmwm/CMSKubernetes/pull/1477#issuecomment-2115427365
[2] https://cmscrab.docs.cern.ch/technical/crab-rest/deploy.html#deploy-on-kubernetes-use-helm-template-to-generate-manifest-preferred


mapellidario:
fyi: @belforte

belforte:
thank you Dario

arooshap:
Looks awesome 🤩 👏

PR #1482, opened 2024-05-17
Export USER and WMA_USER variables from $WMA_ENV_FILE

todor-ivanov:
Fixes: https://github.com/dmwm/WMCore/issues/11989 

With the current PR we export  the USER and WMA_USER variables from the $WMA_ENV_FILE in parallel to `run.sh` in order to make no difference in the environment between
* Docker entry point - `run.sh` 
* bash login session

### DIFF
--- a/docker/pypi/wmagent/install.sh
+++ b/docker/pypi/wmagent/install.sh
@@ -144,6 +144,8 @@ export CRYPTOGRAPHY_ALLOW_OPENSSL_102=true
 export YUI_ROOT=$WMA_DEPLOY_DIR/yui/
 export PATH=\$WMA_INSTALL_DIR/bin\${PATH:+:\$PATH}
 export PATH=\$WMA_DEPLOY_DIR/bin\${PATH:+:\$PATH}
+export USER=\$(id -un)
+export WMA_USER=\$(id -un)
 EOF
 }
 


docker/pypi/wmagent/install.sh
```diff
@@ -144,6 +144,8 @@ export CRYPTOGRAPHY_ALLOW_OPENSSL_102=true
 export YUI_ROOT=$WMA_DEPLOY_DIR/yui/
 export PATH=\$WMA_INSTALL_DIR/bin\${PATH:+:\$PATH}
 export PATH=\$WMA_DEPLOY_DIR/bin\${PATH:+:\$PATH}
+export USER=\$(id -un)
+export WMA_USER=\$(id -un)
```
amaltaro, line 148: I think all these imports are incorrectly escaping the $ sign. I will comment further in the final review comment.

amaltaro (changes requested):
Todor, as mentioned in the code, unless I am missing something, I think we are incorrectly escaping the $ sign when assigning/exporting those variables.

Here is a quick test I did:
```
$ cat install.sh 
#!/bin/bash
export USER=\$(id -un)
export WMA_USER=\$(id -un)
```

then executing it:
```
$ ./install.sh 
./install.sh: line 2: syntax error near unexpected token `('
./install.sh: line 2: `export USER=\$(id -un)'
```

Similar problem gets reported if we source it.

docker/pypi/wmagent/install.sh
todor-ivanov: well the reallity is is that we should also escape the \( and \), because hat we need is to echo tis command as it is in the WMA_ENV_FILE, so that we delay the execution to be at the time the $WMA_ENV_FILE is actually sourced. We have no use of these if we  execute them t build time inside the `install.sh` script. At this stage noone knows  what the user to run the container would be.
I'll make the fix in a while

amaltaro: Oh, now it makes sense. You are basically defining those environment variables with other environment variables that might not yet be defined.

anpicci:
I confirm it works on FNAL:
```
[cmsdataops@cmssrv810 wmagent]$ docker exec -it wmagent bash
(WMAgent-2.3.3) [cmsdataops@cmssrv810:current]$ python
Python 3.8.16 (default, May 23 2023, 14:26:40) 
[GCC 10.2.1 20210110] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import os
>>> os.environ["USER"]
'cmsdataops'
>>> os.environ["WMA_USER"]
'cmsdataops'
```

I'm gonna injecting a workflow again for further testing it. In the meantime, I approve the PR

anpicci (approved):

docker/pypi/wmagent/install.sh
anpicci: @amaltaro I think we can consider this conversation resolved, and move forward

amaltaro: We cannot, this code needs to be fixed.

amaltaro:
@anpicci how did you test this? Did you build a new image and run it?
From the tests I've done above - and confirmed by Todor - this code needs to be fixed.

todor-ivanov:
@amaltaro What you do in your test, is not the same what is happening in the `install.sh` script:

* This
```
$ cat install.sh
#!/bin/bash
export USER=\$(id -un)
export WMA_USER=\$(id -un)

$ ./install.sh
```

* Is not the same as  this:
```
$ cat >> some_file.sh <<EOF
#!/bin/bash
export USER=\$(id -un)
export WMA_USER=\$(id -un)
EOF

$ cat some_file.sh 
#!/bin/bash
export USER=$(id -un)
export WMA_USER=$(id -un)
```

What happens in our `install.sh` script is the later. Take a look at the whole block (not only at the so added  two lines): https://github.com/dmwm/CMSKubernetes/blob/aaeb3b53d0a0b8b4f83740741ee75e6c1d477e44/docker/pypi/wmagent/install.sh#L136-L147 So we are creating the $WMA_ENV_FILE in the way how it would be executed  at the host at tuntime, but we are not executing it during the execution of `install.sh`. Initially I got confused of your test last night midnight and I did not notice the difference. So my comment about the extra escapes is valid only if you want to make your code work (if even a true comment), but is not applicable to the actual `install.sh` file. So the fix as I provided to this PR is actually the correct one. I can confirm this. See my console output bellow: 

```
cmst1@vocms0260:wmagent $ ./wmagent-docker-build.sh -t 2.3.4rc3
#0 building with "default" instance using docker driver

#1 [internal] load build definition from Dockerfile
#1 transferring dockerfile: 1.84kB done
#1 DONE 0.0s
...


cmst1@vocms0260:wmagent $ ./wmagent-docker-run.sh -t 2.3.4rc3 && docker logs -f wmagent
Checking if there is no other wmagent container running and creating a link to the 2.3.4rc3 in the host mount area.
Starting wmagent:2.3.4rc3 docker container with user: cmst1:zh
4d6c219c4b4748679a9a92ff0831b8f8999fba6225b918ee6cb8617d44a0d48c
Running WMAgent container with user: cmst1 (ID: 31961) and group: zh (ID: 1399)
Setting up bashrc for user: cmst1 under home directory: /home/cmst1
Start initialization

=======================================================
Starting WMAgent with the following initialisation data:
-------------------------------------------------------
 - WMAgent Version            : 2.3.4rc3
 - WMAgent User               : cmst1
 - WMAgent Root path          : /data
 - WMAgent Host               : vocms0260.cern.ch
 - WMAgent TeamName           : testbed-vocms0260
 - WMAgent Number             : 0
 - WMAgent Relational DB type : mysql
 - Python  Version            : Python 3.8.16
 - Python  Module path        : /usr/local/lib/python3.8/site-packages
=======================================================

....
Start sleeping now ...zzz...


cmst1@vocms0260:wmagent $ docker exec -it wmagent bash 

(WMAgent-2.3.4rc3) [cmst1@vocms0260:current]$ 
(WMAgent-2.3.4rc3) [cmst1@vocms0260:current]$ env |grep -i user
X509_USER_CERT=/data/certs/servicecert.pem
USER=cmst1
WMA_USER=cmst1
X509_USER_KEY=/data/certs/servicekey.pem
X509_USER_PROXY=/data/certs/myproxy.pem
```

anpicci:
@amaltaro I modified the `install.sh` as proposed in this PR, then build a new local Docker image with
`./wmagent-docker-build.sh -t 2.3.3 && docker logs -f wmagent`, which gave
```
-----------------------------------------------------------------------
-----------------------------------------------------------------------
Start Generating and preserving current build id
WMA_BUILD_ID:a03331a18dc4c5c9b1db9d07e762dd3c7b4062cd861341cbe0d3d5e4e6be945a
WMA_BUILD_ID preserved at: /data/.dockerBuildId 
Done Generating and preserving current build id!

-----------------------------------------------------------------------
-----------------------------------------------------------------------
Start Replace the current /data/manage script coming from 'dmwm-base' image with a symlink link
Done Replace the current /data/manage script coming from 'dmwm-base' image with a symlink link!

-----------------------------------------------------------------------
-----------------------------------------------------------------------
Start Tweaking runtime environment for user: 
Edit $WMA_ENV_FILE script to point to $WMA_ROOT_DIR
Edit $WMA_ENV_FILE script to point to the correct install, config and manage
Edit /usr/local/deploy/renew_proxy.sh script to point to $WMA_ROOT_DIR
Done Tweaking runtime environment for user: !

-----------------------------------------------------------------------
-----------------------------------------------------------------------
WMAgent image build finished!!

Have a nice day!

=======================================================================
```
As next step:
```
./wmagent-docker-run.sh -t 2.3.3 && docker logs -f wmagent

-------------------------------------------------------
-------------------------------------------------------
Start: Performing checks for successful Docker initialisation steps...
WMA_BUILD_ID: a03331a18dc4c5c9b1db9d07e762dd3c7b4062cd861341cbe0d3d5e4e6be945a
dockerInitId: a03331a18dc4c5c9b1db9d07e762dd3c7b4062cd861341cbe0d3d5e4e6be945a
OK

Docker container has been initialised! However you still need to:
  1) Double check agent configuration: less /data/[dockerMount]/srv/wmagent/current/config/config.py
  2) Start the agent by either of the methods bellow:
     a) From inside the already running container
          * Access the running WMAgent container:
            docker exec -it wmagent bash
          * Use the regular manage script inside the container:
            manage start-agent

     b) From the host - by restarting the whole container
          * Kill the currently running container:
            docker kill wmagent
          * Start a fresh instance of wmagent:
            ./wmagent-docker-run.sh -t <WMA_TAG> && docker logs -f wmagent
Have a nice day!

Start sleeping now ...zzz...
```
and finally activated the agent inside the container.

amaltaro (approved):

amaltaro:
@todor-ivanov @anpicci you are both correct! I didn't see it was enclosed in a "here document", sorry about the noise!
Given the tests and follow up, I am merging this such that we have a cleaner/better image on Monday.

PR #1485, opened 2024-05-20
Parsing the WMA_TAG and using only the release part for creating the WMA_BUILD_ID

todor-ivanov:
Fixes: https://github.com/dmwm/WMCore/issues/11990

With the current PR we create the WMA_BUILD_ID based only on the release part of the WMA_TAG of the currently built  agent. This way we should avoid restarting initialization process (hence agent data and databse deletion) on patch releases or release candidates.

### DIFF
--- a/docker/pypi/wmagent/install.sh
+++ b/docker/pypi/wmagent/install.sh
@@ -48,11 +48,19 @@ done
 WMA_TAG_REG="^[0-9]+\.[0-9]+\.[0-9]{1,2}((\.|rc)[0-9]{1,2})?$"
 [[ $WMA_TAG =~ $WMA_TAG_REG ]] || { echo "WMA_TAG: $WMA_TAG does not match requered expression: $WMA_TAG_REG"; echo "EXIT with Error 1"  ; exit 1 ;}
 
+# Parsing the WMA_TAG
+declare -A WMA_VER
+WMA_VER[release]=$(echo $WMA_TAG| sed -nr 's/.*(^[0-9]+\.[0-9]+\.[0-9]+).*$/\1/p')
+WMA_VER[patch]=$(echo $WMA_TAG| sed -nr 's/.*(^[0-9]+\.[0-9]+\.[0-9]+)//p')
+WMA_VER[major]=$(echo $WMA_TAG| sed -nr 's/.*(^[0-9]+\.[0-9]+).*/\1/p')
+WMA_VER[minor]=$(echo $WMA_TAG| sed -nr 's/^[0-9]+\.[0-9]+\.([0-9]+).*$/\1/p')
+
 echo
 echo "======================================================================="
 echo "Starting new WMAgent deployment with the following initialisation data:"
 echo "-----------------------------------------------------------------------"
 echo " - WMAgent Version            : $WMA_TAG"
+echo " - WMAgent Release Cycle      : ${WMA_VER[release]}"
 echo " - WMAgent User               : $WMA_USER"
 echo " - WMAgent Root path          : $WMA_ROOT_DIR"
 echo " - Python  Version            : $(python --version)"
@@ -105,7 +113,8 @@ echo "-----------------------------------------------------------------------"
 stepMsg="Generating and preserving current build id"
 echo "-----------------------------------------------------------------------"
 echo "Start $stepMsg"
-echo $RANDOM |sha256sum |awk '{print $1}' > $WMA_ROOT_DIR/.dockerBuildId
+
+echo ${WMA_VER[release]}| sha256sum |awk '{print $1}' > $WMA_ROOT_DIR/.dockerBuildId
 echo "WMA_BUILD_ID:`cat $WMA_ROOT_DIR/.dockerBuildId`"
 echo "WMA_BUILD_ID preserved at: $WMA_ROOT_DIR/.dockerBuildId "
 echo "Done $stepMsg!" && echo


todor-ivanov:
Not to copy paste the test results in multiple places, I just linking them in the current comment here from the WMcore issue: https://github.com/dmwm/WMCore/issues/11990#issuecomment-2121391425

docker/pypi/wmagent/install.sh
```diff
@@ -48,11 +48,22 @@ done
...
+WMA_VER[patch]=$(echo $WMA_TAG|sed -nr 's/.*(^[0-9]+\.[0-9]+\.[0-9]+)//p')
+WMA_VER[major]=$(echo $WMA_TAG|sed -nr 's/.*(^[0-9]+\.[0-9]+).*/\1/p')
+WMA_VER[minor]=$(echo $WMA_TAG|sed -nr 's/^[0-9]+\.[0-9]+\.([0-9]+).*$/\1/p')
+
 echo
 echo "======================================================================="
 echo "Starting new WMAgent deployment with the following initialisation data:"
 echo "-----------------------------------------------------------------------"
-echo " - WMAgent Version            : $WMA_TAG"
+echo " - WMAgent Tag                : $WMA_TAG"
```
amaltaro, line 62: I would suggest to keep it saying "WMAgent Version".
Then rephrase "WMAgent Release" to "WMAgent Release Cycle".
Lastly, remove the Major, Minor and Patch, and those might be confusing and maybe even incorrect (e.g., for release 2.3.3).

docker/pypi/wmagent/install.sh
```diff
@@ -105,7 +116,8 @@ echo "-----------------------------------------------------------------------"
 stepMsg="Generating and preserving current build id"
 echo "-----------------------------------------------------------------------"
 echo "Start $stepMsg"
-echo $RANDOM |sha256sum |awk '{print $1}' > $WMA_ROOT_DIR/.dockerBuildId
+
+echo ${WMA_VER[release]}|sha256sum |awk '{print $1}' > $WMA_ROOT_DIR/.dockerBuildId
```
amaltaro, line 120: To make it more readable, I would suggest whitespaces around the pipe operation.

docker/pypi/wmagent/install.sh
```diff
@@ -48,11 +48,22 @@ done
 WMA_TAG_REG="^[0-9]+\.[0-9]+\.[0-9]{1,2}((\.|rc)[0-9]{1,2})?$"
 [[ $WMA_TAG =~ $WMA_TAG_REG ]] || { echo "WMA_TAG: $WMA_TAG does not match requered expression: $WMA_TAG_REG"; echo "EXIT with Error 1"  ; exit 1 ;}
 
+# Parsing the WMA_TAG
+declare -A WMA_VER
+WMA_VER[release]=$(echo $WMA_TAG|sed -nr 's/.*(^[0-9]+\.[0-9]+\.[0-9]+).*$/\1/p')
+WMA_VER[patch]=$(echo $WMA_TAG|sed -nr 's/.*(^[0-9]+\.[0-9]+\.[0-9]+)//p')
```
amaltaro, line 54: To make it more readable, I would suggest whitespaces around the pipe operation. Same for the line above.

amaltaro (commented):
@todor-ivanov changes are looking good to me. However, I have made suggestion for minor improvements.

docker/pypi/wmagent/install.sh
todor-ivanov: ok, 

... the way I dd it was to be extra explicit in the meaning and the matching between the print outs and the variables themselves, but that's minor. What I cannot understand is why would the `major/minor` and `patch`  part of the tag be incorrect for a tag like `WMA_TAG=2.3.3` ? I have tested it many times  and in the case of a tag like this what those pirntouts would say would be:
```
#11 0.115 =======================================================================
#11 0.115 Starting new WMAgent deployment with the following initialisation data:
#11 0.115 -----------------------------------------------------------------------
#11 0.116  - WMAgent Tag                : 2.3.3
#11 0.116  - WMAgent Release            : 2.3.3
#11 0.116  - WMAgent Major version      : 2.3
#11 0.116  - WMAgent Minor version      : 3
#11 0.116  - WMAgent Patch/Release cand : 
#11 0.116  - WMAgent User               : 
#11 0.116  - WMAgent Root path          : /data
#11 0.118  - Python  Version            : Python 3.8.16
#11 0.118  - Python  Module path        : /usr/local/lib/python3.8/site-packages
#11 0.118 =======================================================================
``` 

Which is exactly what is expected I believe.

todor-ivanov:
hi @amaltaro addressed  your comments. could you please  take another look?

docker/pypi/wmagent/install.sh
amaltaro: I think the concept of release type for our tagging convention is not straight forward. For instance, for 2.3.4, I would say:
* major release: 2
* middle release: 3
* minor release: 4

but for 2.3.4.5, I see two ways to read it. Either as:
* major release: 2
* middle release: 3
* minor release: 4
* patch release: 5
OR
* major release: 2
* middle release: 3.4
* minor release: 5
* patch release: 

Due to the potential confusion on this interpretation, I feel like it is better not to give such information.

amaltaro (approved):
Thanks Todor. You missed the whitespaces around `|` operator, but we can move forward as is.

PR #1486, opened 2024-05-21
Fix certs directory in WMAgent CouchDB image as well

amaltaro:
Fixes https://github.com/dmwm/WMCore/issues/11987
Complement to https://github.com/dmwm/CMSKubernetes/pull/1480

Summary of changes is:
* use the same /data/certs directory for WMAgent CouchDB image
* remove certificates from the WMAgent MariaDB image (as those are not needed)
* remove check for old certs directory in the WMAgent run


docker/pypi/wmagent-couchdb/couchdb-docker-run.sh
```diff
@@ -89,7 +88,7 @@ dockerOpts="
 --user $(id -u):$(id -g) \
 --name=couchdb \
 --mount type=bind,source=/tmp,target=/tmp \
---mount type=bind,source=$HOST_MOUNT_DIR/certs,target=/data/certs \
+--mount type=bind,source=/data/certs,target=/data/certs \
```
vkuznet, line 91: I'm not sure why you decided to drop `$HOST_MOUNT_DIR` in favor of `/data/certs`. The former can be located anywhere in a host node, while later is more restrict and will require:
- presence of `/data` on a host node which may or may not be the case
- since it is `/data` under the root `/` you may require sudo access on host node unless it is preconfigured for your by sysadmin and you may not have ability to write to this area to store your original certificates. Look at a line below which allows to keep `$HOST_MOUNT_DIR/src` anywhere in a host node without requiring access to root partition or sysadmin rights.

todor-ivanov (approved):
looks good to me

PR #1490, opened 2024-05-24
WМАgent switch deployment model copy runtime fix 11990

todor-ivanov:
Fixes: https://github.com/dmwm/WMCore/issues/11990

With the current PR we are changing few more things in addition to change of behavior for triggering the agent initialization process only on release change, but rather on any change in the patch version or release candidate.
The additional changes are as follows:

* We shift and now we base the host mount point on the release version, rather than on the full `WMA_TAG` as up to now
* We Enforce copying the Runtime code in case the following two conditions are fulfilled:
    * The agent has been previously initialized
    * There is a change between the full WMA_TAG of the currently starting container and the one previously initialized at the host in the mount area pointed by the current release.

### DIFF
--- a/docker/pypi/wmagent/init.sh
+++ b/docker/pypi/wmagent/init.sh
@@ -50,6 +50,7 @@ echo "======================================================="
 echo "Starting WMAgent with the following initialisation data:"
 echo "-------------------------------------------------------"
 echo " - WMAgent Version            : $WMA_TAG"
+echo " - WMAgent Release Cycle      : $WMA_VER_RELEASE"
 echo " - WMAgent User               : $WMA_USER"
 echo " - WMAgent Root path          : $WMA_ROOT_DIR"
 echo " - WMAgent Host               : $HOSTNAME"
@@ -118,6 +119,15 @@ _copy_runtime() {
         echo "$FUNCNAME: Copying content from: $WMA_DEPLOY_DIR/etc/ to: $WMA_CONFIG_DIR/"
         cp -ra $WMA_DEPLOY_DIR/etc/ $WMA_CONFIG_DIR/
         echo $WMA_BUILD_ID > $wmaInitRuntime
+
+        # NOTE: If the sql database have been previously initialized, then the
+        #       current call of _copy_runtime has been due to a change of the
+        #       $WMA_TAG - either a patch version or a release candidate change
+        #       and we need to update the record at the database as well
+        _init_valid $wmaInitSqlDB && {
+            echo "$FUNCNAME: Preserving the current WMA_TAG at the wma_init database"
+            _exec_sql "update wma_init set init_value='$WMA_TAG' where init_param='wma_tag';"
+        }
     else
         echo "$FUNCNAME: ERROR: \$WMA_DEPLOY_DIR: $WMA_DEPLOY_DIR is not a root path for \$pythonLib: $pythonLib"
         echo "$FUNCNAME: ERROR: We cannot find the correct WMCore/WMRuntime source to copy at the current host!"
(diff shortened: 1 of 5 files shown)


todor-ivanov:
And here are the test results: https://github.com/dmwm/WMCore/issues/11990#issuecomment-2128276723
The Issue description has been updated accordingly as well.

docker/pypi/wmagent/init.sh
```diff
@@ -118,6 +119,15 @@ _copy_runtime() {
         echo "$FUNCNAME: Copying content from: $WMA_DEPLOY_DIR/etc/ to: $WMA_CONFIG_DIR/"
         cp -ra $WMA_DEPLOY_DIR/etc/ $WMA_CONFIG_DIR/
         echo $WMA_BUILD_ID > $wmaInitRuntime
+
+        # NOTE: If the sql database have been previously initialized, then the
+        #       current call of _copy_runtime has been due to a change of the
+        #       $WMA_TAG - either a patch version or a release candidate change
+        #       and we need to update the record at the database as well
+        _init_valid $wmaInitSqlDB && {
```
amaltaro, line 127: I would suggest to make a new function and separate this action from the runtime action.

amaltaro:
Even though the WMCore CD pipeline succeeded building the `wmagent` image, it does not work in my Alma9 VM, as it fails with:
```
Step 6/38 : ENV WMA_ROOT_DIR=/data
 ---> Using cache
 ---> 606bd930db6e
Step 7/38 : ENV WMA_VER_MINOR=${WMA_TAG#*.*.}
failed to process "${WMA_TAG#*.*.}": unsupported modifier (#) in substitution
```

This can be resolved if we install the `buildx` client with:
```
sudo dnf install docker-buildx-plugin.x86_64
```

and enable it to be used with: `export DOCKER_BUILDKIT=1`.

Just sharing it in case others face the same issue.

amaltaro:
Another issue, it fails to pull the image from the registry, as now it's misusing/mislabeling the images:
```
cmst1@vocms0258:pypi $ ./wmagent/wmagent-docker-run.sh -t 2.3.4rc5 -p && docker logs -f wmagent
ln: failed to create symbolic link '/data/srv/wmagent/wmagent': File exists
Pulling Docker image: registry.cern.ch/cmsweb/wmagent:2.3.4
...
Error response from daemon: unknown: artifact cmsweb/wmagent:2.3.4 not found
Error response from daemon: No such image: registry.cern.ch/cmsweb/wmagent:2.3.4
Error response from daemon: No such image: registry.cern.ch/cmsweb/wmagent:2.3.4
Checking if there is no other wmagent container running and creating a link to the 2.3.4 in the host mount area.
Starting wmagent:2.3.4rc5 docker container with user: cmst1:zh
Unable to find image 'local/wmagent:2.3.4rc5' locally
docker: Error response from daemon: pull access denied for local/wmagent, repository does not exist or may require 'docker login': denied: requested access to the resource is denied.
```

I am working on this.

docker/pypi/wmagent/init.sh
```diff
@@ -331,6 +341,29 @@ EOF
...
+    # NOTE: Before we even start checking all the rest of the initialization flags, we should
+    #       check if this agent has ever been used, and if there is any patch version
+    #       or release candidate change between the current one and the one previously
+    #       initialized at the host. If such discrepancy is confirmed, we should enforce Runtime
+    #       code copy by deleting only the $wmaInitSqlDB and let the rest of the init process take over.
+    _init_valid $wmaInitSqlDB && {
+        echo "$FUNCNAME: This agent has been previously initialized. Checking for WMAgent version change since last run."
+        local wmaTagCurr=$WMA_TAG
+        local wmaTagSqlDB=$(_exec_sql "select init_value from wma_init where init_param = 'wma_tag';")
+        [[ $wmaTagCurr == $wmaTagSqlDB ]] || {
```
amaltaro, line 359: Shouldn't it be `!=`?

docker/pypi/wmagent/init.sh
todor-ivanov: Actually it must be inside the  `_copy_runtime` action, because we have to make  sure that once the runtime files have been copied we must update the database reflecting the actual flag. What I mean those two steps are sequential and dependent.

docker/pypi/wmagent/init.sh
todor-ivanov: No, It must be `==` and we should act on broken condition i.e. with `||` logical operator. Or... if we change it 
to `!=`  to include the negative inside the condition itself,  then we should act on satisfied (negative) condition with `&&` ... And you know my opinion about double negation in the logic.....

amaltaro: Ahh, I am always getting confused with those. It looks like my brain has a hard time to read this without an explicit if :) Thanks Todor.

PR #1484, opened 2024-05-20
Find and gracefully stop WMAgent CouchDB processes

amaltaro:
Fixes https://github.com/dmwm/WMCore/issues/11991

WMAgent CouchDB does not run under the _couchdb account, so I refactored the `ps aux` command and I also decided to change the SIGKILL by a more soft SIGTERM signal (even though [this post](https://github.com/apache/couchdb/issues/873) suggests a sigkill not to be a problem).

Just tested it and all of the couchdb/couchjs processes are terminated; I also tested restarting the CouchDB process and status and everything looks fine.

PR #1496, opened 2024-05-28
Fix CouchDB manage restart option

amaltaro:
Fixes https://github.com/dmwm/WMCore/issues/11945

* 1st commit resolves a regression injected with https://github.com/dmwm/CMSKubernetes/pull/1484, which makes the manage script terminates itself when executing the `stop` function. With this fix, when listing couchdb processes with grep, we filter out `grep` and `/manage` now.
* 2nd commit adds some useful aliases to the bashrc template copied to the container. Note that the condor aliases could be deleted, given that we don't have any of the condor binaries in the path.

### DIFF
--- a/docker/pypi/wmagent-couchdb/manage
+++ b/docker/pypi/wmagent-couchdb/manage
@@ -266,12 +266,12 @@ sysboot()
 # Start the service.
 start()
 {
+  echo "Starting CouchDB service..."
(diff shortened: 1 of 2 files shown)


anpicci:
Hi @amaltaro , should I check this once the test workflows are done?

amaltaro:
> Hi @amaltaro , should I check this once the test workflows are done?

No need for the moment. It's known to be broken and I'd repeat this test once we merge this fix in and have a new image to verify this functionality.

anpicci (approved):
Thank you @amaltaro

PR #1497, opened 2024-05-29
WMAgent: Install voms-clients-java (same package used in FNAL/CERN hosts)

khurtado:
Fixes [Fixes #11999](https://github.com/dmwm/WMCore/issues/11999)
With a newer version (java based) of voms-proxy-init. This is the same voms clients used by FNAL hosts (3.x)

CPP version:
$ voms-proxy-init2 --version
voms-proxy-init
Version: 2.1.0
Compiled: Jul 24 2020 08:53:00

Java version:
$ voms-proxy-init --version
voms-proxy-init v. 3.3.2 (voms-api-java/3.3.2 canl/2.6.0 bcprov/1.68.0 bcpkix/1.68.00.0)

### DIFF
--- a/docker/pypi/wmagent-base/Dockerfile
+++ b/docker/pypi/wmagent-base/Dockerfile
@@ -3,7 +3,7 @@ FROM registry.cern.ch/cmsweb/dmwm-base:pypi-20230525
 
 # Install basic OS package dependencies
 RUN apt-get update
-RUN apt-get install -y libmariadb-dev-compat libmariadb-dev apache2-utils cron mariadb-server myproxy voms-clients rlwrap libaio1 && apt-get clean
+RUN apt-get install -y libmariadb-dev-compat libmariadb-dev apache2-utils cron mariadb-server myproxy voms-clients voms-clients-java rlwrap libaio1 && apt-get clean
 
 # Install some debugging tools
 RUN apt-get install -y hostname net-tools iputils-ping procps && apt-get clean


khurtado:
@amaltaro Note we may not need `voms-clients` anymore, but I don't know if we ever use `voms-proxy-list`, which is not available in the java flavor, so I kept both.

amaltaro:
@khurtado don't we need any of the jdk and whatnot in addition to this java-based client?

I grepped WMCore and it looks like we don't really use `voms-proxy-list` anywhere, so we could consider removing it indeed.

amaltaro:
@khurtado do you know what is the size added  to the image with this new library?

amaltaro:
I see that you replied this in the other wmcore-docs ticket "voms-clients-java will increase size in: 192MB The cpp v2 version is only 1 MB". 

AFAICT, this voms-client is still fully functional - other than the exit status 1 and extra warning record - so I am inclined to say that we should remain with the current package (cpp) and loose a bit the `_renew_proxy` function. Am I missing anything else? Any other thoughts @khurtado @todor-ivanov ?

khurtado:
@amaltaro The v2 client is fully functional right now except for the exit status, but to me this is sort of a warning for us. We are still using 2 old VOMS servers in our infrastructure. There are 2 newer VOMS servers (with token features, etc) that our v2 client is not compatible with, and the older servers may get upgraded (or removed in favor of the newer ones as it has been done on FNAL host configurations already) and only the v3 client and the v2 client + an OSG patch for RedHat based distribution are compatible with them. Whether Debian updates their v2 client with the OSG patch and when are questions I can't answer.

This means we can easily workaround today's issue, but we are likely going to have a more serious issue later on. I think our containerized infrastructure should A) Use at least the same software we used in the hosts (or better with the proper validations) and B) Keep it as small as possible. Here, we are currently using and older VOMS client than used in the hosts, which benefits us for B, but we should consider A as well.

amaltaro (approved):
Thanks Kenyi! With your arguments here and in the other ticket under https://gitlab.cern.ch/dmwm/wmcore-docs/-/issues/2#note_8043213, I have to agree with you that this client is likely a positive long term investment, despite the almost 200MB increased size of the image.

We are likely cutting a new release tomorrow with this in, and another PR I have under CMSKubernetes. Thanks

PR #1498, opened 2024-05-31
Enable cron service; save env vars into /etc/environment

amaltaro:

Enable cron service in the Dockerfile;
And save environment variables into `/etc/environment`, which AFAIU is used when executing crontab.

### DIFF
--- a/docker/pypi/wmagent/run.sh
+++ b/docker/pypi/wmagent/run.sh
@@ -34,11 +34,17 @@ export WMA_USER=$wmaUser
 export USER=$wmaUser
 [[ -d ${HOME} ]] || mkdir -p ${HOME}
 
+echo "Saving environment variables into /etc/environment"
+printenv > /etc/environment && chmod 666 /etc/environment
+
 cp -f ${WMA_DEPLOY_DIR}/etc/wmagent_bashrc $HOME/.bashrc
 source $HOME/.bashrc
 
 echo "Start initialization"
 $WMA_ROOT_DIR/init.sh | tee -a $WMA_LOG_DIR/init.log || true
+
+echo "Starting cron service"
+sudo service cron restart
 
 echo "Start sleeping now ...zzz..."
 sleep infinity &
--- a/docker/pypi/wmagent/wmagent-docker-run.sh
+++ b/docker/pypi/wmagent/wmagent-docker-run.sh
@@ -81,6 +81,12 @@ if ! [ -f $HOST_MOUNT_DIR/admin/etc/group ]; then
     echo $groupEntry >> $HOST_MOUNT_DIR/admin/etc/group
 fi
 
+# workaround for being able to start cron service with ordinary user
+if ! [ -f $HOST_MOUNT_DIR/admin/etc/sudoers ]; then
+    echo "Creating sudoers file"
+    echo "$wmaUser ALL=(ALL:ALL) NOPASSWD: ALL" >> $HOST_MOUNT_DIR/admin/etc/sudoers
+fi
+
 # create regular mount points at runtime
 [[ -d $HOST_MOUNT_DIR/admin/wmagent ]] || (mkdir -p $HOST_MOUNT_DIR/admin/wmagent) || exit $?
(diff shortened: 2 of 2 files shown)


amaltaro:
It still fails with something like:
```
Starting cron service
sudo: /etc/sudoers is owned by uid xxx, should be 0
sudo: no valid sudoers sources found, quitting
sudo: error initializing audit plugin sudoers_audit
```

I am running out of options  on getting this fixed. So I will proceed with the validation and check on any suggestions on Monday.

vkuznet:
Alan, please review how crons are setup for our frontends, e.g.
- https://github.com/dmwm/CMSKubernetes/blob/master/docker/cmsweb-base/Dockerfile#L11
- https://github.com/dmwm/CMSKubernetes/blob/master/docker/frontend/Dockerfile#L27
- https://github.com/dmwm/CMSKubernetes/blob/master/docker/frontend/install.sh#L121
- https://github.com/dmwm/CMSKubernetes/blob/master/docker/frontend/install.sh#L127
- https://github.com/dmwm/CMSKubernetes/blob/master/docker/frontend/run.sh#L173

Botton line you'll need:
- cron daemon to be installed in an image
- sudoers to be installed in an image
- add your image user to sudoers
- setup your crons via crontab command, e.g. `echo ".... cron command" | crontab -`
- you need to start cron daemon via sudo, e.g. `sudo /usr/sinb/crond -n`, this can be your last command or you should correctly redirect it via `nohup` to detach from terminal.

amaltaro:
Thanks Valentin. If we knew the user beforehand, the we could set things up similar to what is done here.
In our case, containers run with a user that is defined during runtime, and this is what is making this setup difficult, mainly because:
* `cron` service can only be started by root
* or someone with sudo powers, but for that we would have to update `/etc/sudoers` file to add an entry for the runtime user. Which I do not know how to, given that runtime user has not permissions for that.

That's why in this PR I tried to mount an /etc/sudoers file, but then there is yet another issue when triggering a system service operation, as reported here https://github.com/dmwm/CMSKubernetes/pull/1498#issuecomment-2143451306

vkuznet:
> Thanks Valentin.
> 
> If we knew the user beforehand, then we could set things up similarly to what you have linked in your comment above. In our case, containers run with a user that is defined during runtime, and this is what is making this setup difficult, mainly because:
> 
> * `cron` service can only be started by root
> * or someone with sudo powers, but for that we would have to update `/etc/sudoers` file to add an entry for the runtime user. Which I do not know how to, given that runtime user has not permissions for that.
> 
> That's why in this PR I tried to mount an /etc/sudoers file, but then there is yet another issue when triggering a system service operation, as reported here [#1498 (comment)](https://github.com/dmwm/CMSKubernetes/pull/1498#issuecomment-2143451306)

Alan, you have two options here:
1. Either run crons outside of container image and let cron to write data to local file, then mount this file area within container such that image can use them. This is the case of CMSWEB k8s approach (we renew proxies, tokens via external crons and mount them to our containers)
2. Create separate user within your image and grant sudo permission to this user. You need to pass this user password at build time of the image such that you will assign it to this user. Then at run time of image you mount a new secret with this user password and use it for `su - <user>` access where you can then perform `sudo` operations. This seems more complicated but it should work since your user will be already available at run time.

Maintenance wise, I advice to use first approach which separates image workflow from crons. Please keep in mind that if you succeed to embed crons into your image and at run time one cron will crash your entire image must be restarted or you will need to manually login into your image to perform the restart of the cron which may not be possible. This is why separation of services into individual images is better than monolotic image running multiple services since you can manage them independently. The same argument goes with crons, you should be able to manage them independently from your image such that it will not require access to image or image restart.

todor-ivanov:
Hi @amaltaro  @vkuznet  I prefer the solution suggested  here: https://github.com/dmwm/CMSKubernetes/pull/1506

amaltaro:
Thanks Todor.
This PR has been superseeded by https://github.com/dmwm/CMSKubernetes/pull/1506, closing this one out.

PR #1504, opened 2024-06-11
Make APS run on standard 443 port and use X509 for auth

vkuznet:
To perform migration from CC7 to Alma9 we have two options:
- use Go based APS with x509 on standard port (443), or
- migrate apache with CMS patches

This PR implements necessary changes for former solution (APS+X509). I put appropriate comment about added options for APS. It also requires APS version 0.2.49 or above (current version is 0.2.50) where support for port option was provided. In this version the port option will overwrite port provided via configuration and therefore it allows us to re-use single configuration for both APS+X509 and XPS servers. The unification of configuration should be done within gitlab repo.

@arooshap please review and we can merge the changes. I'll provide further details via private chat.

### DIFF
--- a/kubernetes/cmsweb/daemonset/auth-proxy-server.yaml
+++ b/kubernetes/cmsweb/daemonset/auth-proxy-server.yaml
@@ -60,11 +60,15 @@ spec:
         # NOTE: to migrate CMSWEB frontend to Go APS using x509 auth on port 443
         #       we explicitly specify this port and -useX509 flag below
         #       when we'll transition to tokens we should remove these two options
+        # NOTE: we explicitly use port, metricsPort and logFile options
+        #       to re-use single config.json between APS and XPS
         args:
           - /data/auth-proxy-server
           - -config=/etc/secrets/config.json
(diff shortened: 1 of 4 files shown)


muhammadimranfarooqi:
Hi @vkuznet 

APS is deployed as Daemonsets in cmsweb. So, changes should be done here https://github.com/dmwm/CMSKubernetes/blob/master/kubernetes/cmsweb/daemonset/auth-proxy-server.yaml

vkuznet:
@muhammadimranfarooqi , thanks for reminder about daemon sets, I added additional changes over there too. I think we should keep consistent both services and daemonset aps manifest files and change them back to tokens together.

muhammadimranfarooqi:
Looks good to me

vkuznet:
@arooshap , following our MM discussion, I have added explicit `-port`, `-metricsPort` and `-logFile` options to allow re-use of `config.json` between APS and XPS.  Now this PR is ready. Please review and merge at your convenience.

arooshap (approved):

PR #1510, opened 2024-06-18
Add auth-proxy-server logstash rules

nikodemas:
I have updated the logstash rules for the `auth-proxy-server`. However, this filtering is used only if the `tag` is `aps` and the `tag` should come from the fileabeat, but I don't know were the filebeat is running, so I cannot really change the `tag`.

I have tested out the parsing assuming that the `tag` is set to`aps` and [here](https://codimd.web.cern.ch/s/Dxo16MEhh) you can see how the output for some old frontend log that I found in the docs (`tag` is `frontend` in this case) and the output of the auth-proxy-server. There are some other fields that come from the filebeat such as `ephemeral_id` or `agent`, so they are not getting filled in this testing.

### DIFF
--- a/kubernetes/cmsweb/daemonset/auth-proxy-server.yaml
+++ b/kubernetes/cmsweb/daemonset/auth-proxy-server.yaml
@@ -17,6 +17,7 @@ data:
       scan_frequency: 10s
       backoff: 5s
       max_backoff: 10s
+      tags: ["aps"]
     output.console:
       codec.format:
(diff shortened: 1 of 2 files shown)


vkuznet:
Nikodemas, thanks for providing these changes.

@arooshap , please keep this open until we will finish migration and we can apply this only to testbed to check how it will work.

arooshap (approved):

PR #1524, opened 2024-07-16
Adding new cronjob for fixing certificates issues.

arooshap:


docker/fetch-crl/update-crl.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+
+# Run fetch-crl to update CRLs
+/usr/sbin/fetch-crl
+
+# Copy updated CRLs to the appropriate directory
+cp /etc/grid-security/*.pem /host/etc/grid-security/
```
vkuznet, line 7: You may add -f flag to force overwrite.

vkuznet (changes requested):

arooshap:
> Can we add logging to the script such that we will have ability to inspect logs if necessary.

@vkuznet the fetch-crl command is very verbose. Other than that, we are only copying the certificates.

vkuznet (approved):

PR #1527, opened 2024-07-17
Increase memory limit for MSOutput and MSRuleCleaner to 1Gi

amaltaro:
Fixes https://github.com/dmwm/WMCore/issues/12042

It is a small memory limit increase, but it should help us to better deal with peak loads in these microservices, especially after outages and/or misbehaviors of the services.


arooshap:
@amaltaro sure, please proceed! :)

amaltaro:
Thanks Aroosha. Right now we cannot push in these limits because both MSOutput and MSRuleCleaner are under stress, and they were getting killed even with 1.25GB. I edit'ed the deployment template of those 2 services to limit it to 2GB.

I think we can roll it back to the 1GB provided in this PR, in the beginning of the next week. So we can either merge it now, or wait until the next week or so. Thanks

PR #1532, opened 2024-08-01
Ensure X509_USER_PROXY is set for MSUnmerged restarts

amaltaro:
Partial fix for https://github.com/dmwm/WMCore/issues/12061

Without these changes, whenever the service is restarted inside the pod, it fails to load the proxy file, hence also failing to interact with site storage endpoints.


arooshap (approved):

amaltaro:
Thanks @arooshap

amaltaro:
And I just noticed that without updating the dmwm-base image, these changes will not be picked up. I will work on that soon and provide a new PR.

PR #1534, opened 2024-08-01
Execute cron jobs via cron.d rather than crontab

khurtado:

This change will run cron as root + sudo (to the user), rather than having cron use `su` when forking the process (which led to Authentication errors at FNAL)

### DIFF
--- a/docker/pypi/wmagent/init.sh
+++ b/docker/pypi/wmagent/init.sh
@@ -324,10 +324,10 @@ set_cronjob() {
     local errVal=0
 
     # Populating proxy related cronjobs
-    crontab -u $WMA_USER - <<EOF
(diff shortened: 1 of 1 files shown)


amaltaro:
@khurtado if you tested these changes in both CERN and Fermilab agents and everything is working as expected, then we also need to update the CouchDB cron entries created by this script:
https://github.com/dmwm/WMCore/blob/master/bin/wmagent-couchapp-init#L121

todor-ivanov:
Hi @khurtado 
This change will break T0 agents, because they do not have access to `/etc`.

khurtado:
@amaltaro Will do
@todor-ivanov Thank you! Didn't notice that. I guess I will just keep the crontab -u method for Tier0, since they don't seem to have this issue

khurtado:
Closed in favor of https://github.com/dmwm/CMSKubernetes/pull/1538

PR #1538, opened 2024-08-16
Fix crontab issue with cmsdataops at FNAL

khurtado:

The wmagent user inside docker has authentication issues with `su`, which is needed and used by crontab.
This fixes that issue.


khurtado:
@amaltaro @todor-ivanov I found a simpler way to fix the crontab issue without system environment changes and without using `cron.d` instead. You will notice this basically sets a password to the wmaUser. I chose anything because I don't think it supposes a security issue, given that it's only for the user inside the container.

todor-ivanov:
thanks @khurtado

todor-ivanov (approved):
looks good to me

amaltaro (approved):
Thanks Kenyi! Next agents we deploy will come with this fix in the run script.

PR #1540, opened 2024-08-21
Update aps logstash for prod

nikodemas:
This should solve the issues reported recently related to the logstash of the APS:
- frontend host in the testbed always had port and in prod it sometimes doesn't, so making it optional
- some logs like the ones from `t0wmadatasvc` don't have `x_forwarded_host` and `x_forwarded_port`, so making them optional as well
- removing `system` replacement in case `api` is empty as it doesn't make sense
- stopping the discard of logs that failed to be properly parsed to catch edge cases, but maybe we can bring back this part after some time when we are sure about the logstash working properly

FYI @arooshap @vkuznet @leggerf

### DIFF
--- a/kubernetes/cmsweb/monitoring/logstash.conf
+++ b/kubernetes/cmsweb/monitoring/logstash.conf
@@ -78,7 +78,7 @@ filter {
   if "aps" in [tags] {
       mutate { replace => { "type" => "aps" } }
       grok {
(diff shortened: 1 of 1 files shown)


vkuznet:
thanks!!! Will wait when it will propagate to production and check the timber dashboard.

PR #1541, opened 2024-08-22
WMAgent: add component names to restartComponent script cron

amaltaro:
Component names got lost in the transition to container-based deployment, see:
https://github.com/dmwm/WMCore/blob/master/deploy/deploy-wmagent.sh#L430

This PR adds back the components that we want to watch and automatically restart whenever they go down.

### DIFF
--- a/docker/pypi/wmagent/init.sh
+++ b/docker/pypi/wmagent/init.sh
@@ -327,7 +327,7 @@ set_cronjob() {
     crontab -u $WMA_USER - <<EOF
 55 */12 * * * date -Im >> $WMA_LOG_DIR/renew-proxy.log && $WMA_MANAGE_DIR/manage renew-proxy 2>&1 >> $WMA_LOG_DIR/renew-proxy.log
 58 */12 * * * python $WMA_DEPLOY_DIR/deploy/checkProxy.py --proxy /data/certs/myproxy.pem --time 120 --send-mail True --mail alan.malta@cern.ch
-*/15 * * * *  source $WMA_DEPLOY_DIR/deploy/restartComponent.sh > /dev/null
+*/15 * * * *  source $WMA_DEPLOY_DIR/deploy/restartComponent.sh 2>&1 >> $WMA_LOG_DIR/component-restart.log
 EOF
     let errVal+=$?
 


docker/pypi/wmagent/init.sh
```diff
@@ -327,7 +327,7 @@ set_cronjob() {
     crontab -u $WMA_USER - <<EOF
 55 */12 * * * date -Im >> $WMA_LOG_DIR/renew-proxy.log && $WMA_MANAGE_DIR/manage renew-proxy 2>&1 >> $WMA_LOG_DIR/renew-proxy.log
 58 */12 * * * python $WMA_DEPLOY_DIR/deploy/checkProxy.py --proxy /data/certs/myproxy.pem --time 120 --send-mail True --mail alan.malta@cern.ch
-*/15 * * * *  source $WMA_DEPLOY_DIR/deploy/restartComponent.sh > /dev/null
+*/15 * * * *  source $WMA_DEPLOY_DIR/deploy/restartComponent.sh ErrorHandler JobSubmitter AgentStatusWatcher > /dev/null
```
vkuznet, line 330: Even though it is fine you may redirect stdout/stderr elsewhere which will allow quickly to look it up for debugging purposes. I suggest `2>&1 1>& /tmp/component.log`. If you will decide it is useful then may be you need to provide individual restarts which redirects its stdout/stderr to individual files.

amaltaro: I am not sure I understand which debugging information you would like to see. The component "ComponentLog" has all the information we need, in this script the only information we would gather is that component X was restarted (plus tail of the log).

Instead, we actually send an email notification with the restart information - and tail of the log, see:
https://github.com/dmwm/WMCore/blob/master/deploy/restartComponent.sh#L29-L32

I do think we should update the list of recipients though and ensure that all core developers are getting these emails.

vkuznet: Alan, the restartComponent.sh script performs this action:
```
$manage execute-agent wmcoreD --restart --components=$comp
```
which is not piped to any log. I think this is what is most interesting to see if there is a restart which  may fail. So, either add pipe to the log inside of this script or redirect entire stdout/stderr from this script to some place.

vkuznet: And we should not use personal emails in a script which is published to github. It is a way to get more spam in person's folders. Instead, put e-group email and assign people to get notification through it.

amaltaro: I could pipe the out/err of the script to a log file, but I don't think we gain anything with that. The reason is that if the component fails to start, it will be logged in the ComponentLog as well, which is tail'ed and sent out via email. If you still think it's important, I can add it.

About emails, I am sorry to say it, but every single Dockerfile in this repository has your address(!)
Nonetheless, I have provided the following WMCore changes: https://github.com/dmwm/WMCore/pull/12080

One day we will be using CERN gitlab for the agent configuration, then it will be easier to use such data without fear.

vkuznet: Alan, the `manage` command provides useful information about restart of the service and I'm not sure if everything is recorded to component log, e.g.
```
manage execute-agent wmcoreD --restart --components=WorkflowUpdater
execute_command_agent: Executing: wmcoreD --restart --components=WorkflowUpdater ...
Components List Specified:
[WorkflowUpdater]
Checking default database connection... ok.
Stopping components: ['WorkflowUpdater']
Stopping: WorkflowUpdater
Starting components: ['WorkflowUpdater']
Starting : WorkflowUpdater
Starting WorkflowUpdater as a daemon
Log will be in /data/srv/wmagent/2.3.4/install/WorkflowUpdater
Waiting 1 seconds, to ensure daemon file is created

started with pid 529262
```
we see pid, database connection check, etc. So, if it is in component log then I don't have any complains. Otherwise I suggest to record it.

Regarding emails. I hope we can discuss this issue during our regular WMCore meeting, and thanks for opening ticket to address it.

amaltaro:
I removed the component names because I think we could go ahead and enable this script for every single component of the agent, see this WMCore PR for further details: https://github.com/dmwm/WMCore/pull/12080

That means, this PR is simply adding a log file to receive the output of the cron script.

vkuznet (approved):

amaltaro:
Thanks Valentin!

PR #1545, opened 2024-08-30
CRABServer helm: deploy all crab services in single helm chart

novicecpp:
Fix https://github.com/dmwm/CRABServer/issues/8677
This change make it possible to deploy all crab related services in single helm chart.
The changes also including:
- Move `config.py` from secret to normal ConfigMap
- New `crabserver-canary` deployment manifest for https://github.com/dmwm/CRABServer/issues/8650
- New CI serviceaccount that need cluster admin to deploy it for us (only testbed cluster).

@arooshap We (CRAB) are already reviewed this changes and I would like you to review the template part. I am not helm chart expert so I am not sure if I am do it right. You can merge if there is no change needed.

Tested in cmsweb-test12 cluster.


novicecpp:
@arooshap Done

PR #1546, opened 2024-09-01
Script to test APS endpoints.

arooshap:


kubernetes/cmsweb/scripts/test_k8s_aps.sh
```diff
@@ -0,0 +1,119 @@
+#!/bin/bash
+##H Usage: test_k8s_aps.sh <base_url>
+##H
+
+if [ "$1" == "-h" ] || [ "$1" == "-help" ] || [ "$1" == "--help" ] || [ "$1" == "help" ] || [ "$1" == "" ]; then
+    perl -ne '/^##H/ && do { s/^##H ?//; print }' < $0
```
vkuznet, line 6: Please do not make dependency on perl, it requires to have perl installed on a system just for one line parsing. Instead use standard UNIX tools like grep, sed, awk. For example, this line can be replaced with `grep "##H" $0 | grep -v grep | sed -e "s,##H,,g"` and it does not require perl at all.

kubernetes/cmsweb/scripts/test_k8s_aps.sh
```diff
@@ -0,0 +1,119 @@
...
+base_url=$1
+
+# Function to perform the request with automatic redirect following
+perform_request() {
+    local target_url="$1"
+    final_status_code=$(curl -L -i --key $X509_USER_KEY --cert $X509_USER_CERT -o /dev/null -w '%{http_code}' -s "$target_url")
+    echo "$target_url HTTP code: $final_status_code"
+}
+
+# CouchDB
```
vkuznet, line 22: You have repeated pattern, i.e. you execute `perform_request` over provided list of URL end-points. So, generalize the script and remove individual lists. This way the script will be much smaller and it can take external file (list of end-points). We'll keep such list separate from the file and can adjust it as necessary, but the script itself will stay intact and only take input list.

vkuznet (changes requested):

arooshap:
@vkuznet changed the script to use a file, while keeping it readable, also removed the perl dependency..

```
[apervaiz@lxplus916 scripts]$ ./test_k8s_aps.sh https://cmsweb-testbed.cern.ch urls.txt

== CouchDB ==
  https://cmsweb-testbed.cern.ch/couchdb/ -> HTTP code: 200

== Crabserver ==
  https://cmsweb-testbed.cern.ch/crabserver/preprod/info -> HTTP code: 200
  https://cmsweb-testbed.cern.ch/scheddmon/059/crabint1/ -> HTTP code: 200

== DAS ==
  https://cmsweb-testbed.cern.ch/das -> HTTP code: 200

== DBS ==
...
```

kubernetes/cmsweb/scripts/test_k8s_aps.sh
```diff
@@ -0,0 +1,52 @@
...
+# Check if required arguments are provided
+if [ $# -ne 2 ]; then
+    echo "Error: You need to provide <base_url> and <url_list_file>"
+    grep "##H" $0 | sed -e "s,##H,,g"
+    exit 1
+fi
+
+X509_USER_KEY=$HOME/.globus/userkey.pem
+X509_USER_CERT=$HOME/.globus/usercert.pem
+
```
vkuznet, line 19: Aroosha, please add check for X509_USER_PROXY env, if it is set we should use it instead of X509_USER_KEY/CERT since it is a proxy rather a user certificate. We may need to test both from time to time.

vkuznet (changes requested):

arooshap:
@vkuznet I added both functionalities.. is it better?

PR #1549, opened 2024-09-16
New Alma9 wmcore-dev image

d-ylee:
Currently used in the Jenkins EL9/Docker based testing found in https://github.com/d-ylee/jenkins-test

The image will grab the latest requirements.txt from the https://github.com/dmwm/WMCore repository.

Not all scripts in ContainerScripts and TestScripts are currently used. All of these scripts were pulled from the cmsdocks/dmwm:wmcorepy3_tests image. Some of these will be removed, once deemed unneeded for Jenkins testing.

### DIFF
--- a/docker/pypi/wmcore-dev/.dockerignore
+++ b/docker/pypi/wmcore-dev/.dockerignore
@@ -1,0 +1,5 @@
+*
+!TestScripts
+!ContainerScripts
+!entrypoint.sh
+!etc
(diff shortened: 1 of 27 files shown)


amaltaro:
@d-ylee just a different idea on the python files that are shipped with this image.

Instead of adding them to the image, perhaps we could have a script that would be cloning a given repository, and in that repo we store all of these python/bash files.

d-ylee:
@amaltaro I think that could be a good idea for a future wmcore-dev release. I also am adding some UserScripts/DeveloperScripts that could help set up a containerized environment for developers.

khurtado (approved):
This has been tested already and it is fully functional, so approve. I believe any potential future improvements should be part of a new ticket.

anpicci:
@amaltaro Do you think this is in good shape to merge?

PR #1558, opened 2024-10-23
Properly deal with WMAgent images in the localhost and tagged as latest

amaltaro:
Fixes https://github.com/dmwm/WMCore/issues/12157

Summary of changes is:
* whenever user asks to run `latest` WMAgent tag, in addition to using the `latest` tag from localhost, it also identifies which `WMA_VER_RELEASE` is supposed to be used.
* for localhost images, name registry after the CERN registry, as we don't have anything called `local`

### DIFF
--- a/docker/pypi/wmagent/wmagent-docker-run.sh
+++ b/docker/pypi/wmagent/wmagent-docker-run.sh
@@ -30,6 +30,11 @@ usage(){
 
 PULL=false
 WMA_TAG=latest
+# This is the root at the host only, it may differ from the root inside the container.
+# NOTE: This is parametriesed, so that the container can run on a different mount point.
+#       A soft link is needed to mimic the same /data tree as inside the container so
+#       that condor may find the job cache and working directories:
+HOST_MOUNT_DIR=/data/dockerMount
 
 ### Argument parsing:
 while getopts ":t:hp" opt; do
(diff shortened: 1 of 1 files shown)


amaltaro:
The basic assumption is that the image has already been pulled down. Based on that, I tested variations like:
```
./wmagent/wmagent-docker-run.sh -t latest -p
```
and
```
./wmagent/wmagent-docker-run.sh -t latest
```
and
```
./wmagent/wmagent-docker-run.sh
```

and it seems to be working fine now.

todor-ivanov (approved):
looking good. thanks @amaltaro

mapellidario:
Before I leave a review, I have a couple of questions. 

I fail to see the purpose of these changes. It seems that when you ask for the `latest` WMA_TAG, you define WMA_VER_REALEASE based on what was previously installed on the machine. Is this really the desired behaviour? why pulling the latest image from the registry should use something that is already present on the machine? when do you envision this feature to be used?

moreover, I run your new command to get the version of the releases on `submit13`, where data from multiple releases is present, and I think that the result is not what you want [1]

---

[1]

```plaintext
[cmsdataops@cmsgwms-submit13 ~]$ WMA_VER_RELEASE=$(ls -l /data/dockerMount/srv/wmagent | grep -v ^l | grep -v total | awk '{print $9}')
[cmsdataops@cmsgwms-submit13 ~]$ echo $WMA_VER_RELEASE
2.3.4 2.3.7
```

amaltaro:
The goal is to provide non-WM expert people with a generic command that they can use in case they have to manipulate the node running WMAgent services. For instance, every now and then the Fermilab team will reboot the node for upgrades. Usually they stop the services, reboot the node, restart services. So it would be great to give them a static command that does not depend on the version currently running.

For submit13, maybe I should have mention that we are expected to have only 1 WMAgent area. Or, I could deal with this in the code as well by checking the latest timestamp or something like that.

mapellidario:
I see, thanks alan for the explanation! maybe `current` would be a better word instead of `latest` in this context, but ok it's not a problem and I am not asking about changing anything, I just need to catch up with what the goal is. sounds good then :)

mapellidario (approved):
thanks alan!

amaltaro:
Thanks guys!

mapellidario:
I am puzzled. how am i supposed to run a docker image that i built locally? 

In particular: I built an image with

```plaintext
cmst1@vocms0262:pypi $ ./wmagent/wmagent-docker-build.sh -t 2.3.7.1
```

then I have the following images in my agent, notice how the image with ID `257854303f19` does not exist on the remote registry

```plaitnext
cmst1@vocms0262:wmagent $ docker image ls
REPOSITORY                                TAG       IMAGE ID       CREATED         SIZE
local/wmagent                             2.3.7.1   257854303f19   8 minutes ago   2.3GB
local/wmagent                             latest    257854303f19   8 minutes ago   2.3GB
local/wmagent                             2.3.7     9813a1586e4e   8 days ago      2.3GB
registry.cern.ch/cmsweb/wmagent           2.3.7     9813a1586e4e   8 days ago      2.3GB
registry.cern.ch/cmsweb/wmagent           2.3.7.1   181d431f7144   2 weeks ago     2.3GB
registry.cern.ch/wmagent                  2.3.7.1   181d431f7144   2 weeks ago     2.3GB
registry.cern.ch/wmagent                  latest    181d431f7144   2 weeks ago     2.3GB
[...]
```

and notice how we use `local` to tag a docker image that has been built on the wmagent itself, for testing purposes, and has not been pushed to the remote. If this is a usecase that we are willing to support, I will open an issue and a PR, otherwise i will create my own fork of the `wmagent-docker-run.sh` script

amaltaro:
Oups, it looks like I made too many changes in my change, sorry.

I forgot that this script could be used to run images built locally - for testing/dev purposes. Dario, if you know what the fix is supposed to be, please go for it. Otherwise I can provide a PR to fix that.

amaltaro:
@mapellidario you might want to check out this PR: https://github.com/dmwm/CMSKubernetes/pull/1560
Even better if you can test it ;)

PR #1559, opened 2024-10-25
Prepare udp-collector for Kubernetes

nikodemas:
Currently `udp-collector` is running on a virtual machine. These changes together with the PR https://github.com/dmwm/udp-collector/pull/5 moves it to the Kubernetes infrastructure.
Few things to note:

- the service is pretty lightweight (memory rarely never goes over 1GiB, more metrics can be seen [here](https://monit-grafana.cern.ch/goto/uTr5UGWHR?orgId=1)), so there is no need to create a separate cluster for it and just an extra node on a cluster that already exists is enough
- for the same reason there is no need to use an ingress controller and simply using the `hostPort` option on the service is enough
- service is currently deployed on a test cluster and is working fine. after these PRs are approved, I will deploy it in production, but will start sending the traffic there only after we pass all of the security checks

FYI @vkuznet @leggerf @brij01


docker/udp-server/Dockerfile
```diff
@@ -1,13 +1,23 @@
...
+    CGO_ENABLED=0
+
+WORKDIR $WDIR
+
+RUN git clone https://github.com/dmwm/udp-collector && \
+    cd udp-collector && \
+    if [ "$TAG" = "v00.00.00" ]; then git checkout master; else git checkout tags/$TAG -b build; fi && \
+    make
+
+FROM alpine:3.14
```
vkuznet, line 15: Since UDP server compiles into static executable I suggest to use distroless image which does not provide any shell in it, and neither contains any Linux stack, it is bare kernel and it tights security to its maximum. We use this for our APS frontends, see https://github.com/dmwm/CMSKubernetes/blob/master/docker/auth-proxy-server/Dockerfile#L19 and comment about with URL explaining what it is.

In new k8s cluster (Aroosha can tell you which version) it supports `kubectl debug` command which will allow to get access to distroless image from another container. Doing this way you image will have always zero vulnerability and since we expose UDP server to outside world CERN security team will welcome that.

kubernetes/monitoring/services/udp-collector/udp-server.yaml
```diff
@@ -28,8 +14,13 @@ spec:
       labels:
         app: udp-server
```
vkuznet, line 15: I suggest to expose metrics port for Prometheus scraping, see https://github.com/dmwm/CMSKubernetes/blob/master/kubernetes/cmsweb/services/das-server.yaml#L86 and use appropriate env too

vkuznet (changes requested):

nikodemas:
@vkuznet thanks a lot for the comments. I have updated the code as per your requests and updated the `udp-collector` accordingly.

docker/udp-server/Dockerfile
```diff
@@ -1,13 +1,24 @@
...
+ARG TAG=v00.00.00
+ENV WDIR=/data \
+    CGO_ENABLED=0
+
+WORKDIR $WDIR
+
+RUN git clone https://github.com/dmwm/udp-collector && \
+    cd udp-collector && \
+    if [ "$TAG" = "v00.00.00" ]; then git checkout master; else git checkout tags/$TAG -b build; fi && \
+    make
```
vkuznet, line 13: I suggest to double check that you indeed compile static executable. In your Dockerfile you setup CGO_ENABLED in ENV, I usually put it in front of make, e.g. `CGO_ENABLED=0 make`. Just check with `ldd` command that final executable indeed a static one.

vkuznet (changes requested):

vkuznet (approved):

PR #1560, opened 2024-11-06
Recover local wmagent registry option

amaltaro:
Complement to https://github.com/dmwm/CMSKubernetes/pull/1558

It recovers the `local` option, in case someone builds a WMAgent image in the same node that it is going to run.


mapellidario:
I tested it and it works, so I will merge it :)

PR #1561, opened 2024-11-07
wmagent - do not init if couch is not empty

mapellidario:
Following up on an operation mishap where I and @anpicci  deployed a new fnal wmagent with a non empty couchdb and Alan suggested to drain it soon not to encounter performance issues down the line, I decided that it was time to add some checks to our scripts.

I know that these changes are not really "required" but more of a "nice to have" , it was nonetheless a fun exercise to discover how we build our docker images and explore our init scripts.

This PR provides the following changes:

- I added `jq` to wmagent-base. it adds [100kB](https://packages.debian.org/bookworm/jq) to our docker image size, so i do not consider it a problem
- I built and pushed a new version of this image to [harbour](https://registry.cern.ch/harbor/projects/1771/repositories/wmagent-base/artifacts-tab) with tag `registry.cern.ch/cmsweb/wmagent-base:pypi-20241107-stable`
- I added a new `_couch_db_isclean` function to `manage-common.sh` script. after we check the connection to the local couch db, we retrieve some statistics about that instance. if it contains 1 document and 0 deleted documents, then we proceed with agent initialization. otherwise we stop the init.sh script and print how to manually clean the local couchdb.

Status:

- [x] tested that we can start an agent with an empty couch
- [x] tested that the commands used in this PR detect if a couch is not empty
- [ ] not tested yet that the init.sh script exits when couch is not empty, but i have no reason to believe it does not work :)

### DIFF
--- a/docker/pypi/wmagent-base/Dockerfile
+++ b/docker/pypi/wmagent-base/Dockerfile
@@ -6,7 +6,7 @@ RUN apt-get update
 RUN apt-get install -y libmariadb-dev-compat libmariadb-dev apache2-utils cron mariadb-server myproxy voms-clients voms-clients-java rlwrap libaio1 && apt-get clean
 
 # Install some debugging tools
-RUN apt-get install -y hostname net-tools iputils-ping procps && apt-get clean
+RUN apt-get install -y hostname net-tools iputils-ping procps jq && apt-get clean
 
 # Install recursive ps utility tool
 RUN apt-get install -y pslist && apt-get clean
--- a/docker/pypi/wmagent/bin/manage-common.sh
+++ b/docker/pypi/wmagent/bin/manage-common.sh
@@ -219,6 +219,31 @@ _sql_db_isclean(){
     esac
 }
 
+_couch_db_isclean(){
+  # assume that we can connect to local CouchDB
+  # assume that env contains local couch user and password.
+  echo "$FUNCNAME: Checking if local CouchDB is empty (curl to wmagent_summary db name)"
+  agent_summary=$(curl -s -u "$COUCH_USER:$COUCH_PASS" http://$COUCH_HOST:$COUCH_PORT/wmagent_summary)
+
+  if ! type jq > /dev/null ; then
+    echo "$FUNCNAME: jq is not installed, we cannot check if CouchDB is empty. Skipping this step."
+    return $(true)
+  fi
+  agent_summary_notclean=$(jq "select(.doc_count > 1 or .doc_del_count > 0)" <<< $agent_summary)
+  
+  if [ -n "${agent_summary_notclean}" ]; then
+    # the variable is not empty -> means that couch is not empty
+    echo "$FUNCNAME: local CouchDB is not empty. We cannot init WMAgent"
+    doc_count=$(jq ".doc_count" <<< $agent_summary)
+    doc_del_count=$(jq ".doc_del_count" <<< $agent_summary)
+    echo "$FUNCNAME: it contains $doc_count docs, $doc_del_count deleted docs"
+    return $(false)
+  else
+    # the variable is empty -> means that couch is empty
+    echo "$FUNCNAME: local CouchDB is empty. We can init WMAgent"
+  fi
+}
+
 _sql_write_agentid(){
     # Auxiliary function to write the current agent build id into the sql database
     echo "$FUNCNAME: Preserving the current WMA_BUILD_ID and HostName at database: $wmaDBName."
--- a/docker/pypi/wmagent/init.sh
+++ b/docker/pypi/wmagent/init.sh
@@ -284,8 +284,10 @@ _check_couch() {
     echo "$FUNCNAME: Checking whether the CouchDB database is reachable..."
     _status_of_couch || return $(false)
 
-    # echo "$FUNCNAME: Additional checks for CouchDB:"
+    echo "$FUNCNAME: Additional checks for CouchDB:"
+    local cleanMessage="You may consider cleaning local couchdb with 'rm -rf /data/dockerMount/srv/couchdb/'"
     # NOTE: To implement any additional check to the CouchDB similar to the relational databases
+    _couch_db_isclean || { echo "$FUNCNAME: Error: non empty local CouchDB. $cleanMessage"; return $(false); }
 }
 
 check_databases() {
@@ -313,7 +315,10 @@ check_databases() {
     esac
 
     # Checking CouchDB:
-    _check_couch
+    [[ -n $COUCH_HOST ]] && [[ -n $COUCH_PORT ]] && [[ -n $COUCH_USER ]] && [[ -n $COUCH_PASS ]] &&
+      couchdbCred=true
+    $couchdbCred || { echo "$FUNCNAME: ERROR: No local CouchDB credentials provided at $WMA_SECRETS_FILE" ; return $(false); }
+    _check_couch 
     echo "-----------------------------------------------------------------------"
 }
 
--- a/docker/pypi/wmagent/Dockerfile
+++ b/docker/pypi/wmagent/Dockerfile
@@ -1,5 +1,5 @@
 # syntax=docker/dockerfile-upstream:master
-FROM registry.cern.ch/cmsweb/wmagent-base:pypi-20240923-stable
+FROM registry.cern.ch/cmsweb/wmagent-base:pypi-20241107-stable
 MAINTAINER Valentin Kuznetsov vkuznet@gmail.com
 
 # TAG to be passed at build time through `--build-arg TAG=<WMA_TAG>`. Default: None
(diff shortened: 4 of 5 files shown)


docker/pypi/wmagent-base/Dockerfile
```diff
@@ -6,7 +6,7 @@ RUN apt-get update
 RUN apt-get install -y libmariadb-dev-compat libmariadb-dev apache2-utils cron mariadb-server myproxy voms-clients voms-clients-java rlwrap libaio1 && apt-get clean
 
 # Install some debugging tools
-RUN apt-get install -y hostname net-tools iputils-ping procps && apt-get clean
+RUN apt-get install -y hostname net-tools iputils-ping procps jq && apt-get clean
```
amaltaro, line 9: I feel like it would be even better to add this dependency to the `dmwm-base` image, but that requires then much more effort to update everything else. So I think it's fine to keep it here, eventually when we update dmwm-base, we can add it there and remove it from here.

docker/pypi/wmagent/bin/manage-common.sh
```diff
@@ -219,6 +219,31 @@ _sql_db_isclean(){
     esac
 }
 
+_couch_db_isclean(){
+  # assume that we can connect to local CouchDB
+  # assume that env contains local couch user and password.
+  echo "$FUNCNAME: Checking if local couch db is empty"
```
amaltaro, line 225: Perhaps add a detail in parenthesis like "(checking wmagent_summary db name)"?
In addition, I would suggest writing "CouchDB" in all these print statements that you added.

docker/pypi/wmagent/bin/manage-common.sh
```diff
@@ -219,6 +219,31 @@ _sql_db_isclean(){
...
+
+  if ! type jq > /dev/null ; then
+    echo "$FUNCNAME: jq is not installed, we can not check if couchdb is empty"
+    return $(false)
+  fi
+  agent_summary_notclean=$(jq "select(.doc_count > 1 or .doc_del_count > 0)" <<< $agent_summary)
+  
+  if [ -n "${agent_summary_notclean}" ]; then
+    # the variable is not empty -> means that couch is not empty
+    echo "$FUNCNAME: local couch is not empty. we can not init wmagent"
```
amaltaro, line 236: "We cannot init WMAgent"?

docker/pypi/wmagent/bin/manage-common.sh
```diff
@@ -219,6 +219,31 @@ _sql_db_isclean(){
...
+    doc_count=$(jq ".doc_count" <<< $agent_summary)
+    doc_del_count=$(jq ".doc_del_count" <<< $agent_summary)
+    echo "$FUNCNAME: it contains $doc_count docs, $doc_del_count deleted docs"
+    return $(false)
+  else
+    # the variable is empty -> means that couch is empty
+    echo "$FUNCNAME: local couch db is empty. we can init wmagent"
```
amaltaro, line 243: "We cannot init WMAgent"?

docker/pypi/wmagent/bin/manage-common.sh
```diff
@@ -219,6 +219,31 @@ _sql_db_isclean(){
...
 
+_couch_db_isclean(){
+  # assume that we can connect to local CouchDB
+  # assume that env contains local couch user and password.
+  echo "$FUNCNAME: Checking if local couch db is empty"
+  agent_summary=$(curl -s -u "$COUCH_USER:$COUCH_PASS" http://$COUCH_HOST:$COUCH_PORT/wmagent_summary)
+
+  if ! type jq > /dev/null ; then
+    echo "$FUNCNAME: jq is not installed, we can not check if couchdb is empty"
+    return $(false)
```
amaltaro, line 230: To increase the compatibility with the current images, I would suggest to simply skip this check if `jq` is not available (instead of failing the check).

docker/pypi/wmagent/init.sh
```diff
@@ -313,7 +315,10 @@ check_databases() {
     esac
 
     # Checking CouchDB:
-    _check_couch
+    [[ -n $COUCH_HOST ]] && [[ -n $COUCH_PORT ]] && [[ -n $COUCH_USER ]] && [[ -n $COUCH_PASS ]] &&
+      couchdbCred=true
+    $couchdbCred || { echo "$FUNCNAME: ERROR: No couch credentials provided at $WMA_SECRETS_FILE" ; return $(false); }
+    _check_couch || return
```
amaltaro, line 321: Dario, other than skipping the next line, was there any other reason to add this `return` statement here?

amaltaro (changes requested):
@mapellidario these changes are looking good in general. I left a few aesthetic comments and another question or two as well for your consideration.

docker/pypi/wmagent-base/Dockerfile
mapellidario: As you wish, I am fine either way. I put it here because it is not "necessary", it is a "nice to have" and felt consistent with these "debugging tools". but as you wish, i can move it to dmwm-base image and rebuild other images if you prefer this way :)

docker/pypi/wmagent/bin/manage-common.sh
mapellidario: oook, agree, done

docker/pypi/wmagent/bin/manage-common.sh
mapellidario: I mean, I agree that i can use WMAgent, but still if the local CouchDB is empty then we can proceed and init the agent

docker/pypi/wmagent/bin/manage-common.sh
mapellidario: ok, i agree, however keep in mind that this script is copied [1] inside the docker image, so nothing is going to change for existing images

---

[1] https://github.com/dmwm/CMSKubernetes/blob/e1106b4aa00fef9b1a655c6505ea20a8d7e5cbe1/docker/pypi/wmagent/Dockerfile#L41

mapellidario:
I added the suggested changes that I agree with and squashed them into the previous commit.

If you also want me to implement the two unresolved comments then feel free to say so and i will take care.

docker/pypi/wmagent/init.sh
amaltaro: I think the dashed lines in the echo below give us a nice separation in the logs. If it is not separated (with a dashed line) by a different method, I would be in favor of removing the return such that we execute the echo statement below.

amaltaro (approved):
Other than the one comment left in the code, these changes look good to me.
I would leave any extra work in this script for a future occasion.

docker/pypi/wmagent/init.sh
mapellidario: oook

mapellidario:
Thanks Alan for the review! I will merge this PR. we can move jq from the wmagen-base to dmwm-base next time, if the need for having it in another image emerges

docker/pypi/wmagent/init.sh
```diff
@@ -313,7 +315,10 @@ check_databases() {
     esac
 
     # Checking CouchDB:
-    _check_couch
+    [[ -n $COUCH_HOST ]] && [[ -n $COUCH_PORT ]] && [[ -n $COUCH_USER ]] && [[ -n $COUCH_PASS ]] &&
```
todor-ivanov, line 318: @mapellidario , add a default value of:
 ```
 couchdbCred=false
```
at the line just above this one. Otherwise it will break the execution with a syntax error on line 321, in the cases when even one of the CouchDb credential variables is missing  at `WMAgent.secrets`

File: docker/pypi/wmagent/bin/manage-common.sh [hunk 8]
```diff
@@ -219,6 +219,31 @@ _sql_db_isclean(){
     esac
 }
 
+_couch_db_isclean(){
+  # assume that we can connect to local CouchDB
+  # assume that env contains local couch user and password.
+  echo "$FUNCNAME: Checking if local CouchDB is empty (curl to wmagent_summary db name)"
+  agent_summary=$(curl -s -u "$COUCH_USER:$COUCH_PASS" http://$COUCH_HOST:$COUCH_PORT/wmagent_summary)
+
+  if ! type jq > /dev/null ; then
```
todor-ivanov, line 228: you could write: 
```
type jq > /dev/null || { echo "$FUNCNAME: ..." ; return $(false) ;}

```
and if you wanted to suppress also the stdErr the line should have been actually:
```
type jq > /dev/null 2>&1 || { echo "$FUNCNAME: ..." ; return $(false) ;}
```

just a suggestion, though, no need to follow on this one, but the way how it is right now the `stdError` for sure won't be redirected to `/dev/null`

todor-ivanov (commented):
looks good. I left only tow comments inline.

docker/pypi/wmagent/init.sh
mapellidario: I implemented this comment in another PR :)

PR #1563, opened 2024-11-08
wmagent - stop init if local couchdb is not empty - take2

mapellidario:
follow up to #1561 , implementing todor's suggestions


todor-ivanov (approved):
looking good

mapellidario:
thanks todor! <3

PR #1564, opened 2024-11-14
Avoid -u parameter when using the OS crontab for WMAgent

todor-ivanov:
Fixes: https://github.com/dmwm/WMCore/issues/12166

With the current change we no longer  try to use the `-u` parameter when the OS `crontab` is used. This is vital for the virtual environment setups, because it breaks with:
```
must be privileged to use -u
``` 

#### Related PRs
https://github.com/dmwm/WMCore/pull/12167

### DIFF
--- a/docker/pypi/wmagent/init.sh
+++ b/docker/pypi/wmagent/init.sh
@@ -332,11 +332,14 @@ set_cronjob() {
     # Avoid -u parameter when using the OS crontab
     _is_venv && crontabParams="" || crontabParams="-u $WMA_USER"
 
+    # Activate the environment if cronjobs are to run under virtual env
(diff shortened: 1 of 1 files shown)


todor-ivanov:
I tested it for prod WMAgnts. @LinaresToine  could you please test it also for T0 ?

LinaresToine:
The changes work correctly during a deployment of the agent. Thank you for providing this solution @todor-ivanov

mapellidario (approved):
thanks todor, looks good to me!

todor-ivanov:
ok, While further testing the solution I figured out two more, minor but important differences  between the docker and virtual env cronjobs at runtime. I resolved them in the next commit. @mapellidario  @LinaresToine  please take another look, before we merge.

mapellidario (approved):
Thanks Todor, looks good to me!

PR #1566, opened 2024-11-21
Use s-nail host mail server (FNAL/CERN)

khurtado:
Fixes https://github.com/dmwm/WMCore/issues/12159

Issue: Both CERN and FNAL already run mail agents at the host-level. Deploying a mail server inside docker (like exim4) conflicts with the host-server, since both would operate on ports 25. Running an independent mail agent on different port can lead to firewall issues as well. 

Solution: Both CERN and FNAL use the same mail agent (s-nail). Therefore, configure the containers to use the host mail agent instead.

### DIFF
--- a/docker/pypi/wmagent/wmagent-docker-run.sh
+++ b/docker/pypi/wmagent/wmagent-docker-run.sh
@@ -72,17 +72,37 @@ ln -s $HOST_MOUNT_DIR/srv/wmagent /data/srv/wmagent
 passwdEntry=$(getent passwd $wmaUser | awk -F : -v wmaHome="/home/$wmaUser" '{print $1 ":" $2 ":" $3 ":" $4 ":" $5 ":" wmaHome ":" $7}')
 groupEntry=$(getent group $wmaGroup)
 
+# @TODO: Create needed unix accounts in container, rather than bind mounting files from the host
 # workaround case where Unix account is not in the local system (e.g. sssd)
 [[ -d $HOST_MOUNT_DIR/admin/etc/ ]] || (mkdir -p $HOST_MOUNT_DIR/admin/etc) || exit $?
+
+# Validation step
+# Delete local docker passwd/group files if uucp is not present (so it can be recreated)
+[[ -f "$HOST_MOUNT_DIR/admin/etc/passwd" ]] && ! grep -q uucp "$HOST_MOUNT_DIR/admin/etc/passwd" &&  {
+        rm $HOST_MOUNT_DIR/admin/etc/passwd
+}
+[[ -f "$HOST_MOUNT_DIR/admin/etc/group" ]] && ! grep -q uucp "$HOST_MOUNT_DIR/admin/etc/group" &&  {
+        rm $HOST_MOUNT_DIR/admin/etc/group
+}
+
 if ! [ -f $HOST_MOUNT_DIR/admin/etc/passwd ]; then
     echo "Creating passwd file"
     getent passwd > $HOST_MOUNT_DIR/admin/etc/passwd
     echo $passwdEntry >> $HOST_MOUNT_DIR/admin/etc/passwd
+    # add back original system related unix users
+    echo "Debian-exim:x:103:105::/var/spool/exim4:/usr/sbin/nologin" >> $HOST_MOUNT_DIR/admin/etc/passwd
+    echo "_apt:x:100:65534::/nonexistent:/usr/sbin/nologin" >> $HOST_MOUNT_DIR/admin/etc/passwd
+    echo "uucp:x:10:10:uucp:/var/spool/uucp:/usr/sbin/nologin" >> $HOST_MOUNT_DIR/admin/etc/passwd
 fi
 if ! [ -f $HOST_MOUNT_DIR/admin/etc/group ]; then
     echo "Creating group file"
     getent group > $HOST_MOUNT_DIR/admin/etc/group
     echo $groupEntry >> $HOST_MOUNT_DIR/admin/etc/group
+    # add back original system related groups
+    echo "Debian-exim:x:105:" >> $HOST_MOUNT_DIR/admin/etc/group
+    echo "messagebus:x:104:" >> $HOST_MOUNT_DIR/admin/etc/group
+    echo "crontab:x:103:" >> $HOST_MOUNT_DIR/admin/etc/group
+    echo "uucp:x:10:" >> $HOST_MOUNT_DIR/admin/etc/group
 fi
 
 # create regular mount points at runtime
@@ -116,6 +136,7 @@ $tnsMount \
 --mount type=bind,source=/etc/sudoers.d,target=/etc/sudoers.d,readonly \
 --mount type=bind,source=/etc/grid-security,target=/etc/grid-security,readonly \
 --mount type=bind,source=/etc/vomses,target=/etc/vomses,readonly \
+--mount type=bind,source=/etc/s-nail.rc,target=/etc/s-nail_host.rc,readonly \
 "
 
 registry=local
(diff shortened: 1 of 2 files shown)


todor-ivanov:
> My opinion is that this change here should workaround exim4, but in the future a general solution would be to stop bind mounting /etc/group and /etc/passwd into the docker container and somehow create them at build time, in order to avoid services like cron or exim4 failing. 


I completely agree with you Kenyi

todor-ivanov (approved):
looking good

amaltaro:
@khurtado can you please also test this at FNAL? Once we confirm this is working in both CERN and FNAL (and virtual env, but Todor already confirmed it works for him), then we can move forward and also work on a new hot-patch for the agents.

amaltaro:
@khurtado I am not sure I have already mentioned this, but if needed, we should also consider reaching out to the CERN/FNAL teams and request new packages/services/etc to be installed. In case that would help to resolve this issue.

khurtado:
@amaltaro @todor-ivanov 
I gave up on fixing exim4 because 

1) Running on port 25 is not possible, since it is already taken (even though it worked at CERN somehow)
2) Using a different port would require firewall settings at the host-level

So, I changed the solution to use the host-level mail agent instead.

This current solution is working both at CERN and FNAL.

There is one thing I would likely change though:
This line:
https://github.com/dmwm/CMSKubernetes/pull/1566/files#diff-d276e67ac0a9cb57c1f967acb04ff6f223d21bb3cded596c8de279941c934dcfR178

Installs the s-nail package at runtime.
I would rather add this line as part of the wmagent-base image:

https://github.com/dmwm/CMSKubernetes/blob/master/docker/pypi/wmagent-base/Dockerfile

But of course, for that I would need to 
1) Add that line as a PR
2) Wait for someone to approve merge
3) Produce a new wmagent-base
4) Adjust this line accordingly to use the new base
https://github.com/dmwm/CMSKubernetes/blob/master/docker/pypi/wmagent/Dockerfile#L2

If you guys approve the rest of the code here and prefer the s-nail installation in the base image, let me know and I can create the PR to proceed with the steps above.

vkuznet:
Alan, please keep in mind the following:
- implementing in Python does not mean it will be easier to maintain. Over time learnt that python implementation is not the easiest to maintain, quite the opposite (sorry, it is my opinion). Too many complexities, dependencies, run-time environment, setup, etc.
- What if we'll eventually write a service in another language, or use another (Python) framework. How custom python module will be integrated and handle this?
- It is unclear what are requirements?

Here is my suggestion:
- write small HTTP server (preferably in Go) and put it on CMSWEB
  - use **ANY** HTTP client to make POST request to such service
  - use HTTP GET API to get alerts as JSON docs
  - optionally: write service web UI to display alerts as dashboard (there are plenty of ready-to-use JS libraies)
  - optionally, add redirect API to route alerts to CMSMonitoring Prometheus and then use Prometheus/AM for alerts routing
  
In Go, it would be one page server, static executable (i.e. no run-time dependencies), it will be accessible in CMSWEB, it will be reachable from anywhere, it will work with any language via HTTP protocol, it will support customizable auth/authz as any other services we run, etc. In my view, it is the easiest and most portable and scalable solution.

khurtado:
@amaltaro I think we should first put some effort into fixing the root problem both this issue and the cron issue had in common: getting rid of the host-level passwd/group binding for the production user/group uuids inside docker. I say this because we may encounter other issues in the future with other services because of this binding trick.

For the mail issue, considering this solution is working and uses the host-level mail agent, I would vote for using this now and perhaps open another issue with the tag "new feature", with a feature migration proposal, and assign a priority accordingly. In my opinion, while reimplementing this logic would be nice, if the goal is to make maintenance easier, then getting rid of the binding trick would remove a lot of the hacks we are doing at runtime and make maintenance much easier already though.

todor-ivanov:
hi @khurtado 

> In my opinion, while reimplementing this logic would be nice, if the goal is to make maintenance easier, then getting rid of the binding trick would remove a lot of the hacks we are doing at runtime and make maintenance much easier already though.

I couldn't agree more!

docker/pypi/wmagent/wmagent-docker-run.sh
```diff
@@ -148,3 +172,13 @@ userStatus="$(docker exec -u root -it wmagent sh -c "passwd -S $wmaUser" | awk '
 if [ "${userStatus:0:1}" == "P" ]; then
     docker exec -u root -it wmagent sh -c "echo $wmaUser:$wmaUser | chpasswd"
 fi
+
+# Install s-nail
```
todor-ivanov, line 176: This indeed  would be better to go into the Docker file, either `wmagent` or `wmbase`. Which might also alleviate/remove  the need for the package configuration here.

docker/pypi/wmagent/wmagent-docker-run.sh
```diff
@@ -72,17 +72,40 @@ ln -s $HOST_MOUNT_DIR/srv/wmagent /data/srv/wmagent
 passwdEntry=$(getent passwd $wmaUser | awk -F : -v wmaHome="/home/$wmaUser" '{print $1 ":" $2 ":" $3 ":" $4 ":" $5 ":" wmaHome ":" $7}')
 groupEntry=$(getent group $wmaGroup)
 
+# @TODO: Create needed unix accounts in container, rather than bind mounting files from the host
 # workaround case where Unix account is not in the local system (e.g. sssd)
 [[ -d $HOST_MOUNT_DIR/admin/etc/ ]] || (mkdir -p $HOST_MOUNT_DIR/admin/etc) || exit $?
+
+# Validation step
+# Delete local docker passwd/group files if uucp is not present (so it can be recreated)
+if [ -f "$HOST_MOUNT_DIR/admin/etc/passwd" ]; then
```
todor-ivanov, line 81: Those two nested `if` conditions would have the same effect but would be much more readable if written in the form: 
```
[[ -f "$HOST_MOUNT_DIR/admin/etc/passwd" ]] && ! grep -q uucp "$HOST_MOUNT_DIR/admin/etc/passwd" &&  {
        rm $HOST_MOUNT_DIR/admin/etc/passwd
}
```

docker/pypi/wmagent/wmagent-docker-run.sh
```diff
@@ -72,17 +72,40 @@ ln -s $HOST_MOUNT_DIR/srv/wmagent /data/srv/wmagent
...
+    if ! grep -q uucp "$HOST_MOUNT_DIR/admin/etc/passwd"; then
+        rm $HOST_MOUNT_DIR/admin/etc/passwd
+    fi
+fi
+if [ -f "$HOST_MOUNT_DIR/admin/etc/group" ]; then
```
todor-ivanov, line 86: same as above

todor-ivanov (commented):
Hi @khurtado  I had only two comments inline concerning readability of the code, but not game changing ones. In case we decide to continue in that direction consider my approval of this PR.

khurtado:
@todor-ivanov Thank you! I added implemented your change proposals for the nested conditionals.
@amaltaro If you agree with installing s-nail (500 KB package) in wmagent-base (or wmagent itself, but I would rather do it in the base image, because the wmagent Dockerfile has no package installations, so I would like to keep it consistent), then I can proceed with a new PR for that part.

todor-ivanov (approved):
thanks @khurtado

amaltaro:
@khurtado can you please update this PR (Dockerfile and scripts) with the new image tag available in CERN Registry? Please see: https://github.com/dmwm/CMSKubernetes/pull/1573#issuecomment-2605091527

khurtado:
@amaltaro I just made the changes and tested with 2.3.8rc10 (which has the latest htcondor python bindings, as 10.9.0 is not on pypi anymore and couldn't build the image).
Everything worked fine.

docker/pypi/wmagent/wmagent-docker-run.sh
```diff
@@ -148,3 +169,10 @@ userStatus="$(docker exec -u root -it wmagent sh -c "passwd -S $wmaUser" | awk '
 if [ "${userStatus:0:1}" == "P" ]; then
     docker exec -u root -it wmagent sh -c "echo $wmaUser:$wmaUser | chpasswd"
 fi
+
+# Configure s-nail to use the host s-nail mail server
+docker exec -u root -it wmagent cp /etc/s-nail_host.rc /etc/s-nail.rc
```
amaltaro, line 174: I was thinking that we could keep the modified s-nail.rc under `$HOST_MOUNT_DIR/admin/etc/` and simply mount it from there to `/etc/s-nail.rc`. This way we could do some of this tweaking only once in a lifetime of an agent.

The `update-alternatives` seem to be still required though - but perhaps it could go to the wmagent Dockerfile?

amaltaro (approved):
@khurtado thank you for making the relevant changes and testing them out.

I left one comment along the code, which I think it would provide further improvements to the way we are dealing with this mail service setup. In case we ever need to review this, we might consider that.

docker/pypi/wmagent/wmagent-docker-run.sh
amaltaro: That's a good point! Thank you for this clarification, Kenyi.

amaltaro:
Let me merge this before we forget. This is a great timing as well, as we are about to release a final stable release for WMAgent.

PR #1569, opened 2024-11-27
Fix broken cron jobs redirect

todor-ivanov:
Fixes: https://github.com/dmwm/WMCore/issues/12184

With the current PR we fix the broken order of `stdError` to  `stdOut` redirect in the WMAgent's cronjobs in order to stop the eventual spam which would have been generated, once we fix the e-mailing mechanism in our Docker containers.


todor-ivanov:
this Fix goes hand in hand with: https://github.com/dmwm/WMCore/pull/12185

PR #1570, opened 2024-11-28
Reduce tweaking at deploy time

todor-ivanov:
Fixes: https://github.com/dmwm/WMCore/issues/12182

**Description:**
With the current change we  tartget to reduce the dependence on the old WMA_ENV_FILE ergo its tweaking at deploy time. 

So far we are changing only the env. variables which we were keeping in the WMA_ENV_FILE only for backwords compatibility with the  old rpm based agents and we do not touch any of the variables requiring WMCore code base change. We are  moving the setup of those variables as early as possible in the container environment  configuration.

**Is it backward compatible:**
NO

**External dependencies  && Related PRs:**
This PR must go hand in hand with: https://github.com/dmwm/WMCore/pull/12183


mapellidario (approved):
I agree, it's always better to set variables explicitly, it makes reading and using grep easier, and it improves portability to whatever new set of scripts way easier. thanks :)

PR #1581, opened 2025-03-13
Quick and dirty fix for addressing DBS preprod endpoint discrepancy

todor-ivanov:
With this PR we partially address one of the issues mentioned in this DBS related ticket:  https://github.com/dmwm/dbs2go/issues/119

### DIFF
--- a/kubernetes/cmsweb/scripts/deploy-srv.sh
+++ b/kubernetes/cmsweb/scripts/deploy-srv.sh
@@ -211,13 +211,12 @@ elif  [ "$cmsweb_env" == "k8s-preprod" ] ; then
 elif [ "$srv" == "crabserver" ]; then
             cat $srv.yaml | sed -e "s, #imagetag,$cmsweb_image_tag,g" |  sed -e 's+crabserver/prod+crabserver/preprod+g' | sed -e "s,k8s #k8s#,$cmsweb_env,g"  >  $srv.yaml.new
             cat $srv.yaml | sed -e "s, #imagetag,$cmsweb_image_tag,g" |  sed -e 's+crabserver/prod+crabserver/preprod+g' | sed -e "s,k8s #k8s#,$cmsweb_env,g" | kubectl apply -f -
- 
 elif  [[ "$cmsweb_env" == "k8s-auth" ]] && [[ "$srv" == "dbs2go-global-r" || "$srv" == "dbs2go-global-w" || "$srv" == "dbs2go-phys03-r" || "$srv" == "dbs2go-phys03-w" || "$srv" == "dbs2go-global-m" || "$srv" == "dbs2go-phys03-m" || "$srv" == "dbs2go-global-migration" || "$srv" == "dbs2go-phys03-migration" ]] ; then
         cat $srv.yaml | sed -e "s, #imagetag,$cmsweb_image_tag,g" |  sed -e 's+dbs/prod+dbs/int+g' | sed -e "s,k8s #k8s#,$cmsweb_env,g"  >  $srv.yaml.new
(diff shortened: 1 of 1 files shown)


arooshap (changes requested):
Hi @todor-ivanov, the usage of image tag is crucial to ensure the correct working of the script.

kubernetes/cmsweb/scripts/deploy-srv.sh
```diff
@@ -211,13 +211,12 @@ elif  [ "$cmsweb_env" == "k8s-preprod" ] ; then
...
         cat $srv.yaml | sed -e "s, #imagetag,$cmsweb_image_tag,g" |  sed -e 's+dbs/prod+dbs/int+g' | sed -e "s,k8s #k8s#,$cmsweb_env,g"  >  $srv.yaml.new
         cat $srv.yaml | sed -e "s, #imagetag,$cmsweb_image_tag,g" |  sed -e 's+dbs/prod+dbs/int+g' | sed -e "s,k8s #k8s#,$cmsweb_env,g"  | kubectl apply -f - 
-elif [[ "$srv" == "dbs-global-r"  || "$srv" == "dbs-global-w"  ||  "$srv" == "dbs-migrate"  ||  "$srv" == "dbs-phys03-r" || "$srv" == "dbs-phys03-w" || "$srv" == "dbs2go-global-r" || "$srv" == "dbs2go-global-w" || "$srv" == "dbs2go-phys03-r" || "$srv" == "dbs2go-phys03-w" || "$srv" == "dbs2go-global-m" || "$srv" == "dbs2go-phys03-m" || "$srv" == "dbs2go-global-migration" || "$srv" == "dbs2go-phys03-migration" ]] ; then
-        cat $srv.yaml | sed -e "s, #imagetag,$cmsweb_image_tag,g" |  sed -e 's+dbs/prod+dbs/dev+g' | sed -e "s,k8s #k8s#,$cmsweb_env,g"  >  $srv.yaml.new
```
vkuznet, line 219: Todor, I second Aroosha, you should not break existing behavior. Please restore imagetag  here and later add additional change for dbs instance change.

kubernetes/cmsweb/scripts/deploy-srv.sh
```diff
@@ -211,13 +211,12 @@ elif  [ "$cmsweb_env" == "k8s-preprod" ] ; then
...
-        cat $srv.yaml | sed -e "s, #imagetag,$cmsweb_image_tag,g" |  sed -e 's+dbs/prod+dbs/dev+g' | sed -e "s,k8s #k8s#,$cmsweb_env,g"  | kubectl apply -f -
+elif [[ "$cmsweb_env" == "k8s-preprod" ]] && [[ "$srv" =~ dbs2go.* ]]; then
+        cat $srv.yaml | sed -e "s|/dbs/prod/|/dbs/int/|g"  >  $srv.yaml.new
```
vkuznet, line 218: what is a purpose of newly created `$srv.yaml.new` here if it is not used afterwards?

vkuznet (changes requested):

kubernetes/cmsweb/scripts/deploy-srv.sh
todor-ivanov: Now, this is a question I would also like to have an answer to. I just repeated the actions taken for all other custom cases addressed in the lines above, but never found where would that file be used in reality. I am in favor of completely dropping this line.

kubernetes/cmsweb/scripts/deploy-srv.sh
todor-ivanov: BTW, I just overlooked the `$cmseb_image_tag`  tuneup here. The removal of the above lines was simply because they relate to the old python based services, which we never plan to use any more. 
I'll take care of the missing tag.

PR #1599, opened 2025-04-23
Adjust image names with respect to WM pep-625 changes

vkuznet:
This PR includes changes to image names with respect to WM pep-625 changes, see https://github.com/dmwm/WMCore/issues/12181


vkuznet:
Based on MM exchange Alan requested to proceed with merging.

PR #1600, opened 2025-04-24
Set 2 replicas instead of 5 for WM testbed services

amaltaro:
Any WM service that is currently (24/Apr/2025) running with 5 replicas, will start running with 2 replicas - only for cmsweb-testbed environment (defined in the script as `k8s-preprod`).


amaltaro:
I've done a quick test and it looks good to me:
```
amaltaro@lxplus917:~/private/CMSKubernetes/kubernetes/cmsweb $ cat services/reqmgr2.yaml | sed -e "s,1 #PROD# 5,2,g" | sed -e "s,#PROD#,      ,g" |  sed -e "s,logs-cephfs-claim,$cmsweb_log,g" | sed -e "s, #imagetag,$cmsweb_image_tag,g" | sed -e "s,k8s #k8s#,$cmsweb_env,g" | grep replicas
  replicas: 2
  ```
  
  @arooshap could you please review it?

arooshap (approved):

todor-ivanov:
@amaltaro, @arooshap   this PR actually broke DAS deployment. I think we need to think of a better way to identify the number of replicas in `prod` and `preprod` per service. It seems to me those parsing /editing stanzas here should be better identifiable  per service and not applied on common ground.

amaltaro:
Sorry for introducing this problem for DAS, @todor-ivanov . I created a more restrict fix for this - while rolling back the previous `else` behavior: https://github.com/dmwm/CMSKubernetes/pull/1606

PR #1607, opened 2025-05-13
Update cron-proxy.yaml

arooshap:


mapellidario (approved):
looks great, thank you aroosha :)

PR #1611, opened 2025-05-14
wmcore - switch to py3.12 for pypi docker images

mapellidario:
### status

tested, if we have consensus on the format of the new tags, it can be merged

warning: do not create a new tag in dmwm/WMCore yet, we (i) need to build and push the new tags for dmwm-base and wmagent-base first

### description

move wmcore images from python:3.8 based on debian 11 (`bullseye`) to python:3.12 based on debian 12 (`bookworm`).

moving from a "full" debian image to a slim one requires adding a few packages manually:
    - `git`, `build-essential` (gcc), `unzip`

~~moreover, i specified the latest available tag for `cmsweb/cmsweb-base` and `cmsweb/exportes`, as i do not like to rely on `latest`.~~

Moreover:

- using the most recent cmsweb-alma9-base image
- using the most recent exporter image

### DIFF
--- a/docker/pypi/dmwm-base/Dockerfile
+++ b/docker/pypi/dmwm-base/Dockerfile
@@ -1,14 +1,17 @@
-FROM registry.cern.ch/cmsweb/cmsweb-base as cmsweb-base
-FROM registry.cern.ch/cmsweb/exporters as exporters
-FROM python:3.8-bullseye
+FROM registry.cern.ch/cmsweb/exporters:20250519-stable as exporters
+FROM python:3.12-slim-bookworm
 MAINTAINER Valentin Kuznetsov vkuznet@gmail.com
-RUN apt-get update && \
-    apt-get install -y curl vim libcurl4 libcurl4-openssl-dev python3-pycurl pip apache2-utils sudo less && \
-    apt-get clean
+
+# see https://docs.docker.com/build/building/best-practices/#apt-get
+RUN apt-get update && apt-get install -y --no-install-recommends \
+    curl libcurl4 libcurl4-openssl-dev python3-pycurl \ 
+    apache2-utils \
+    sudo less vim git \
+    && apt-get clean \
+    && rm -rf /var/lib/apt/lists/*
+
 RUN ln -s /usr/bin/python3 /usr/bin/python
 RUN mkdir /etc/grid-security
-COPY --from=cmsweb-base /etc/grid-security/certificates /etc/grid-security/certificates
-COPY --from=cmsweb-base /etc/grid-security/vomsdir /etc/grid-security/vomsdir
 COPY --from=exporters /data/cmsweb-ping /usr/bin/cmsweb-ping
 COPY --from=exporters /data/process_exporter /usr/bin/process_exporter
 COPY --from=exporters /data/cpy_exporter /usr/bin/cpy_exporter
--- a/docker/pypi/wmagent-base/Dockerfile
+++ b/docker/pypi/wmagent-base/Dockerfile
@@ -1,18 +1,23 @@
 FROM registry.cern.ch/cmsweb/oracle:21_5-stable as oracle
-FROM registry.cern.ch/cmsweb/dmwm-base:pypi-20240923-stable
+FROM registry.cern.ch/cmsweb/dmwm-base:pypi-20250523-stable
 
-# Install basic OS package dependencies
-RUN apt-get update
-RUN apt-get install -y libmariadb-dev-compat libmariadb-dev apache2-utils cron mariadb-server myproxy voms-clients voms-clients-java rlwrap libaio1 && apt-get clean
-
-# Install some debugging tools
-RUN apt-get install -y hostname net-tools iputils-ping procps jq && apt-get clean
-
-# Install recursive ps utility tool
-RUN apt-get install -y pslist && apt-get clean
-
-# Install s-nail package
-RUN apt-get install -y s-nail && apt-get clean
+# see https://docs.docker.com/build/building/best-practices/#apt-get
+RUN apt-get update && apt-get install -y --no-install-recommends \
+    # basic OS packages
+    cron \
+    mariadb-server libmariadb-dev-compat libmariadb-dev \
+    myproxy voms-clients voms-clients-java \
+    rlwrap libaio1 \
+    # required for cx_oracle
+    build-essential \
+    # required for yui 
+    wget unzip \
+    # debugging 
+    hostname net-tools iputils-ping procps jq \
+    # Install recursive ps utility tool
+    pslist \
+    s-nail \
+    && rm -rf /var/lib/apt/lists/*
 
 # copy oracle client:
 COPY --from=oracle /usr/lib/oracle /usr/lib/oracle
--- a/docker/pypi/msmonitor/Dockerfile
+++ b/docker/pypi/msmonitor/Dockerfile
@@ -1,4 +1,4 @@
-FROM registry.cern.ch/cmsweb/dmwm-base:pypi-20240923-stable
+FROM registry.cern.ch/cmsweb/dmwm-base:pypi-20250523-stable
 MAINTAINER Valentin Kuznetsov vkuznet@gmail.com
 # TAG to be passed at build time through `--build-arg TAG=<PYPI_TAG>`. Default: None
 ARG TAG=None
--- a/docker/pypi/msoutput/Dockerfile
+++ b/docker/pypi/msoutput/Dockerfile
@@ -1,4 +1,4 @@
-FROM registry.cern.ch/cmsweb/dmwm-base:pypi-20240923-stable
+FROM registry.cern.ch/cmsweb/dmwm-base:pypi-20250523-stable
 MAINTAINER Valentin Kuznetsov vkuznet@gmail.com
 # TAG to be passed at build time through `--build-arg TAG=<PYPI_TAG>`. Default: None
 ARG TAG=None
--- a/docker/pypi/mspileup/Dockerfile
+++ b/docker/pypi/mspileup/Dockerfile
@@ -1,4 +1,4 @@
-FROM registry.cern.ch/cmsweb/dmwm-base:pypi-20240923-stable
+FROM registry.cern.ch/cmsweb/dmwm-base:pypi-20250523-stable
 MAINTAINER Valentin Kuznetsov vkuznet@gmail.com
 # TAG to be passed at build time through `--build-arg TAG=<PYPI_TAG>`. Default: None
 ARG TAG=None
--- a/docker/pypi/msrulecleaner/Dockerfile
+++ b/docker/pypi/msrulecleaner/Dockerfile
@@ -1,4 +1,4 @@
-FROM registry.cern.ch/cmsweb/dmwm-base:pypi-20240923-stable
+FROM registry.cern.ch/cmsweb/dmwm-base:pypi-20250523-stable
 MAINTAINER Valentin Kuznetsov vkuznet@gmail.com
 # TAG to be passed at build time through `--build-arg TAG=<PYPI_TAG>`. Default: None
 ARG TAG=None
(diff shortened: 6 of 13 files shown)


docker/pypi/dmwm-base/Dockerfile
```diff
@@ -1,6 +1,6 @@
-FROM registry.cern.ch/cmsweb/cmsweb-base as cmsweb-base
-FROM registry.cern.ch/cmsweb/exporters as exporters
-FROM python:3.8-bullseye
+FROM registry.cern.ch/cmsweb/cmsweb-base:20230901-stable as cmsweb-base
```
vkuznet, line 1: The timestamp for stable images is 2 years old, if I recall correctly the stable images retention policy is 1 year, maybe it was extended. I rather prefer to make extra effort and create new stable images with 202506xx-stable tag such that we'll perform major migration ensuring that stable images will not disappear suddenly.

docker/pypi/dmwm-base/Dockerfile
```diff
@@ -1,6 +1,6 @@
...
+FROM registry.cern.ch/cmsweb/exporters:20231002-stable as exporters
```
vkuznet, line 2: same here, we should rebuild stable image for exporters and use the newest one, i.e. 202505xx-stable

vkuznet (changes requested):

docker/pypi/dmwm-base/Dockerfile
mapellidario: ah, maybe i should look at the `cmsweb/cmsweb-alma9-base` image, give me a sec

amaltaro:
@mapellidario I understand you have already uploaded an image for `dmwm-base`, according to the tag you provided in this PR. Did you also upload it for the `wmagent-base` image? I don't think we update this image in our CD pipeline, so please consider that as well.

docker/pypi/dmwm-base/Dockerfile
mapellidario: the latest tag for `cmsweb/cmsweb-alma9-base` is `20250506` without any `-stable` suffix. for the time being i would stick with this tag

vkuznet: Dario, I rather to not use non stable image since their retention policy is quite short, if I recall correctly it is about 3 months. Instead, just do the following:
```
docker pull registry.cern.ch/cmsweb/cmsweb-alma9-base:20250506
docker tag registry.cern.ch/cmsweb/cmsweb-alma9-base:20250506 registry.cern.ch/cmsweb/cmsweb-alma9-base:20250506-stable
docker push registry.cern.ch/cmsweb/cmsweb-alma9-base:20250506-stable
```
and then use stable image. This will create an alias tag and push it to registry where new image will be hold for longer period of time.

mapellidario: I know i know, i just do not have permissions to push to `cmsweb-alma9-base`

mapellidario: but ok, assuming you or aroosha will do it, i changes the tag to use the `-stable`

vkuznet: done:
```
docker push registry.cern.ch/cmsweb/cmsweb-alma9-base:20250506-stable
The push refers to repository [registry.cern.ch/cmsweb/cmsweb-alma9-base]
bb19aa74b58c: Layer already exists
b92fcbc8bdec: Layer already exists
1e4ed39e79fd: Layer already exists
84bb803a6a3c: Layer already exists
ec623c3d76de: Layer already exists
4480616e9270: Layer already exists
20250506-stable: digest: sha256:6f910ff0a84960fba4f58f4df5e60fa88658ea95f7bb2f362a864ec70e89625f size: 1582
```

mapellidario: thanks, then i mark this conversation as resolved

docker/pypi/dmwm-base/Dockerfile
mapellidario: i opened a tentative PR https://github.com/dmwm/CMSKubernetes/pull/1613

mapellidario:
@amaltaro  nope i have not pushed any `-stable` tag yet for dmwm-base nor wmagent-base.

docker/pypi/dmwm-base/Dockerfile
mapellidario: meged

docker/pypi/dmwm-base/Dockerfile
```diff
@@ -1,6 +1,6 @@
-FROM registry.cern.ch/cmsweb/cmsweb-base as cmsweb-base
-FROM registry.cern.ch/cmsweb/exporters as exporters
-FROM python:3.8-bullseye
+FROM registry.cern.ch/cmsweb/cmsweb-alma9-base:20250506-stable as cmsweb-base
```
vkuznet, line 1: Dario, this image is only used to copy grid certificates from the image to new one. But as far as I know the CMSWEB managed to build k8s cluster where certificates are supplied by core node. If this is the case, we don't need `cmsweb-base` image (and copies of certs from it) at all. I would suggest to double check this with @arooshap and decide if we need to manage certificates on our own or if they are provided by k8s nodes we can just mount them within deployment.

docker/pypi/dmwm-base/Dockerfile
```diff
@@ -1,6 +1,6 @@
...
+FROM registry.cern.ch/cmsweb/exporters:20250519-stable as exporters
+FROM python:3.12-bookworm
```
vkuznet, line 3: Dario, I wonder why did you choose `3.12-bookworm` image instead of smaller `python:3.12-slim` which is stripped-down version of the standard image, removing unnecessary files and tools to reduce size? Since we install all our tools via pip I don't see any more point in having full image and we most likely can use stripped one. Sorry for pointing this later, but it would be useful I think to test and reduce image size.

vkuznet (changes requested):

docker/pypi/dmwm-base/Dockerfile
mapellidario: Apparently yeah we should be mounting them from the host [1] and use them to override what is inside the image [2]. I will have a look

[1] https://github.com/dmwm/CMSKubernetes/blob/9185d15c0c6f8a0408d243bd7fbac8a52212498a/kubernetes/cmsweb/services/reqmgr2.yaml#L138
[2] https://github.com/dmwm/CMSKubernetes/blob/9185d15c0c6f8a0408d243bd7fbac8a52212498a/kubernetes/cmsweb/services/reqmgr2.yaml#L52

docker/pypi/dmwm-base/Dockerfile
mapellidario: done

docker/pypi/dmwm-base/Dockerfile
mapellidario: done

mapellidario:
so, i implemented latest valentin suggestions:

- i removed the certificates from the cmsweb-alma9-base image, i confirmed with aroosha and the machinery to use the certificates from the hosts is already in place
- i switched to the slim version. this required adding a few packages that were used down the line. I took the liberty of organizing a bit better how we call apt get

again, i am open to suggestions

amaltaro:
@mapellidario do you want to run some extra tests with these new changes (if not yet executed)? 
If you are happy with the tests already performed, once Valentin checks the outcome of your changes, I think we can move forward.

docker/pypi/dmwm-base/Dockerfile
```diff
@@ -1,14 +1,18 @@
...
+FROM registry.cern.ch/cmsweb/exporters:20250519-stable as exporters
+FROM python:3.12-slim-bookworm
 MAINTAINER Valentin Kuznetsov vkuznet@gmail.com
-RUN apt-get update && \
-    apt-get install -y curl vim libcurl4 libcurl4-openssl-dev python3-pycurl pip apache2-utils sudo less && \
-    apt-get clean
+
+# see https://docs.docker.com/build/building/best-practices/#apt-get
+RUN apt-get update && apt-get install -y --no-install-recommends \
+    curl wget libcurl4 libcurl4-openssl-dev python3-pycurl \
```
vkuznet, line 7: - Why do we need wget if we have curl, either tool is sufficient to make HTTP requests.
- Why do we need git, this implicitly invites people to log into docker image and to development which should not be promoted. I rather prefer to put effort to perform development on local laptop using plain python. That said, if we don't have it you may have by simply login to image and perform `apt-get update && apt-get install git`. Let developers to install tools within run-time environment which can be used to test/debug stuff but should not be part of image itself. Once debugging is done image can be restarted to pristine state without dev tools.
- vim is improved vi, and vi is present by default, I don't see any need for vim if we can use vi.

docker/pypi/wmagent-base/Dockerfile
```diff
@@ -1,18 +1,23 @@
...
-# Install recursive ps utility tool
-RUN apt-get install -y pslist && apt-get clean
-
-# Install s-nail package
-RUN apt-get install -y s-nail && apt-get clean
+# see https://docs.docker.com/build/building/best-practices/#apt-get
+RUN apt-get update && apt-get install -y --no-install-recommends \
+    # basic OS packages
+    cron \
+    mariadb-server libmariadb-dev-compat libmariadb-dev \
```
vkuznet, line 8: I'm not sure why do we need mariadb server in this image, it it run outside of wmagent in a separate container?

docker/pypi/wmagent-base/Dockerfile
```diff
@@ -1,18 +1,23 @@
...
+    myproxy voms-clients voms-clients-java \
+    rlwrap libaio1 \
+    # required for cx_oracle
+    build-essential \
+    # required for yui 
+    unzip \
+    # debugging 
+    hostname net-tools iputils-ping procps jq \
```
vkuznet, line 16: I'm not in favor of putting debugging tools within an image. If you need them you can login to image and install them. Permanently having them only blow up dependencies and size of the image and increase vulnerabilities. That said, I suggest to bring this up to the team and make a decision about that.

vkuznet (changes requested):

mapellidario:
@vkuznet I see your point and I agree with almost every statement you make. However, I think that they are out of the scope of the original issue that we are trying to address with this PR [1]. I am eager to start the validation of the python3.12 images, the longer we wait the later we find problems.

I suggest that we merge the current changes and come back to your suggestions in [2].

---

[1] https://github.com/dmwm/WMCore/issues/12208
[2] https://github.com/dmwm/WMCore/issues/12374

docker/pypi/dmwm-base/Dockerfile
mapellidario: I do not find vi in the default base image

```
> podman run -it --rm python:3.12-slim-bookworm bash
root@ff6a9ce1ff0e:/# vi
bash: vi: command not found
```

vkuznet: then it's fine to keep vim, I didn't look explicitly nowadays what is supplied in slim image, maybe vi is no longer default editor in Debian :), or maybe slim does not need vi per se as it is very strict and small image. Either way we can install vim or do not handle it at all and leave it up to developers to install during runtime of debugging process.

mapellidario: wget was used by a script to install yui and was originally provided by the python:3.8 image. i changed the script to use curl and removed wget

mapellidario: regarding `git`, i understand your point, i do not care too much. i am fine either way

docker/pypi/wmagent-base/Dockerfile
mapellidario: i have no clue, i do not want to investigate now. it was there before, maybe we can remove it

docker/pypi/dmwm-base/Dockerfile
vkuznet: since you created separate issue for clean-up images https://github.com/dmwm/WMCore/issues/12374 then I'm ok with leaving this for time being.

docker/pypi/wmagent-base/Dockerfile
vkuznet: this better ask @amaltaro as MariaDB is quite a lot of dependencies we may not need at all.

amaltaro (approved):
Thank you very much for providing these changes and for the discussion. It all looks good to me as well.

Given that the base images are already available, I will also proceed with a new WMCore tag.

PR #1612, opened 2025-05-14
Update das-mongo-exporter.yaml

arooshap:
Update busybox to the stable version(1.36.1), and apply a patch for the newer nslookup command.

### DIFF
--- a/kubernetes/cmsweb/services/das-mongo-exporter.yaml
+++ b/kubernetes/cmsweb/services/das-mongo-exporter.yaml
@@ -60,5 +60,5 @@ spec:
           name: das-mongo-p
       initContainers:
       - name: checkcouchdb
-        image: busybox:1.28
-        command: ['sh', '-c', 'until nslookup das-mongo.das; do echo "Waiting for das-mongo"; sleep 10; done;']
+        image: busybox:1.36.1 #the stable tag
+        command: ['sh', '-c', 'until nslookup das-mongo.das.svc.cluster.local; do echo "Waiting for das-mongo"; sleep 10; done;']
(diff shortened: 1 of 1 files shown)


kubernetes/cmsweb/services/das-mongo-exporter.yaml
```diff
@@ -60,5 +60,5 @@ spec:
           name: das-mongo-p
       initContainers:
       - name: checkcouchdb
```
todor-ivanov, line 62: change this to `checkmongodb`

kubernetes/cmsweb/services/das-mongo-exporter.yaml
```diff
@@ -60,5 +60,5 @@ spec:
...
-        image: busybox:1.28
-        command: ['sh', '-c', 'until nslookup das-mongo.das; do echo "Waiting for das-mongo"; sleep 10; done;']
+        image: busybox:1.36.1 #the stable tag
```
todor-ivanov, line 63: if you have  tested with the latest stable tag like: 
`busybox:stable` please change it to this one instead of fixed number

todor-ivanov (commented):
hi @arooshap, please take a look at my two comments inline

kubernetes/cmsweb/services/das-mongo-exporter.yaml
arooshap: the stable tag is not resolving. That's why I used this. I can upload this image to registry.

todor-ivanov: ok

PR #1613, opened 2025-05-14
cmsweb/exporters - build a new version

mapellidario:
we noticed that the `latest` tag of cmsweb/exporters is a few years old. we should build a new one.

changes

- new version of golang compiler
- new version of debian
- new version of node_exporter
- new version of apache_exporter
- removed scripts that are not used [1]

suggestion

- maybe we need to change the provider of mongodb_exporter since the one that we are using did not receive any update since 2016. however, checking that alternatives are compatible with what we are running at the moment is out of scope for this PR.

fyi: @amaltaro 

---

[1]

The following scripts have been removed from all the dockerfiles in `docker/exporters/` and are not imported by any other docker image (or so it seems from a few greps)

```plaintext
docker/exporters/install.sh
docker/exporters/probe.sh
docker/exporters/run.sh
```


docker/exporters/Dockerfile
```diff
@@ -25,16 +29,21 @@ RUN cd cmsweb-exporters && \
...
-
-RUN wget https://github.com/Lusitaniae/apache_exporter/releases/download/v0.9.0/apache_exporter-0.9.0.linux-amd64.tar.gz && tar -xvzf apache_exporter-0.9.0.linux-amd64.tar.gz --directory /data/gopath && cp /data/gopath/apache_exporter-0.9.0.linux-amd64/apache_exporter /build
+#        https://github.com/Lusitaniae/apache_exporter/releases/download/v1.0.10/apache_exporter-1.0.10.linux-amd64.tar.gz
+RUN wget https://github.com/Lusitaniae/apache_exporter/releases/download/v${APACHEEXP_VER}/apache_exporter-${APACHEEXP_VER}.linux-amd64.tar.gz && \
+  tar -xvzf apache_exporter-${APACHEEXP_VER}.linux-amd64.tar.gz --directory /data/gopath && \
+  cp /data/gopath/apache_exporter-${APACHEEXP_VER}.linux-amd64/apache_exporter /build
 
 # build mongo exporter
+# TODO: consider using an actively developed exporter, such as
+#       https://github.com/percona/mongodb_exporter
```
vkuznet, line 39: For this I suggest you open CMSMONIT issue and ask Monitoring group to install and test it with existing dashboard. If it is compatible then we can easily switch otherwise we need a new dashboard which will use this exporter metrics.

docker/exporters/Dockerfile
```diff
@@ -1,13 +1,17 @@
-FROM golang:latest as go-builder
+FROM golang:1.24 as go-builder
```
vkuznet, line 1: latest always guarantee latest go compiler image. If your intend to use latest compiler I will leave it as is, otherwise it is ok to stick to concrete version. The beauty of Go is backward compatibility. All its newer compiler guarantee that old code will compile with it.

vkuznet (changes requested):

docker/exporters/Dockerfile
mapellidario: > The beauty of Go is backward compatibility

okok, you convinced me. switching back to `latest`

docker/exporters/Dockerfile
mapellidario: i see, but since we are not using mongodb_expoerter in our wmcore images, can we merge this PR, build a new exporter image and tag it with `-stable`, use it for https://github.com/dmwm/CMSKubernetes/pull/1611 , then build a new tag with the new `percona/mongodb_exporter` after cmsmonit team finish their investigations?

vkuznet: as I suggested please open up new CMSMON ticket and link it here for cross references, then we can proceed with merge and stable images.

mapellidario: I opened the ticket [1] and added it to the comments in the dockerfile. 

> if you claim we (WM) group not using particular exporter it does not mean it is not in use.

I know, and this is why i proposed a change instead of just removing removing or changing it :)


[1] https://its.cern.ch/jira/projects/CMSMONIT/issues/CMSMONIT-650?filter=allissues

vkuznet:
@mapellidario , one additional comment I want to clarify with you. You deleted few scripts and I want you to double check that they are no longer in use anywhere. Please add this to the description of the issue that we can refer to this change if necessary. These scripts I used/developed while we port CMSWEB services from RPM based deployment to k8s but it would be useful to grep and see if they are no longer in use. Thanks.

vkuznet:
thank you Dario for cross-checking, I'm ok with deleting those scripts I just want to have a references to easily find them if we'll need them to recover something. That's why I suggested to include explicitly what we delete in a description of this PR.

mapellidario:
I updated the PR description

vkuznet (approved):
thank you Dario, changes look good to me.

PR #1618, opened 2025-05-26
argocd - deploy dmwm services to test11 usign helm templates

mapellidario:
I started experimenting with argocd. Tested and working on my test11 cluster :)

I tried using test11-specific manifest files, but i found using helm templates easier as required less configuration changes.

I had to make small changes to the helm templates, likely changes to the manifest files that have not been backported to helm templates. Then, i created all the `helm/**/values-test11.yml` files.

I suggest that another volunteer from the wmcore team requests the installation of argocd in their test cluster and start experimenting with it according to the instructions I added at [1]. then we can spread the word across our team and consider adoption in testbed. 

fyi: @anpicci @arooshap @vkuznet 

---

[1] https://cms-wmcore.docs.cern.ch/deployment/WMCore-in-kubernetes/#deploy-wmcore-services-with-argocd


arooshap (approved):
Wow, I really appreciate you looking into it!

PR #1604, opened 2025-05-08
cmsweb-alma9-base, dmwm-base alma9 with py3.12, msunmerged alma9

d-ylee:
fixes [#11922](https://github.com/dmwm/WMCore/issues/11922)
extends https://github.com/dmwm/CMSKubernetes/pull/1452

## Images Added
* cmsweb-alma9-base
* dmwm-base based on cmsweb-alma9-base and including python 3.12
* msunmerged based on dmwm-base alma9

PR #1628, opened 2025-07-14
wmcore-dev: Updates image to use Python 3.12

d-ylee:
Part of fixing https://github.com/dmwm/WMCore/issues/12383

## Changes
1. Adds  wmcore-development image (for user development)
2. Uses Python 3.12 from uv
3. Fixes permissions on `/home/cmsbld`
4. Update Makefile
5. Uses ARGs for the image to be more easily automated

## Depends on
https://github.com/dmwm/CMSKubernetes/pull/1604
Depends on the cmsweb-alma9-base and gfal2 images in this PR.


d-ylee:
This image doesn't use the cmsweb-alma9-base image anymore, so PR #1631 can be disregarded.

docker/pypi/wmcore-dev/Dockerfile
```diff
@@ -1,11 +1,27 @@
-FROM --platform=linux/amd64 registry.cern.ch/cmsweb/cmsweb-alma9-base:20240530
+ARG PYTHON_VERSION=3.12
+ARG IMAGE_VERSION=0.4.0
 
-LABEL version="0.2.0"
+FROM --platform=linux/amd64 registry.cern.ch/cmsweb/gfal:alma9-py${PYTHON_VERSION}-gfal-1.13.0 as gfal2
+FROM --platform=linux/amd64 registry.cern.ch/cmsweb/dmwm-base:alma9-py3.12-20250728
```
amaltaro, line 5: I would suggest to use stable images for these 2 dependencies.

amaltaro (changes requested):
@d-ylee this looks good in general, however I have several questions/comments to make, which are likely going to trigger updates to this PR.
1. Can you please update the PR description clarifying why all of the python scripts have been removed from this build?
2. what is the difference between Dockerfile and Dockerfile.dev? If we need both, can you please clarify them in the README file?
3. I am not sure I understand the usage of the `IMAGE_VERSION` argument. Is it supposed to be the same as the label used in CERN Registry? Perhaps add a comment line above that ARG statement?
4. I think we could also add instructions to the README on how to use the Makefile

PR #1629, opened 2025-07-15
python-gfal2 Alma9 Py3.12 image

d-ylee:
Docker configuration for a gfal2 alma9 image.  This builds gfal2 against Alma9 and Python 3.12.

Depends on:
https://github.com/dmwm/CMSKubernetes/pull/1630

Required for https://github.com/dmwm/CMSKubernetes/pull/1628

PR #1630, opened 2025-07-15
dmwm-alma9-base

d-ylee:
Docker configuration for a dmwm-alma9-base image. One for Python 3.9, the other for Python 3.12.

Required by:
https://github.com/dmwm/CMSKubernetes/pull/1629
https://github.com/dmwm/CMSKubernetes/pull/1628


amaltaro:
@d-ylee unless there is a reason that I am missing, I would suggest to convert the py312 as the default Dockerfile image (and just drop the py3.9) one. What do you think?

PR #1641, opened 2025-08-20
Upgrade CouchDB and prometheus export to newer versions; adjust aux files

vkuznet:
This PR upgrades:
- CouchDB to version 3.4.1
- Prometheus CouchDB exporter to version 30.14.0
- setup couchapp into virtual environment as Debian distributions prohibit installation of 3rd party python packages in system area
- adjust manage script to use couchdapp from virtual environment

### DIFF
--- a/docker/pypi/wmagent-couchdb/Dockerfile
+++ b/docker/pypi/wmagent-couchdb/Dockerfile
@@ -6,7 +6,7 @@ ARG TAG
 ENV TAG=${TAG}
 RUN echo TAG=$TAG
 
-RUN apt-get update && apt-get install -y vim less sudo wget unzip python pip
+RUN apt-get update && apt-get install -y vim less sudo wget unzip python3 pip
 
 # # Install some debugging tools
 RUN apt-get install -y hostname net-tools iputils-ping procps emacs-nox tcpdump && apt-get clean
@@ -14,7 +14,17 @@ RUN apt-get install -y hostname net-tools iputils-ping procps emacs-nox tcpdump 
 # Install recursive ps utility tool
 RUN apt-get install -y pslist && apt-get clean
 
-RUN pip install CMSCouchapp
+# RUN pip install CMSCouchapp
+# when we install CMSCouchapp on Debian system it fails with PEP 668 error
+# which prevents installing external (Python packages from pip) on Debian system
+# there are two ways to resolve it:
+# - option A: install with (potential) break of system packages
+#   RUN pip install --break-system-packages CMSCouchapp
+# - option B: install in virtual environment
+RUN apt install -y python3.11-venv
+RUN python3 -m venv /data/venv
+ENV PATH="/data/venv/bin:$PATH"
+RUN pip install --no-cache-dir CMSCouchapp
 
 ENV COUCH_PORT=5984
 ENV COUCH_ROOT_DIR=/data


arooshap (approved):
Looks good to me!

docker/couchdb/Dockerfile
```diff
@@ -1,8 +1,20 @@
...
+RUN apt-get update && apt-get install -y vim less sudo wget unzip python3 pip
+# NOTE:
+# RUN pip install CMSCouchapp
+# when we install CMSCouchapp on Debian system it fails with PEP 668 error
+# which prevents installing external (Python packages from pip) on Debian system
+# there are two ways to resolve it:
+# - option A: install with (potential) break of system packages
+#   RUN pip install --break-system-packages CMSCouchapp
+# - option B: install in virtual environment
+RUN apt install -y python3.11-venv
```
amaltaro, line 14: Why python3.11? Is it the base python version that gets installed with `apt-get`?

BTW, do we still have to install python? Or does it come with the CouchDB vanilla image?

docker/couchdb/Dockerfile
```diff
@@ -24,9 +36,10 @@ RUN mkdir -p $WDIR
 WORKDIR ${WDIR}
 
 # get binary build of couchdb exporter
-RUN curl -ksLO https://github.com/gesellix/couchdb-prometheus-exporter/releases/download/v27.0.0/couchdb-prometheus-exporter_27.0.0_Linux_x86_64.tar.gz
-RUN tar xfz couchdb-prometheus-exporter_27.0.0_Linux_x86_64.tar.gz
-RUN rm couchdb-prometheus-exporter_27.0.0_Linux_x86_64.tar.gz LICENSE README*.md
+ARG VER=30.14.0
```
amaltaro, line 39: I would suggest moving this argument variable to the top of the file and rename it to a more meaningful name (EXPORTER_VER ?).

docker/couchdb/monitor.sh
```diff
@@ -2,13 +2,13 @@
 
 # start couchdb exporter
 COUCH_CONFIG=/data/srv/current/auth/couchdb/couchdb_config.ini
-export COUCHDB_NODE_NAME="couchdb@$(hostname)"
```
amaltaro, line 5: Why are you removing this line? Do you have to rebase this branch?

docker/couchdb/monitor.sh
```diff
@@ -2,13 +2,13 @@
...
 # test if file is not zero size
 if [ -s "${COUCH_CONFIG}" ]; then
-    sudo cp /etc/secrets/$fname /data/srv/current/auth/$srv/$fname
-    sudo chown $USER.$USER /data/srv/current/auth/$srv/$fname
-    nohup couchdb-prometheus-exporter -telemetry.address=":9984" -logtostderr=true \
-      --config=$COUCH_CONFIG -databases.views=false --databases=_all_dbs 2>&1 1>& couchdb_exporter.log < /dev/null &
+    #sudo cp /etc/secrets/$fname /data/srv/current/auth/$srv/$fname
+    #sudo chown $USER.$USER /data/srv/current/auth/$srv/$fname
```
amaltaro, line 8: Why secrets are removed? I think it is needed to be able to query CouchDB and collect the metrics.

amaltaro (changes requested):
@vkuznet I left a few comments along the code.

In addition, let us please not merge this until we can verify that CouchDB works well - AND that manage script is still functional, explicitly:
* we can run couchapp to deploy WMCore code
* we can run status/stop/start/restart (for this, I think a restart/stop will kill the container, but that is another problem...)
Thanks!

PR #1649, opened 2025-09-04
remove obsoletes. Add s3_bucket

belforte:
try the change on test clusters first. I.e. only change `test/config.py`


belforte:
@arooshap please merge. I tested already in cmsweb-test2

PR #1663, opened 2025-10-01
Add Oracle libraries to wmcore-dev images && Ramp up WMCore tag to 2.4.4.rc5

todor-ivanov:
With the current PR we intend to fix the missing Oracle libraries which are needed during our  Jenkins runs. The change is driven by the needs which arose during the upgrade of the oracle python driver. The new  package `python-oracledb` brings an extra OS level dependency from the oracle client libraries.


todor-ivanov:
@vkuznet This docker image is not Debian. It is Alma9

PR #1685, opened 2026-04-15
Revert rucio_con_mon server to old image

mrguilima:
Reverting to old image is necessary, because all recently built images were missing an important functionality (Javascript-based charting).


arooshap (approved):

arooshap:
Hi,

It is done.

PR #1686, opened 2026-04-21
Add path for global DBSMigration service in Ingress

arooshap:
It never existed in our old and new Ingress, but it should be added. I was not communicated to that this service is effective in production.

FYI @uhomsuwa


arooshap:
It is closed by design, confirmed by @vkuznet.

PR #1704, opened 2026-07-15
Add the capabilities for builds from local tree for dbs2go

todor-ivanov:
This change is supposed to add support for local-tree build mode introduced as part of dbs2go issue: https://github.com/dmwm/dbs2go/issues/153 and  required for the corresponding solution: https://github.com/dmwm/dbs2go/pull/158. The functionality is implemented through a separate `Dockerfile.localtree`, while preserving the existing tagged-release `Dockerfile` flow. The only difference between the two is that for the local tree builds no upstream code fetching and build is happening during docker build time, instead the flow expects the dbs2go binary to have been pre-buit in advance, and it is just copied into the image.


vkuznet (changes requested):
Todor, I think this PR requires few changes:
- you should put your name as MAINTAIER
- Remove all commented lines as they are not required here
- The biggest problem here is compatibility of local dbs2go with underlying OS you put in place. Since dbs2go is dynamic executable (due to shared ORACLE libraries) there is a potential incompatibility between local build and `trixie-slim` OS choose. It comes at different levels
  - incompatibility due to GLIBC across all Linux OSes
  - incompatibility of ORACLE libs with dbs2go on specific Linux OS

For these reasons I don't think you can have reliable local build tree. In fact, to make it more compatible you should not choose `debian:trixie-slim` but rather you have choose the OS used for ORACLE libs. But this does not solve compatibility among your local OS and that one. What you should do is to ensure in your dockerfile this compatibility by checking , e.g. `objdump -T dbs2go | grep GLIBC` and check against OS used for final docker that this GLIBC version is supported.

todor-ivanov:
hi @vkuznet,
the main goal of this is not to have it runnable in local host, but just to have a mechanism for temporary docker builds (reusing exactly the `Dockerfile` for the regular docker builds) with one and only one difference, which is - the local `go` build command to be executed from the local tree instead of an already created upstream tag. It is for interim changes tests. So that said,  I was intentionally cutting all possible deviations form the current [Dockerfile](https://github.com/dmwm/CMSKubernetes/blob/master/docker/dbs2go/Dockerfile) preserving even the comment lines form the original, other than the 2 lines of code related to the actual process of copying the locally built executable inside the docker image. Here follows the full diff:
 
```
$ diff -u Dockerfile Dockerfile.localtree 
--- Dockerfile	2026-07-15 17:20:15.000000001 +0200
+++ Dockerfile.localtree	2026-07-15 16:28:01.000000001 +0200
@@ -1,3 +1,4 @@
+# syntax=docker/dockerfile-upstream:master
 FROM registry.cern.ch/cmsweb/oracle:21_5-stable as oracle
 #FROM cmssw/oracle:21_1 as oracle
 FROM cmssw/exporters:latest as exporters
@@ -18,23 +19,15 @@
 ENV WDIR=/data
 RUN mkdir -p $WDIR
 WORKDIR ${WDIR}
-ADD oci8.pc $WDIR/oci8.pc
+ADD oracle-env/oci8.docker.pc $WDIR/oci8.pc
 ENV PKG_CONFIG_PATH=$WDIR
 
-# build dbs2go with specific tag
-ENV TAG=v00.06.09
-WORKDIR $GOPATH/src/github.com/dmwm
-RUN git clone https://github.com/dmwm/dbs2go && \
-    git clone https://github.com/dmwm/cmsweb-exporters
-WORKDIR $GOPATH/src/github.com/dmwm/dbs2go
-
-# example how to checkout specific branch
-# RUN git checkout --track origin/api-struct && \
-RUN git checkout tags/$TAG -b build && \
-    sed -i -e "s,_ \"gopkg.in/rana/ora.v4\",,g" web/server.go && \
-    sed -i -e "s,_ \"github.com/mattn/go-sqlite3\",,g" web/server.go && \
-    sed -i -e "s,_ \"github.com/go-sql-driver/mysql\",,g" web/server.go && \
-    make && cp -r dbs2go static $WDIR
+# use dbs2go executable and static files built from the local tree
+ARG DBS2GO_BINARY=dbs2go
+ARG DBS2GO_BINARY_PATH=./${DBS2GO_BINARY}
+ARG DBS2GO_STATIC_PATH=${DBS2GO_BINARY_PATH%/*}/static
+ADD ${DBS2GO_BINARY_PATH} $WDIR/dbs2go
+ADD ${DBS2GO_STATIC_PATH} $WDIR/static

```

As of the Oracle shared libs and the drivers, If one looks at the latest version of the `Makefile` from the dbs2go PR: https://github.com/dmwm/dbs2go/pull/158 itself, one would immediately notice the local target `oracle-env`: https://github.com/dmwm/dbs2go/pull/158/changes#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R121-R153 , which fetches the exact same oracle drivers from the CMSKubernetes as it is for the upstream docker build. And the purpose of all this  is quite obvious - to perfectly repeat (1 to 1) the container builds, with only one difference - the code tree to not be fetched from upstream `dbs2go`  repository, but rather from the developer's local tree and to produce a docker image fully compatible with what is about to be created once a (pre)release is cut. And indeed, we know about the other limitations, that may come out of this - like the need this to be executed on a Linux machine possibly at CERN (due to the autmatic architecture identification) so that the produced container is an exact match. I have already fully tested it and it works perfectly at CERN machines. As for the local builds, which are supposed to be testable as locally running executable on the developer's machine,  the developers are still supposed to use the corresponding arch dependent targets as `make build_arm64` or just `make build` and rely on the auto arch identification etc... and eventually follow the logic with or without the oracle flags. The current docker file is used only through a `make docker build localtree` target. At the `Makefile` in question, I made a very strong effort to preserve all the targets for local builds and also the restore mechanism for `no-oracle` builds (even if a build fails), since those  are intrusive to the source tree itself and we cannot allow to skip a step (hence the existence of the `.IGNORE` clauses etc). But all that, of course, is very well known to you. My only point is - I am preserving things workable and backwards compatible at 100%, and only adding the mechanisms for interim code changes tests, which are not to be dependent on any github actions/tags/repository state. 

I'll address the rest of your  comments  in a moment.

todor-ivanov:
hi @vkuznet upon reconsidering the whole strategy during the night, I indeed agree with you that separating the build process in two steps - go code in the host and oracle drivers and  libraries  precompiled and matched to the container arch is going to always create friction. Even though I can check (as you suggest) for library to binary executable mismatch, the only valid move upon assertion such would be to just block the docker compilation and never produce anything runnable until the mismatch is not resolved on the host side. this would indeed lead to a non-reproducable behaviour which is not the goal here. So I switched to a different mechanism - packing the source code inside the build area and compiling from there. Much more clean and almost no diff between the original Docker file and this one:
```
$ diff -u Dockerfile   Dockerfile.localtree  
--- Dockerfile	2026-07-15 19:31:19.642579621 +0300
+++ Dockerfile.localtree	2026-07-16 13:52:43.661524987 +0300
@@ -1,9 +1,9 @@
 FROM registry.cern.ch/cmsweb/oracle:21_5-stable as oracle
-#FROM cmssw/oracle:21_1 as oracle
+
 FROM cmssw/exporters:latest as exporters
 FROM cmssw/filebeat:latest as filebeat
 FROM golang:latest as builder
-MAINTAINER Valentin Kuznetsov vkuznet@gmail.com
+MAINTAINER Todor Ivanov todor.ivanov@cern.ch
 
 # get oracle for our build
 COPY --from=oracle /usr/lib/oracle /usr/lib/oracle
@@ -21,20 +21,17 @@
 ADD oci8.pc $WDIR/oci8.pc
 ENV PKG_CONFIG_PATH=$WDIR
 
-# build dbs2go with specific tag
-ENV TAG=v00.06.09
+# build dbs2go from the local source tree
+ARG DBS2GO_SOURCE=src
+ARG DBS2GO_SOURCE_PATH=./${DBS2GO_SOURCE}
 WORKDIR $GOPATH/src/github.com/dmwm
-RUN git clone https://github.com/dmwm/dbs2go && \
-    git clone https://github.com/dmwm/cmsweb-exporters
+COPY ${DBS2GO_SOURCE_PATH} dbs2go
+RUN git clone https://github.com/dmwm/cmsweb-exporters
 WORKDIR $GOPATH/src/github.com/dmwm/dbs2go
-
-# example how to checkout specific branch
-# RUN git checkout --track origin/api-struct && \
-RUN git checkout tags/$TAG -b build && \
-    sed -i -e "s,_ \"gopkg.in/rana/ora.v4\",,g" web/server.go && \
+RUN sed -i -e "s,_ \"gopkg.in/rana/ora.v4\",,g" web/server.go && \
     sed -i -e "s,_ \"github.com/mattn/go-sqlite3\",,g" web/server.go && \
     sed -i -e "s,_ \"github.com/go-sql-driver/mysql\",,g" web/server.go && \
-    make && cp -r dbs2go static $WDIR
+    make VERSION=localtree && cp -r dbs2go static $WDIR
 
 # Install latest kubectl
 RUN curl -k -O -L https://storage.googleapis.com/kubernetes-release/release/$(curl -s https://storage.googleapis.com/kubernetes-release/release/stable.txt)/bin/linux/amd64/kubectl
```

vkuznet (commented):
I suggest to put comment that `golang:latest` image you start with to build dbs2go is compatible with `debian:trixie-slim` image you finish with. In fact, golang image is debian based and therefore they are compatible. And, oracle image is debian too, see https://github.com/dmwm/CMSKubernetes/blob/master/docker/oracle/Dockerfile

I think having such comment will be useful.

vkuznet (approved):

todor-ivanov:
thanks @vkuznet

PR #1705, opened 2026-07-20
Create DBS dev Environment

todor-ivanov:
## Summary

  Add a dedicated dbs2go-global-r-dev Kubernetes Service and Deployment to support the fast development workflow proposed in dbs2go issue : https://github.com/dmwm/dbs2go/issues/160.

  The development pod:

  - uses registry.cern.ch/cmsweb/dbs2go:dev;
  - provides the standard DBS Oracle runtime and mounted configuration/Secrets;
  - remains idle until a locally developed payload is copied and started;
  - can receive traffic by redirecting the regular dbs2go-global-r Service selector;
  - includes the corresponding development NetworkPolicy.

  This is initially a Reader pilot and does not change the normal release or production deployment workflow.

### DIFF
--- a/kubernetes/cmsweb/services/dbs2go-global-r-dev.yaml
+++ b/kubernetes/cmsweb/services/dbs2go-global-r-dev.yaml
@@ -1,0 +1,133 @@
+apiVersion: v1
+kind: Service
+metadata:
+  name: dbs2go-global-r-dev
+  namespace: dbs
+  labels:
+    app: dbs2go-global-r-dev
+spec:
+  selector:
+    app: dbs2go-global-r-dev
+  ports:
+  - port: 9252
+    targetPort: 9252
+    name: dbs
+---
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: dbs2go-global-r-dev
+  namespace: dbs
+  labels:
+    app: dbs2go-global-r-dev
+spec:
+  replicas: 1
+  selector:
+    matchLabels:
+      app: dbs2go-global-r-dev
+  template:
+    metadata:
+      labels:
+        app: dbs2go-global-r-dev
+        job: dbs2go-global-r-dev
+      annotations:
+        prometheus.io/scrape: "true"
+        prometheus.io/path: /dbs/prod/global/DBSReader/metrics
+        prometheus.io/port: "9252"
+    spec:
+      securityContext:
+        runAsUser: 1000
+        runAsGroup: 1000
+        fsGroup: 2000
+      containers:
+      - name: dev
+        image: registry.cern.ch/cmsweb/dbs2go:dev
+        imagePullPolicy: Always
+        command:
+        - sleep
+        - infinity
+        env:
+        - name: X509_USER_PROXY
+          value: /etc/proxy/proxy
+        - name: TNS_ADMIN
+          value: /etc/tnsnames.ora
+        ports:
+        - containerPort: 9252
+          protocol: TCP
+          name: dbs
+        resources:
+          requests:
+            memory: 100Mi
+            cpu: 200m
+          limits:
+            memory: 500Mi
+            cpu: 4000m
+        volumeMounts:
+        - name: proxy-secrets
+          mountPath: /etc/proxy
+          readOnly: true
+        - name: secrets
+          mountPath: /etc/secrets
+          readOnly: true
+        - name: robot-secrets
+          mountPath: /etc/robots
+          readOnly: true
+        - name: hmac-secrets
+          mountPath: /etc/hmac
+          readOnly: true
+        - name: token-secrets
+          mountPath: /etc/token
+          readOnly: true
+        - name: tnsnames
+          mountPath: /etc/tnsnames.ora
+          subPath: tnsnames.ora
+        securityContext:
+          allowPrivilegeEscalation: false
+      volumes:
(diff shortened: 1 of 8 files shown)


kubernetes/cmsweb/services/dbs2go-global-r-dev.yaml
```diff
@@ -0,0 +1,133 @@
...
+        secret:
+          secretName: dbs2go-global-r-secrets
+      - name: tnsnames
+        configMap:
+          name: tnsnames-config
+      - name: token-secrets
+        secret:
+          secretName: token-secrets
+---
+apiVersion: networking.k8s.io/v1
```
vkuznet, line 106: I don't think we need separate network policy for DBS. Why did you include it here? If policy indeed required it should be in separate manifest file and include all cmsweb services.

kubernetes/cmsweb/services/dbs2go-global-r-dev.yaml
```diff
@@ -0,0 +1,133 @@
...
+      labels:
+        app: dbs2go-global-r-dev
+        job: dbs2go-global-r-dev
+      annotations:
+        prometheus.io/scrape: "true"
+        prometheus.io/path: /dbs/prod/global/DBSReader/metrics
+        prometheus.io/port: "9252"
+    spec:
+      securityContext:
+        runAsUser: 1000
```
vkuznet, line 39: Are you sure that your container provides _dbs user with user ID 1000 and group ID 1000? If you'll use the groups here in security context then you must ensure that container provides them. Please verify and put appropriate comment.

kubernetes/cmsweb/services/dbs2go-global-r-dev.yaml
```diff
@@ -0,0 +1,133 @@
...
+        - containerPort: 9252
+          protocol: TCP
+          name: dbs
+        resources:
+          requests:
+            memory: 100Mi
+            cpu: 200m
+          limits:
+            memory: 500Mi
+            cpu: 4000m
```
vkuznet, line 64: Please consult Usakorn to ensure that cluster provides enough cpus per node. Since Go concurrency utilizes as many core as possible it is better even to increase the limits (but according to cluster node provisions).

vkuznet (changes requested):

kubernetes/cmsweb/services/dbs2go-global-r-dev.yaml
todor-ivanov: hi @vkuznet, 

The NetworkPolicy is required specifically for the development pods, not for the regular DBS service.

During `make -f devops.mk devinit`, the regular DBS Service selector is redirected from `app=dbs2go-global-r` to `app=dbs2go-global-r-dev`. The policy selects those `*-dev` pods and permits the CMSWEB authentication/frontend components to reach them. Without it, the Service redirection may select the correct development pods while at the same time the traffic to them remains blocked.

The policy therefore supports the tested development workflow only; it does not alter the default DBS deployment or its network behavior.

kubernetes/cmsweb/services/dbs2go-global-r-dev.yaml
todor-ivanov: I can do, but I am not concerned about overloading the cluster. During the DevOps procedures for which these manifests are intended, the regular service is scaled down and its traffic is redirected to a single pod from the corresponding `*-dev` deployment. This fast development workflow therefore releases enough cluster resources to accommodate the requested development pod and will generally reduce, rather than increase, overall resource usage.

kubernetes/cmsweb/services/dbs2go-global-r-dev.yaml
todor-ivanov: Yes, this has been verified against both the published image and a running development pod.

`Dockerfile.dev` creates the _dbs2go user and later configures it as the runtime user here:

- useradd (https://github.com/dmwm/CMSKubernetes/blob/master/docker/dbs2go/Dockerfile.dev#L56-L58)
- USER $USER (https://github.com/dmwm/CMSKubernetes/blob/master/docker/dbs2go/Dockerfile.dev#L82-L84)

The image published as `registry.cern.ch/cmsweb/dbs2go:dev` contains the so configured user with UID:GID :

```
$ podman run --rm --network=none --user 0:0 --entrypoint sh \
      registry.cern.ch/cmsweb/dbs2go:dev \
      -c 'getent passwd _dbs2go; getent group _dbs2go'
  _dbs2go:x:1000:1000::/data:/bin/sh
  _dbs2go:x:1000:
```

and:
```
  $ podman image inspect registry.cern.ch/cmsweb/dbs2go:dev \
      --format 'Configured runtime user: {{.Config.User}}'
  Configured runtime user: _dbs2go
```


And the running development pod is also showing the correct username UID and GUID:

```
_dbs2go@dbs2go-global-r-dev-6f74586ff4-z4grp:~$ id
  uid=1000(_dbs2go) gid=1000(_dbs2go) groups=1000(_dbs2go),2000
```

Therefore, `runAsUser: 1000` and `runAsGroup: 1000` do match what is already provided by the development image. I think we are safe here.

todor-ivanov:
Hi @vkuznet I have addressed your comments. Please take another look.

vkuznet (approved):

PR #1711, opened 2026-08-17
Add das-server/Dockerfile.dev

todor-ivanov:
## Summary

  Add docker/das-server/Dockerfile.dev for building the DAS development image from a locally staged das2go source tree.

  The image:

  - follows the DBS2Go development-image structure;
  - uses debian:trixie-slim;
  - includes debugging, networking, monitoring, filebeat, exporter, and kubectl tools;
  - builds and includes das2go, das2go_monitor, and the DAS web payload;
  - runs as an unprivileged user and remains available for the devpush workflow.

  The inclusion of DASTools is deferred for a separate decision.

  ## Related issues

  - Parent: dmwm/das2go#88 (https://github.com/dmwm/das2go/issues/88)
  - DAS Makefile integration: dmwm/das2go#89 (https://github.com/dmwm/das2go/issues/89)

  ## Validation

  Static file and build-context validation completed. The image has not yet been built or tested.

### DIFF
--- a/docker/das-server/Dockerfile.dev
+++ b/docker/das-server/Dockerfile.dev
@@ -1,0 +1,65 @@
+FROM cmssw/exporters:latest as exporters
+FROM cmssw/filebeat:latest as filebeat
+# The golang builder image is Debian-based and is compatible with the
+# debian:trixie-slim runtime used below.
+FROM golang:latest as builder
+MAINTAINER Valentin Kuznetsov vkuznet@gmail.com
+
+# Build das2go from the local source tree staged by the das2go Makefile.
+ENV WDIR=/data
+RUN mkdir -p $WDIR
+ARG DAS2GO_SOURCE=src
+ARG DAS2GO_SOURCE_PATH=./${DAS2GO_SOURCE}
+ARG CGO_ENABLED=0
+WORKDIR $GOPATH/src/github.com/dmwm
+COPY ${DAS2GO_SOURCE_PATH} das2go
+WORKDIR $GOPATH/src/github.com/dmwm/das2go
+RUN make VERSION=localtree && \
+    go build -o $WDIR/das2go_monitor -ldflags="-s -w -extldflags -static" monitor/das2go_monitor.go && \
+    cp -r das2go js css images templates examples $WDIR
+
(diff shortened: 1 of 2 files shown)


docker/das-server/Dockerfile.dev
```diff
@@ -0,0 +1,65 @@
...
+RUN chmod +x /usr/bin/kubectl
+
+# Pin the runtime distribution so its package names do not change underneath us.
+FROM debian:trixie-slim
+RUN apt-get update && apt-get -y install --no-install-recommends \
+        ca-certificates curl wget jq \
+        less vim-tiny emacs-nox \
+        file binutils strace lsof procps psmisc \
+        iproute2 iputils-ping dnsutils netcat-openbsd traceroute telnet \
+        sqlite3 && \
```
vkuznet, line 33: Todor, can you explain why do you need all of this dependencies? I'm truly curious why do you need sqlite, traceroute, telnet, etc.? Apart from that it only blow up the image. If you need any of this dependencies when you debug things, then login to pod and installing them is a way to go. This way you can install whatever you need and use it in your session, once you log out from the image and restart the pod they will be gone. Image truly does not need them. Please separate you debugging tasks from image itself.

todor-ivanov: this is a dev container  which would have one and only one image created, and no versions/tags preserved in the cern registry. And I need all those, because I need to use this image for development and debugging purposes. Last time I did have to trace network traffic, until I have reached the conclusion that the problems withing the cluster were indeed broken net AUTH layer in the das testbed  cluster. This image will not go in production service deployment.

vkuznet: Todor, this is exactly my point, you don't know a-priory which packages you may need next time you will debug the service. Why not install them at a debug time. Login to your image and install whatever you need. Today you may need one set of packages while tomorrow it will be another set. That said, I'm not against it but I do it differently.

vkuznet (approved):

docker/das-server/Dockerfile.dev
todor-ivanov: It is just a basic setup, all those I refrained to the bare minimum. If anything more is needed, I do add up during debug time.
thanks @vkuznet

PR #1712, opened 2026-08-26
new docker area to build diracx images

vkuznet:


vkuznet:
@uhomsuwa , I created a new area to build diracx images which eventually we may need to support (if it will be decided). Meanwhile I put two Dockerfiles and built/uploaded them to registry under cmsweb/diracx area. I'm going to test images on test18 and meanwhile you can review and merge this PR to automate building diracx images. Thanks!

uhomsuwa (approved):

